PR #9, opened 2018-08-06
Blue Sky Device Layer

jaylogue:
This PR incorporates the Blue Sky Device Layer into the OpenWeave source tree. In its current form, the Device Layer code only supports the Espressif ESP32. Generalization to other targets will happen in future commits.

Building the Device Layer is controlled via the new configure option `--with-device-layer`. Currently this option supports two arguments: esp32 and none. None (the default) disables building any of the Device Layer sources.

For the ESP32, the code within OpenWeave is designed to be built as a component within a larger ESP-IDF project. To support this, an ESP-IDF compatible "openweave" component directory is provided that contains the necessary make and configuration files (see build/esp32/components/openweave).

Note that because of the way the ESP-IDF build system works, there is no way to build OpenWeave by itself for the ESP32; it must always be incorporated into a larger ESP-IDF project (typically as a submodule).

Finally, the original Blue Sky commit history has been preserved for posterity, with changes adapted to the new location within OpenWeave.


gerickson:
I'd be lying if I said I went through every line of this code line-by-line; however, at a high-level it looks good and makes sense. The biggest change I'd advocate is delineating Device Layer Common vs. Device Layer ESP32 and partitioning each of those into their respective directories so it is clear how the layer scales to things beyond ESP32 and what must change for such an addition vs. what can be leveraged / reused.

jaylogue:
@gerickson : Support for other targets is in progress, but is specifically not included in this initial PR in order to bound the size. The goal is to get this into OpenWeave and usable by external parties while we undertake the process of generalizing the code structure.

turon:
@jaylogue, for this initial pass, why not satisfy @gerickson's ask by simply moving the device layer implementations into a `src/adaptations/device-layer/esp` directory triggered on passing `--use-adaptation=esp` to the configure command?

jaylogue:
@turon: I'm trying to avoid doing any throwaway work, especially if it only gives the appearance of introducing portability while not actually achieving it.  As I mentioned, work is underway to introduce a proper mechanism for supporting multiple target platforms.  In the meantime I want the ESP32-specific code to be "in the bank" and available for general use.

PR #28, opened 2018-09-10
Retargetable Device Layer

jaylogue:
OpenWeave Folks,

This PR represents the (previously promised) rework of the Device Layer to make it adaptable to different platforms.  This change represents a major restructuring of the internal workings of the Device Layer but only a minor change to its public API.  As before, the ESP32 is currently the only supported target.  However I’m hoping to change this in the near future.

As part of this effort, I have developed a number of new design patterns that I believe will make it easier to maintain new platform adaptations as time goes on.  The goal is to make it easy to share code between adaptations with less of the spaghetti #ifdef’s which have plagued nlweave-tools and nlweaveplatform.

The new design patterns are documented in a markdown document located in the Device Layer source tree (see [ADAPTATION-PATTERNS.md](https://github.com/openweave/openweave-core/blob/feature/jay/retargetable-device-layer/src/adaptations/device-layer/ADAPTATION-PATTERNS.md) ).  This is intended to serve as an overview for the purpose of this review, as well as a design introduction for new developers looking to adapt Weave for their own devices.

Those with limited time to spend should concentrate on reviewing the overall design.  Details in the code can be worked out later.


jaylogue:
For those interested in what changed in the API to the Device Layer, there's a corresponding PR for the openweave-esp32-demo project that captures the differences: [PR#3](https://github.com/openweave/openweave-esp32-demo/pull/3)

src/adaptations/device-layer/ADAPTATION-PATTERNS.md
```diff
@@ -0,0 +1,503 @@
...
+                                                                              generic base class. */
+{
+    ...
+};
+```
+
+In many cases, the generic implementation base class itself will directy provide some or all of the implementation methods
+needed to satisfy the component’s interface.  The rules of C++ method resolution are such that calls to a forwarding
+method on the interface class are mapped directly to the base class method.  In this situation, the derived implementation
+class needant declare a version of the target method at all, and method calls are forwarded statically, at compile time,
```
gerickson: needant -> needn't

src/adaptations/device-layer/Makefile.am
```diff
@@ -24,44 +24,58 @@ include $(abs_top_nlbuild_autotools_dir)/automake/pre.am
...
+    SystemEventSupport.cpp                \
+    SystemTimerSupport.cpp                \
+    TestDeviceIds.cpp                     \
+    TimeSyncManager.cpp                   \
+    TraitManager.cpp                      \
     trait-support/weave/trait/description/DeviceIdentityTrait.cpp \
     $(NULL)
 
-EXTRA_DIST                       = \
+if WEAVE_DEVICE_LAYER_TARGET_ESP32
```
gerickson: I suspect it will be more scalable over time if we pick these up through a device-layer/ESP32/Makefile.am or some such. Thoughts?

jaylogue: I'm unclear on how that would work.  These source files need to end up in the same library (libDeviceLayer.a), which is built by this Makefile.

Perhaps its possible to include a platform-specific Makefile during the bootstrap step?

gerickson: I was thinking something like:

if WEAVE_DEVICE_LAYER_TARGET_ESP32
include device-layer/ESP32/Makefile.am
endif

or:

if WEAVE_DEVICE_LAYER_TARGET_ESP32
include device-layer/ESP32/ESP32.am
endif

or something like that. We use an approach similar to this in a few other locations in the project.

jaylogue: Per our discussion off-line, I will leave the Makefile structure as-is and revisit when there are other platform adaptations.

PR #32, opened 2018-09-11
Initial Travis-CI and Codecov.io Integration

gerickson:
Part of making openweave-core open and resilient is taking advantage of public continuous integration and code coverage services available to open source projects, including Travis CI and Codecov.io.

This lays the basic ground work for leverage such services. We'll add in (working) build and validation targets with time to build up to our current, internal (downstream) level of coverage.


README.md
```diff
@@ -1,4 +1,8 @@
-[![OpenWeave][ow-logo]][ow-repo]
+[![OpenWeave][ow-logo]][ow-core-repo]
+[![Build Status][ow-core-travis-svg]][ow-core-travis]
```
Vyrastas: Just an aesthetic nit: maybe add a line with only "<br><br>" on it between ow-logo and ow-core-travis-svg so there's some spacing between them in the rendered version of the README.

gerickson: I followed the identical formatting to the README in openthread. I think if we change it, we should be consistent across the projects.

Vyrastas: Yeah on OT the "released by Nest" tagline adds some white space to the bottom of that logo by default, so they don't appear so close together.  A better solution is probably to add a little whitespace to the top/bottom of the OW logo.

gerickson: Fixed.

.travis/script.sh
```diff
@@ -0,0 +1,42 @@
...
+{
+    echo " *** ERROR: " ${*}
+    exit 1
+}
+
+
+case "${BUILD_TARGET}" in
+
+    linux-auto-*)
+        ./configure && make
```
robszewczyk: Any reason why we do not use the Makefile-Standalone here? Does travis impose any constraints on where the artifacts are built?

robszewczyk: Does this change originate from the requirements of the codecov.io scripts?   If not, it seems like we should also run make -f Makefile-Standalone check

PR #43, opened 2018-09-19
Fix Issues That Cause 'make dist' to Fail with ESP32-related Device Layer Adaptation

gerickson:


gerickson:
@jaylogue, I may need to work with you on this given that the ESP32 build is failing: https://travis-ci.org/openweave/openweave-core/jobs/430709916. I am uncertain how the master build can be succeeding with the makefile failures that exist in master against 'make dist'.

PR #45, opened 2018-10-01
Fix for WEAV-3145.

pidarped:
Try binding the TCP connection to the interface.
If that fails due to insufficient privileges, attempt to
bind the connection to one of the source addresses on given
interface.
For IPv6, the address binding selects the first global or ULA
address on that interface to bind to.


src/inet/TCPEndPoint.cpp
```diff
@@ -380,67 +380,53 @@ INET_ERROR TCPEndPoint::Connect(IPAddress addr, uint16_t port, InterfaceId intf)
...
-        // The only way to do this is to bind the local to an address on the desired interface,
-        // or bind the socket to the interface itself using SO_BINDTODEVICE.  The latter,
-        // however, requires privileged access.
-        if (addrType == kIPAddressType_IPv4)
+        // Try binding to the interface
+
+        struct ifreq ifr;
+        memset(&ifr, 0, sizeof(ifr));
+
+        res = GetInterfaceName(intf, intfStr, sizeof(intfStr));
```
jaylogue: You should eliminate intfStr and have GetInterfaceName copy the interface name directly into ifr.ifr_name.

src/inet/TCPEndPoint.cpp
```diff
@@ -2147,6 +2133,64 @@ void TCPEndPoint::LwIPHandleError(void *arg, err_t lwipErr)
 
 #if WEAVE_SYSTEM_CONFIG_USE_SOCKETS
 
+INET_ERROR TCPEndPoint::BindSrcAddrFromIntf(IPAddressType addrType, InterfaceId intf)
+{
+    INET_ERROR err = INET_NO_ERROR;
+
+    // If we are trying to make a TCP connection over IPv4 and a 'specified target interface',
+    // then we bind the TCPEndPoint to an IPv4 address on that target interface and use that
```
jaylogue: IPv4 or IPv6 address

src/inet/TCPEndPoint.cpp
```diff
@@ -2147,6 +2133,64 @@ void TCPEndPoint::LwIPHandleError(void *arg, err_t lwipErr)
...
+    IPAddress curAddr = IPAddress::Any;
+    InterfaceId curIntfId = INET_NULL_INTERFACEID;
+    bool ipAddrFound = false;
+
+    VerifyOrExit(State != kState_Bound, err = INET_ERROR_NOT_SUPPORTED);
+
+    for (InterfaceAddressIterator addrIter; addrIter.HasCurrent(); addrIter.Next())
+    {
+        curAddr = addrIter.GetAddress();
+        curIntfId = addrIter.GetInterface();
```
jaylogue: Test for curIntfId == intf here?

src/inet/TCPEndPoint.cpp
pidarped: Done. Thanks.

src/inet/TCPEndPoint.cpp
pidarped: It would apply for both actually. So, have reworded the text.

src/inet/TCPEndPoint.cpp
pidarped: That's right. It's being redundantly being checked at both places.

PR #47, opened 2018-10-02
Fix bug in error handling of interface binding of TCP connection.

pidarped:


src/inet/TCPEndPoint.cpp
```diff
@@ -419,7 +419,7 @@ INET_ERROR TCPEndPoint::Connect(IPAddress addr, uint16_t port, InterfaceId intf)
                 return res;
         }
 #ifdef SO_BINDTODEVICE
-        else
+        else if (r < 0)
```
jaylogue: This might be easier to understand if the code within the #ifdef was moved into the #ifdef above.

```
#ifdef SO_BINDTODEVICE

int r = setsockopt(mSocket, SOL_SOCKET, SO_BINDTODEVICE, (void *)&ifr, sizeof(ifr));

if (r < 0 && errno != EACCESS)
{
	return res = Weave::System::MapErrorPOSIX(errno);
}
if (r < 0)
#endif // SO_BINDTODEVICE
{
	res = BindSrcAddrFromIntf(addrType, intf);
}
````

PR #49, opened 2018-10-03
Fix a build failure for the Nest Detect simulation platform.

pidarped:


src/inet/TCPEndPoint.cpp
```diff
@@ -2185,6 +2185,7 @@ INET_ERROR TCPEndPoint::BindSrcAddrFromIntf(IPAddressType addrType, InterfaceId
                     break;
                 }
             }
+#endif // INET_CONFIG_ENABLE_IPV4
```
jaylogue: This doesn't make any sense to me.   It effectively makes this whole function a no-op if IPv4 is disabled.

pidarped: This patch was mostly aimed at fixing the build failure quickly. 
Also, I think it stems from the fact that for IPv6, in the absence of a proper checker for a global unicast address, I had to use the function IsIPv4() which is conditionally compiled when IPv4 is enabled.
Possibly, ways to fix it could be : 
1. Take IsIPv4() outside the INET_CONFIG_ENABLE_IPV4
2. Define a separate IsGlobalUnicastIPv6()

PR #51, opened 2018-10-07
WeaveKeyExportDelegate Enhancements

jaylogue:
-- Revised the WeaveKeyExportDelegate API to support platforms where access to the local node’s private key is restricted to a hardware element or a secure enclave.  This is a parallel change to the recent WeaveCASEAuthDelegate change.

-- Restructured the WeaveKeyExportDelegate API such that the delegate methods take a pointer to the key export engine object.  This simplifies the means by which context information is passed to the delegate.

-- Cleaned-up the API of the WeaveKeyExport object by making some members private and adding accessor methods.  Also, moved WeaveMessageInfo context pointer into the WeaveKeyExport object so that it doesn’t have to be passed on each delegate API call.

-- Added a compile-time compatibility mode to the delegate object that supports the legacy WeaveKeyExportDelegate API.  This mode is enabled via the WEAVE_CONFIG_LEGACY_KEY_EXPORT_DELEGATE configuration option.

The legacy mode is enabled by default to provide compatibility for existing platforms.  For stand-alone builds the new API is enabled to allow for testing.

-- Replaced the various GenerateWeaveSignature() public functions with a set of WeaveSignatureGenerator classes whose methods provide similar functionality. These classes rely on virtual methods calls to compute the low-level signature data, which allows an API client (in this case the key export engine) to own the signing operation.

-- Updated various instances of key export delegate objects to support the new API.  In all cases the code supports both the new API and the legacy API.


src/lib/profiles/security/WeaveKeyExport.cpp
```diff
@@ -516,65 +536,115 @@ WEAVE_ERROR WeaveKeyExport::AppendNewECDHKey(uint8_t *& buf)
...
+        GenerateSHA256Hash(msgStart, msgLen, msgHash);
 
-exit:
-    if (callReleaseNodePrivateKey)
-    {
-        WEAVE_ERROR endErr = KeyExportDelegate->ReleaseNodePrivateKey(isInitiator, signingKey);
-        if (err == WEAVE_NO_ERROR)
-            err = endErr;
+        // Determine the location at which the signature should be encoded.
+        uint8_t * msgSigStart = msgStart + msgLen;
```
gerickson: Minor nit: const-qualify msgSigStart.

src/lib/profiles/security/WeaveKeyExport.cpp
```diff
@@ -516,65 +536,115 @@ WEAVE_ERROR WeaveKeyExport::AppendNewECDHKey(uint8_t *& buf)
...
+        uint16_t maxSigSize = msgBufSize - msgLen;
```
gerickson: Minor nit: const-qualify maxSigsize.

src/lib/profiles/security/WeaveKeyExportTest.cpp
```diff
@@ -77,7 +79,106 @@ class SimulatedDeviceKeyExportDelegate : public WeaveKeyExportDelegate
...
+#if !WEAVE_CONFIG_LEGACY_KEY_EXPORT_DELEGATE
+
+    WEAVE_ERROR GetNodeCertSet(WeaveKeyExport * keyExport, WeaveCertificateSet& certSet) __OVERRIDE
+    {
+        WEAVE_ERROR err;
+        WeaveCertificateData *cert;
+
+        VerifyOrExit(!keyExport->IsInitiator(), err = WEAVE_ERROR_INVALID_ARGUMENT);
+
+        err = certSet.Init(10, 4096);
```
gerickson: Magic numbers: 10 and 4096

src/lib/profiles/security/WeaveKeyExportTest.cpp
```diff
@@ -77,7 +79,106 @@ class SimulatedDeviceKeyExportDelegate : public WeaveKeyExportDelegate
...
+
+    WEAVE_ERROR BeginCertValidation(WeaveKeyExport * keyExport, ValidationContext & validCtx,
+            WeaveCertificateSet & certSet) __OVERRIDE
+    {
+        WEAVE_ERROR err;
+        WeaveCertificateData *cert;
+
+        VerifyOrExit(!keyExport->IsInitiator(), err = WEAVE_ERROR_INVALID_ARGUMENT);
+
+        err = certSet.Init(10, 4096);
```
gerickson: Same magic numbers as above: 10 and 4096

PR #58, opened 2018-10-12
Address a Unit Test Failure against TestWeaveTool.py

gerickson:
The test script, TestWeaveTool.py, runs against the 'weave' tool produced by tools/weave/.... Consequently, tools/weave must be iterated and built before test-apps. Otherwise, the TestWeaveTool.py test will fail.

This reorders SUBDIRS such that tools/weave appears before test-apps to satisfy this dependency.


gerickson:
The other option, of course, is to move TestWeaveTool.py to be a test within tools/weave, colocating the code, the tool, and that which tests it, eliminating this dependency.

PR #65, opened 2018-10-19
Address Various Compiler Warnings

gerickson:


src/test-apps/PASEEngineTest.cpp
```diff
@@ -234,18 +244,14 @@ enum
 };
 
 //end PASEEngineTest Initialization
-void PASEEngineTest::Run()
+void PASEEngineTest::Run(void)
 {
+    PASEEngineTestContext *lContext;
     WEAVE_ERROR err;
-    WeavePASEEngine initiatorEng;
-    WeavePASEEngine responderEng;
```
jaylogue: Why can't these be local?

src/test-apps/TestEventLogging.cpp
```diff
@@ -2034,6 +2033,8 @@ static void CheckGapDetection(nlTestSuite *inSuite, void *inContext)
     event_id_t eventId_A = nl::Weave::Profiles::DataManagement::LogEvent(testSchema, nl::SerializedDataToTLVWriterHelper, (void *)&appData);
     event_id_t eventId_B = nl::Weave::Profiles::DataManagement::LogEvent(testSchema, nl::SerializedDataToTLVWriterHelper, (void *)&appData);
 
+    (void)eventId_B;
```
jaylogue: Should we use IgnoreUnusedVariable() for this?

src/lib/profiles/data-management/Current/MessageDef.cpp
```diff
@@ -2741,13 +2746,19 @@ WEAVE_ERROR StatusList::Parser::CheckSchemaValidity(void) const
...
+
         // TODO: The spec says the StatusList should be an array of arrays, but in
         // the current implementation it's an array of structures. The array of
         // arrays is less intuitive but more space efficient.
-        VerifyOrExit(nl::Weave::TLV::AnonymousTag == reader.GetTag(), err = WEAVE_ERROR_INVALID_TLV_TAG);
 
-        VerifyOrExit((nl::Weave::TLV::kTLVType_Structure == reader.GetType() ||
-                      nl::Weave::TLV::kTLVType_Array), err = WEAVE_ERROR_WRONG_TLV_TYPE);
+        tag = reader.GetTag();
+        VerifyOrExit(nl::Weave::TLV::AnonymousTag == tag, err = WEAVE_ERROR_INVALID_TLV_TAG);
```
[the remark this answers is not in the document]
gerickson: Note that the right side of the OR (||) in 👍 

```
(nl::Weave::TLV::kTLVType_Structure == reader.GetType() || nl::Weave::TLV::kTLVType_Array)
```

is effectively always true. The solve is either:

```
(nl::Weave::TLV::kTLVType_Structure == reader.GetType() || nl::Weave::TLV::kTLVType_Array == reader.GetType())
```

and hope there are no side effects to reader.GetType(). But that seems sloppy anyway, so that leaves reading type to a local and then verifying:

```
type = reader.GetType();
(nl::Weave::TLV::kTLVType_Structure == type || nl::Weave::TLV::kTLVType_Array == type)
```

Frankly, I dislike the above coding style anyway and would have done:

```
type = reader.GetType();
(type == nl::Weave::TLV::kTLVType_Structure || type == nl::Weave::TLV::kTLVType_Array)
```

but I tried to keep most of these fixes surgical.

src/test-apps/TestEventLogging.cpp
gerickson: Happy to use that.

src/test-apps/PASEEngineTest.cpp
gerickson: Because to ensure we don't run afoul of both Sapphire and RTOS -Wframe-size=X warnings, we have replicated this warning in Weave. These two plus the other two blow X by about 2x.

Since it's test code that isn't typically run on device (or at least devices that can't run multiple programs with main()), new/delete was an easy solve.

jaylogue: I have it on my to-do list to remove the -Wframe-larger-than options from OpenWeave, as these can never be universally applied across all platforms.  (At a minimum it seems incorrect to limit a 64-bit pointer platform to using the same stack space as a 32-bit one).

PR #70, opened 2018-10-25
Address Overloaded Virtual Warnings

gerickson:


src/test-apps/MockDMPublisher.h
```diff
@@ -46,7 +48,9 @@ class MockDMPublisher : public nl::Weave::Profiles::DataManagement::DMPublisher
 
     WEAVE_ERROR UpdateIndication(ExchangeContext *aResponseCtx, ReferencedTLVData &aDataList);
 
-    void IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport);
+    virtual void IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport);
+
+    using DMPublisher::IncompleteIndication;
```
robszewczyk: I don't quite understand.  I believe the the intention here was to have the method above override the `DMPublisher::IncompleteIndication`; Arguably, it would be appropriate to add the `__OVERRIDE` macro to the `IncompleteIndication`.  Because the `IncompleteIndication` already is declared virtual in base, there is no need to declare it virtual here, right?

src/test-apps/TestDataManagement.cpp
```diff
@@ -327,11 +327,13 @@ class DMTestClient :
         return err;
     }
 
-    void IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport)
+    virtual void IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport)
```
robszewczyk: Unnecessary virtual, but perhaps  a beneficial override?

src/test-apps/wdmtest.cpp
```diff
@@ -510,10 +511,13 @@ class WDMTestClient :
         return err;
     }
 
-    void IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport)
+    virtual void IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport)
```
robszewczyk: Unnecessary virtual but beneficial override?

src/test-apps/MockDMPublisher.h
gerickson: I agree, this seems an odd warning. As Jay and I discussed this morning, it might make sense long-term to get rid of "-Wall" in configure.ac and select those warnings we feel actually make sense and add real, measurable value.

Using __OVERRIDE here might be better though that may then trigger -Winconsistent-missing-override. I'll give it a try.

src/test-apps/TestDataManagement.cpp
gerickson: Agreed. See above.

src/test-apps/wdmtest.cpp
gerickson: Agreed. See above.

src/test-apps/MockDMPublisher.h
gerickson: The using statement here seems to be the required magic but it doesn't make much sense to me based on my knowledge of the language.

The better approach here might be just to disable -Woverloaded-virtual (-Wno-overloaded-virtual) for clang. Thoughts @robszewczyk or @jaylogue?

jaylogue: IncompleteIndication() has a different signature on the base class:
`void IncompleteIndication(Binding *aBinding, StatusReport &aReport)`
vs.
`void IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport)`
This is likely erroneous, as nothing appears to call the subclass method.

src/test-apps/TestDataManagement.cpp
jaylogue: As with MockDMPublisher I think this is an error, as nothing calls `DMTestClient::IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport)`.

src/test-apps/MockDMPublisher.h
gerickson: Looking at this again, `src/lib/profiles/data-management/Legacy/ProtocolEngine.h` defines both:

```
virtual void IncompleteIndication(Binding *aBinding, StatusReport &aReport);
```

and:

```
virtual void IncompleteIndication(const uint64_t &aPeerNodeId, StatusReport &aReport) = 0;
```

The implementation of the former calls the latter making the base class abstract. It seems like the test clients correctly override the latter which seems to imply that the warning is a bit pedantic.

src/test-apps/TestDataManagement.cpp
gerickson: See https://github.com/openweave/openweave-core/pull/70#discussion_r228686517.

PR #72, opened 2018-10-25
Enable applications to skip providing data for an external event.

Glen-Weaver:
Inside an application's fetch event callback, it can call a new
API, LoggingManagement::SkipEvent, instead of
LoggingManagement::BlitEvent if for some reason it is not able
to provide the corresponding event data.

Change-Id: I1ecddb26e7c1aec377a49442a331702221ee9549


Glen-Weaver:
@robszewczyk, I have rebased this commit and unit tests pass.

PR #73, opened 2018-10-25
Address -Wunused-private-field Warnings

gerickson:


src/lib/profiles/data-management/Current/NotificationEngine.h
```diff
@@ -266,8 +266,6 @@ class NotificationEngine
         NotifyRequestBuilderState mState;
         PacketBuffer * mBuf;
         SubscriptionHandler * mSub;
-        uint32_t mMaxNotificationSize;
```
robszewczyk: These were introduced as a part of the SubscriptionlessNotify changeset; however, it looks like they are not needed because the are computed from the Binding/Packet buffer

src/lib/profiles/data-management/Current/TraitData.h
```diff
@@ -127,9 +128,6 @@ class UpdateDirtyPathFilter : public IPathFilter
 private:
     SubscriptionClient *mpSubClient;
     TraitDataHandle mTraitDataHandle;
-    bool mFilterPendingUpdate;
```
robszewczyk: The usage of those was removed when @gandreello refactored the WDM update code

src/lib/profiles/data-management/Current/TraitData.h
```diff
@@ -1,5 +1,6 @@
 /*
  *
+ *    Copyright (c) 2018 Google LLC.
```
emargolis: Is that the new copyright message that we need to incorporate in all new files?
Don't mention Nest Labs any more?

gerickson: Confirmed.

PR #74, opened 2018-10-25
Address -Winconsistent-missing-override Warnings

gerickson:


src/test-apps/TAKEOptions.h
```diff
@@ -114,7 +114,7 @@ class TAKEOptions : public OptionSetBase
     const uint8_t *EncAK;
     bool ForceReauth;
 
-    virtual bool HandleOption(const char *progName, OptionSet *optSet, int id, const char *name, const char *arg);
+    virtual bool HandleOption(const char *progName, OptionSet *optSet, int id, const char *name, const char *arg) __FINAL __OVERRIDE;
```
jaylogue: Not that it needs changing, but was final necessary here to quiet the warning?  Likewise, final override seems redundant.  Was that necessary as well?

gerickson: I doubt it. I'll remove and re-run.

gerickson:
https://github.com/openweave/openweave-core/issues/76 for further discussion.

PR #75, opened 2018-10-25
Simplify and Generalize How Non-portable Include Path Warnings are Handled for clang/LLVM

gerickson:


configure.ac
```diff
@@ -293,36 +293,22 @@ PROSPECTIVE_CXXFLAGS=""
...
-        case "${nl_filtered_target}" in
-        
-            *-apple-darwin)
-                AX_CHECK_COMPILER_OPTIONS([C],   MAYBE_WNONPORTABLE_INCLUDE_PATH_CFLAG,   -Wnonportable-include-path)
-                AX_CHECK_COMPILER_OPTIONS([C++], MAYBE_WNONPORTABLE_INCLUDE_PATH_CXXFLAG, -Wnonportable-include-path)
-                
-                if test ${MAYBE_WNONPORTABLE_INCLUDE_PATH_CFLAG} = "-Wnonportable-include-path"; then
-                    AX_CHECK_COMPILER_OPTIONS([C],   -Wno-nonportable-include-path)
-                fi
+if test `basename ${CC}` = "clang"; then
```
robszewczyk: This hits the case for the native compiler, but not for the crosscompilers based on clang.   Perhaps a better approach would be:

    case `basename ${CC}` in
        *clang*)
            AX_CHECK_COMPILER_OPTIONS([C],   -Wno-nonportable-include-path)
            ;;
    esac

gerickson: Good suggestion.

PR #78, opened 2018-11-01
Added code to abort the mock time sync client connection on failure

gandreello:
Simply closing the connection was leaving a TCP socket retrying
transmissions in a failed case when running with plaid, and that in
turn was being flagged as a resource leak.

Change-Id: Ic25ec1fe16381e74e77c83dc66a35ccc4e519775

### DIFF
--- a/src/test-apps/MockTimeSyncClient.cpp
+++ b/src/test-apps/MockTimeSyncClient.cpp
@@ -329,7 +329,7 @@ void MockTimeSyncClient::CloseConnectionToService(void)
     {
         WeaveLogProgress(TimeService, "App closing connection to service");
 
-        mConnectionToService->Close();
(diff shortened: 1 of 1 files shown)


src/test-apps/MockTimeSyncClient.cpp
```diff
@@ -338,7 +345,7 @@ void MockTimeSyncClient::SetupConnectionToService(void)
 {
     WEAVE_ERROR err = WEAVE_NO_ERROR;
 
-    CloseConnectionToService();
+    CloseConnectionToService(err);
```
jaylogue: Shouldn't this abort the connection?

gandreello: I have reworked this so that it always calls Abort.

PR #81, opened 2018-11-06
Generalize LwIP build to support multiple environments

jaylogue:
The process by which Weave builds its internal version of LwIP has been changed to provide support for building LwIP for different target environments.

Previously, when the `--with-lwip=internal` configure option was specified, the build system would build LwIP specifically for use in "standalone" (i.e. linux desktop) environments.  This change introduces a new configure option, `--with-lwip-target=<target>`, that developers can use to select other target
environments.  The new option is only meaningful when `--with-lwip=internal` has been specified.  Thus the behavior of building Weave with an external LwIP package is unchanged.

The `--with-lwip-target=<target>` option works by selecting among a set of target-specific configuration headers and sources located in the src/lwip directory (e.g. `standalone/arch/cc.h`, `standalone/arch/sys_arch.h`, `standalone/sys_arch.c`, etc.). Currently there are two choices for the `--with-lwip-target` option: _standalone_ and _none_.

_standalone_ builds LwIP as before, for the linux desktop environment. This is the default.  _none_ instructs the system to build LwIP without any environment-specific files.  Developers can use this to incorporate their own files (e.g. via -I in CPPFLAGS) while still relying on Weave to do the work of building LwIP.


jaylogue:
Yeah.  The problem is the way conditionals work in automake.  You can't use a conditional unless its been defined using AM_CONDITIONAL.  But the new conditional I added in nl_with_lwip.m4 only gets invoked when configure.ac calls the NL_WITH_LWIP function, which in turn only happens when WEAVE_SYSTEM_CONFIG_USE_LWIP is defined.  So non-LwIP builds fail with "conditional ... was never defined".

RS suggested that we pass the value of ${WEAVE_SYSTEM_CONFIG_USE_LWIP} to NL_WITH_LWIP as an argument and then move the "are we using LwIP" test into NL_WITH_LWIP itself.  That way NL_WITH_LWIP can declare all the AM_CONDITIONALs associated with LwIP, simplifying configure.ac in the process, and avoiding the undefined conditional problem.

I'm interested in your thoughts on this approach.

PR #85, opened 2018-11-15
Update to nlbuild-autotools 1.6.4

gerickson:


gerickson:
This is, admittedly, an incremental improvement to the current state of affairs for the dependency resolution and pulling. We can improve it over time more fully and decouple the pull / configure / build; however, this is not that change.

PR #93, opened 2018-11-30
Add Common:LocalSetupRequied status code

jaylogue:
Added a new Common Profile status code for use when an operation is rejected because one or more necessary local configuration steps have not been completed. “Local configuration steps” in this context means any prerequisite actions that must be performed locally on the target, e.g. physically connecting wires to a device, before the requested operation can be invoked.


jaylogue:
Note to reviewers: There were a bunch of whitespace changes in StatusReportStr.cpp that make the change look bigger than it really is.

robszewczyk:
I've restarted the failed CI job; it seems like that is a flaky test.

PR #96, opened 2018-12-10
Adding chromecast setup service uuid to the ble scan filter

SukeshGoogle:
Adding chromecast setup service uuid to the ble scan filter

Change-Id: Ia4079af9a4df5034b62b76a92236ea7bf988ccfe


src/device-manager/python/WeaveBluezMgr.py
```diff
@@ -54,6 +54,8 @@
 weave_tx      = uuid.UUID('18EE2EF5-263D-4559-959F-4F9C429F9D11')
 weave_rx      = uuid.UUID('18EE2EF5-263D-4559-959F-4F9C429F9D12')
 weave_service_short = uuid.UUID('0000FEAF-0000-0000-0000-000000000000')
+chromecast_setup_service = uuid.UUID('0000FEA0-0000-1000-8000-00805F9B34FB')
```
jaylogue: Is the chromecast team okay with this information becoming public / open source?

SukeshGoogle: As this information is part of the ble advertisements emitted by cast devices, which can be seen with any ble scanner, I don't think this UUID is private anymore.

SukeshGoogle: Chromecast team seems okay with this change.

PR #99, opened 2018-12-13
DeviceManager.py: Allow add-thread-network with unspecified thread network key.

jhwoodyatt:
When the add-thread-network command is used in more recently defined pairing workflows, the thread network key may be unspecified. On enabling such a network provision, the Weave device is expected to generate its own new thread network key in that case.


jhwoodyatt:
Gentle ping.. @robszewczyk or somebody please? I don't have permission to ask for reviewers.

PR #100, opened 2018-12-14
Fix compiling on 64bit platform

bukepo:
`EncodeTLV` accepts expects `uint32_t`, while `_GetDeviceDescriptorTLV` gives `size_t`, which is different on 64bit system. This PR fixes this issue by provide a temporary `uint32_t` to accept the output of `EncodeTLV()`.


bukepo:
I was playing openweave on Linux.

PR #103, opened 2018-12-17
First Round of Refactoring for UDPEndPoint

gerickson:
This is the first in a series of refactorings that migrates common BSD-socket-specific implementation out of the UDPEndPoint class and into a new, intermediate IPEndPoint basis class from which the UDPEndPoint class will now derive.

The intent of this is two fold:

1. The intermediate IPEndPoint basis class will serve as a future home for forthcoming new common, expanded IP multicast implementation, particularly for BSD sockets, but also for LwIP.
2. The intermediate IPEndPoint basis class will serve as a point to further unify and harmonize IP datagram operations between RawEndPoint and UDPEndPoint, including unifying IPPacketInfo, since these two classes are more similar than different.

### DIFF
--- a/build/cstyle/cstyle.options
+++ b/build/cstyle/cstyle.options
@@ -1,5 +1,5 @@
 -v
---copyright='^( \*|\/\*|\/\/)    Copyright (\(c\) )*\d{4,}(-\d{4,})* Nest Labs, Inc\.'
+--copyright='^( \*|\/\*|\/\/)    Copyright (\(c\) )*\d{4,}(-\d{4,})* (Nest Labs, Inc\.|Google LLC\.*)'
 --cpp-lines-between-conditionals=20
 --line-length=132
 --tab-size=4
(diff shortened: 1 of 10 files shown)


gerickson:
I could see arguments to name IPEndPointBasis as IPDatagramEndPointBasis, instead, since it is not clear that TCPEndPoint should, can, or will leverage it.

gerickson:
Fixed the change set / commit inconsistency that had introduced unrelated WDM / event logging changes.

jhwoodyatt:
At first glance, introducing the `IPEndPointBasis` class for `UDPEndPoint` alone to inherit seems like a case of premature abstraction. The methods defined in `IPEndPointBasis` are only useful in the case where `domain == SOCK_DGRAM && (family == PF_INET || family == PF_INET6)`, i.e. UDP. (I guess you could get at UDP-Lite with `protocol == IPPROTO_UDPLITE`, but I don't actually believe the way to do that would be to define a `UDPLiteEndPoint` class separate from `UDPEndPoint`.)

What is the intention behind introducing this new layer in the InetLayer::EndPointBasis class hierarchy?

PR #104, opened 2018-12-18
Improve TestInetEndPoint Assertions

gerickson:
While the Inet layer raw IP end points are tested in a number of places functionally, return values from several method calls are dropped and go unchecked.

To address this, this change set first enforces the correct state on raw IP end point listen, to make its behavior consistent with UDP and TCP end points.

Second, this change set check returns values from raw IP endpoint bind and listen operations in _TestInetEndPoint_, handling them in a stateful way such that the test passes and fails when it should.

### DIFF
--- a/src/inet/RawEndPoint.cpp
+++ b/src/inet/RawEndPoint.cpp
@@ -300,6 +300,9 @@ INET_ERROR RawEndPoint::Listen()
     if (mState == kState_Listening)
         return INET_NO_ERROR;
 
+    if (mState != kState_Bound)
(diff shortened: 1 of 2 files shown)


gerickson:
@robszewczyk, I addressed the 'l*' naming convention feedback.

robszewczyk:
@gerickson thanks!

jhwoodyatt:
> and it did somewhat ironically.
Hey now... I don't recall emitting even a peep of protest about adopting a naming convention that I personally have never liked not even a little bit. I was earnestly trying to adopt best current practice at the time.

PR #111, opened 2018-12-21
Convert ECDSA signature from DER to Weave

jialong-google:
Code to convert ECDSA signature from DER to Weave


jaylogue:
@liFrank Is this ready to be merged?

liFrank:
@jaylogue Yes it is.

PR #120, opened 2019-01-11
Thread improvements to Java device manager.

jhwoodyatt:
Improve Weave Device Manager (java) to allow Thread network information objects with explicit Thread PAN identifier and channel values.

### DIFF
--- a/src/device-manager/java/src/nl/Weave/DeviceManager/NetworkInfo.java
+++ b/src/device-manager/java/src/nl/Weave/DeviceManager/NetworkInfo.java
@@ -1,6 +1,7 @@
 /*
  *
- *    Copyright (c) 013-2017 Nest Labs, Inc.
+ *    Copyright (c) 2019 Google LLC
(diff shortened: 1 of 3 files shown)


src/device-manager/java/src/nl/Weave/DeviceManager/NetworkInfo.java
```diff
@@ -79,6 +90,15 @@ public NetworkInfo()
...
     public byte[] ThreadNetworkKey;
 
+    /** The 16-bit Thread PAN ID, or kThreadPANId_NotSpecified 
+         */
+    public int ThreadPANId;
+
+    /** The current channel (currently [11..26]) on which the Thread
+         * network operates, or kThreadChannel_NotSpecified
+         */
+    public byte ThreadChannel;
```
jaylogue: I think its not very conventional, nor very convenient for the Java programmer, to use byte as an integer type, except where memory usage is a factor (e.g. big arrays).   So I suggest changing this to int, a la ThreadPANId.

PR #123, opened 2019-01-14
Fix for Issue 122: TestDNSResolution: Assert in dns_exec_found_callback: "invalid table index"

jaylogue:
-- Corrected logic bug in LWIP_ASSERT statement in dns_exec_found_callback().


gerickson:
The code in question is a site-specific change, correct? I don't see any reference to _dns_exec_found_callback_ upstream.

robszewczyk:
While it is not an error, it caused me a double take: I would prefer the `LWIP_ASSERT` before "use".  I know that it is not quite use but a mere address computation, but in C++ we can have a variable declaration anywhere, so this would be a minimal cost.  Not a blocking comment, though.

PR #126, opened 2019-01-14
change IP.py to module util

jenniexie:
change the way test script call IP.py module util because of below Happy PR:
https://github.com/openweave/happy/pull/10

### DIFF
--- a/src/test-apps/happy/lib/WeaveNetworkGateway.py
+++ b/src/test-apps/happy/lib/WeaveNetworkGateway.py
@@ -30,6 +30,7 @@ import sys
 
 from happy.ReturnMsg import ReturnMsg
 from happy.Utils import *
+from happy.utils.IP import IP
 from happy.HappyNetwork import HappyNetwork
 from happy.HappyNode import HappyNode
 import happy.HappyNetworkRoute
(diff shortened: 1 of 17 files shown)


src/test-apps/happy/lib/WeaveNetworkGateway.py
```diff
@@ -30,6 +30,7 @@
 
 from happy.ReturnMsg import ReturnMsg
 from happy.Utils import *
+from happy_utils.IP import IP
```
gerickson: We already have happy.Utils. What is the rationale for creating something parallel called happy_utils.IP? Can we not just create happy.Utils.IP as a new subset of happy.Utils?

jenniexie: Utils is a python file in happy which is Utils.py, 
happy_utils is a folder to put all utils in Happy.
As first step, I only put IP.py util under happy_utils, I plan to review and move Utils.py under happy_utils folder future.

.travis/before_install.sh
```diff
@@ -119,7 +119,8 @@ case "${BUILD_TARGET}" in
         installdeps "happy-deps"
 
         cd $HOME
-        git clone https://github.com/openweave/happy.git
+        git clone https://github.com/jenniexie/happy.git
+        #git clone https://github.com/openweave/happy.git
```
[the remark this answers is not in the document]
jenniexie: I have just changed back to git clone happy master for this one because I merged my changes to happy master:
 " git clone https://github.com/openweave/happy.git "

Issue #124, opened 2019-01-14
TestDNSResolution: Seg fault in udp_sendto


jaylogue:
The TestDNSResolution unit test fails with a segmentation fault in udp_sendto() when built for LwIP.

```
#0  0x000000000047abee in udp_sendto (pcb=0x900, p=0x6dcac8 <ram_heap+8>, dst_ip=0x6dc260 <dns_servers>, dst_port=53)
    at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/core/udp.c:532
#1  0x000000000045f89d in dns_send (idx=1 '\001') at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/core/dns.c:886
#2  0x000000000046034e in dns_check_entry (i=1 '\001') at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/core/dns.c:1150
#3  0x0000000000460434 in dns_check_entries () at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/core/dns.c:1184
#4  0x000000000045eb4b in dns_tmr () at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/core/dns.c:424
#5  0x0000000000482b67 in cyclic_timer (arg=0x6c5da0 <lwip_cyclic_timers+96>)
    at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/core/timeouts.c:177
#6  0x0000000000482f9b in sys_check_timeouts () at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/core/timeouts.c:364
#7  0x000000000048307f in sys_timeouts_mbox_fetch (mbox=0x6dbbd0 <mbox>, msg=0x7ffff69dcec0)
    at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/core/timeouts.c:437
#8  0x000000000045df95 in tcpip_thread (arg=0x0) at /home/jay/projects/openweave-core/src/lwip/../../third_party/lwip/repo/lwip/src/api/tcpip.c:107
#9  0x00007ffff7bc4184 in start_thread (arg=0x7ffff69dd700) at pthread_create.c:312
[...]

PR #127, opened 2019-01-15
Fix for Issue #124: TestDNSResolution: Seg fault in udp_sendto

jaylogue:
Corrected a logic bug in LwIP’s dns_recv() function.  When processing a DNS response not containing any A or AAAA records, the code erroneously left the dns_table_entry in the DNS_STATE_ASKING state after alerting the user that the resolution had failed. Subsequently, when the DNS resolver timer (dns_tmr()) fired, code in dns_check_entry() would attempt to retry the request.  By this time, however, the UDP PCB associated with the dns_table_entry had been released, leading to a bad pointer dereference

The fix is to set the state to DNS_STATE_UNUSED after it has been determined that no further retries are possible.

### DIFF
--- a/third_party/lwip/repo/lwip/src/core/dns.c
+++ b/third_party/lwip/repo/lwip/src/core/dns.c
@@ -1378,6 +1378,7 @@ dns_recv(void *arg, struct udp_pcb *pcb, struct pbuf *p, const ip_addr_t *addr, 
                 entry->state = DNS_STATE_ASKING;
               } else {
                 dns_call_found(i, NULL);
+                dns_table[i].state = DNS_STATE_UNUSED;
               }
             }
             goto memerr;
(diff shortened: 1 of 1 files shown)


third_party/lwip/repo/lwip/src/core/dns.c
```diff
@@ -1518,6 +1518,7 @@ dns_recv(void *arg, struct udp_pcb *pcb, struct pbuf *p, const ip_addr_t *addr,
             entry->state = DNS_STATE_ASKING;
           } else {
             dns_call_found(i, NULL);
+            dns_table[i].state = DNS_STATE_UNUSED;
```
robszewczyk: There are a number of other places where we see a similar pattern, i.e. we make a call to `dns_call_found(i, NULL)` without changing the state of the `dns_table`.  Most similar is in line 1398 above, but also in 1380.  Are those code paths subject to the same issue, and should they result in cleaning up the state of the DNS table?

PR #129, opened 2019-01-15
Bug/repo cleanups

robszewczyk:
Misc checks for issues reported by coverity


robszewczyk:
> Pretty sure we do not want to check in those repo changes, right?

Agreed, that should be fixed now.

PR #135, opened 2019-01-23
Only restart WRMP timer if needed

conormoneill:
This change reduces flooding of the Weave tasks' message queue on
NLER-based platforms in some circumstances. In NLER if a timer has
expired and posted its timer event but that event has not yet been
processed by the destination task (Weave task), and that timer is
then cancelled, the event will remain in the destination task's event
queue until the destination task dequeues it (and subsequently ignores
it, as it has been cancelled). This can lead to flooding of the message
queue by timer messages if we experince a burst of incoming traffic that
rapidly toggles the WRMP timer on and off. This is especially bad in
cases where we are overdue for a WRMP timer expiry, when every timer
toggle is an immediate expiry.

### DIFF
--- a/src/lib/core/WeaveExchangeMgr.cpp
+++ b/src/lib/core/WeaveExchangeMgr.cpp
@@ -104,6 +104,8 @@ WEAVE_ERROR WeaveExchangeManager::Init(WeaveMessageLayer *msgLayer)
     memset(RetransTable, 0, sizeof(RetransTable));
 
     mWRMPTimeStampBase = System::Timer::GetCurrentEpoch();
+
+    mWRMPCurrentTimerExpiry = 0;
 #endif
 
     State = kState_Initialized;
@@ -1726,14 +1728,13 @@ void WeaveExchangeManager::WRMPStartTimer()
         System::Timer::Epoch currentTime = System::Timer::GetCurrentEpoch();
         int32_t timerArmValue = nextWakeTime * mWRMPTimerInterval - (currentTime - mWRMPTimeStampBase);
         System::Timer::Epoch timerExpiryEpoch = currentTime + timerArmValue;
-        System::Timer::Epoch prevTimerExpiryEpoch = 0;
 
 #if defined(WRMP_TICKLESS_DEBUG)
         WeaveLogProgress(ExchangeManager, "WRMPStartTimer wake in %d ms (%" PRIu64" %u %" PRIu64 " %" PRIu64 ")",
                 timerArmValue,
                 timerExpiryEpoch, nextWakeTime, currentTime, mWRMPTimeStampBase);
 #endif
-        if (timerExpiryEpoch != prevTimerExpiryEpoch)
+        if (timerExpiryEpoch != mWRMPCurrentTimerExpiry)
         {
             // If the tick boundary has expired in the past (delayed processing of event due to other system activity),
             // expire the timer immediately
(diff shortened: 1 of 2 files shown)


src/lib/core/WeaveExchangeMgr.cpp
```diff
@@ -1728,28 +1723,45 @@ void WeaveExchangeManager::WRMPStartTimer()
...
 
-        VerifyOrDieWithMsg(res == WEAVE_NO_ERROR, ExchangeManager, "Cannot start WRMPTimeout\n");
 #if defined(WRMP_TICKLESS_DEBUG)
+            WeaveLogProgress(ExchangeManager, "WRMPStartTimer set timer for %d %" PRIu64, timerArmValue, timerExpiryEpoch);
+#endif
+            WRMPStopTimer();
+            res = MessageLayer->SystemLayer->StartTimer((uint32_t)timerArmValue, WRMPTimeout, this);
+
+            VerifyOrDieWithMsg(res == WEAVE_NO_ERROR, ExchangeManager, "Cannot start WRMPTimeout\n");
+            prevTimerExpiryEpoch = timerExpiryEpoch;
```
robszewczyk: Is this dead store?

src/lib/core/WeaveExchangeMgr.cpp
```diff
@@ -1728,28 +1723,45 @@ void WeaveExchangeManager::WRMPStartTimer()
 
     if (foundWake) {
         // Set timer for next tick boundary - subtract the elapsed time from the current tick
-        int32_t timerArmValue = nextWakeTime * mWRMPTimerInterval - (System::Timer::GetCurrentEpoch() - mWRMPTimeStampBase);
+        System::Timer::Epoch currentTime = System::Timer::GetCurrentEpoch();
+        int32_t timerArmValue = nextWakeTime * mWRMPTimerInterval - (currentTime - mWRMPTimeStampBase);
+        System::Timer::Epoch timerExpiryEpoch = currentTime + timerArmValue;
+        System::Timer::Epoch prevTimerExpiryEpoch = 0;
```
robszewczyk: Based on what I'm reading here, `prevTimerExpiryEpoch` should perhaps be moved to be a member variable?

jaylogue:
Can you elaborate on how a flood of timer events can be generated?  Why does rapidly toggling the WRMP timer on and off generate events if that timer never fires?

conormoneill:
> Can you elaborate on how a flood of timer events can be generated? Why does rapidly toggling the WRMP timer on and off generate events if that timer never fires?

This can happen when the system is busy and we miss a WRMP deadline.  When we miss a deadline we request a timeout of length 0, ie immediate expiry.  When NLER sees this it posts a message to the destination task immediately.  So now we have a WRMP timeout event sitting in the Weave task's queue waiting to be dequeued and processed.  Now assume there are some other previously existing messages in front of the timer event, and processing those causes the WRMP timer to be stopped and started again.  When the timer is stopped we do not remove the timeout event from the recipient task's queue, we just set a flag indicating that timer is cancelled (it will be ignored upon dispatch when the task attempts to process that event).  We then start the timer again by posting another event - now there are two pending events in the Weave task's queue.  In this circumstance, for every toggle of the WRMP timer we will accrue another event in the Weave task that will only be evicted once it is dequeued.  Usually this is ok, but in the event that a higher priority task delivers us a large batch of Weave messages all at once, the pressure on the event queue can become significant as we process those messages.  This change attempts to alleviate that pressure by not messing with the timer unless we need to change the timeout.

PR #162, opened 2019-02-23
dns test for lwip and socket

jenniexie:
1. add dns resolution test for lwip and socket
2. make file change for inet layer dns test


jenniexie:
It seems test is failing from travis CI test, I will disable the PR for now, and get it tested from my repo first on travis CI

gerickson:
Overall, the changes here look correct.

jenniexie:
I have fixed my test fail issue by copy the happy/bin files to /usr/local/bin
all tests should pass now.
I have verified from my repo, all tests are passing now
https://travis-ci.org/jenniexie/openweave-core/jobs/500477311

src/test-apps/happy/test-templates/WeaveInetDNS.py
```diff
@@ -0,0 +1,150 @@
...
+                print hred("failed")
+            else:
+                print hgreen("passed")
+                result = True
+        return (result, output)
+    
+
+    def __start_node_dnscheck(self):
+        """
+        lwip and socket use different command for now
```
gerickson: Is there much more that can be reconciled here? These remaining special cases all look very LwIP-specific.

src/test-apps/happy/test-templates/WeaveInetDNS.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __process_results(self, output):
+        """
+        process test output from DNS resolution test
+        """
+        
+        print "test DNS resolution output=====>"
+        print output
+
+        for line in output.split("\n"):
```
gerickson: We shouldn't screen-scrape to determine success or failure. We should, as with TestInetLayer and TestInetLayerMulticast, use the test program return status to indicate success or failure: 0 (EXIT_SUCCESS) = success; 1 (EXIT_FAILURE) = failure.

src/test-apps/happy/test-templates/WeaveInetDNS.py
```diff
@@ -0,0 +1,150 @@
...
+    
+    def __init__(self, opts = options):
+        HappyNode.__init__(self)
+        HappyNetwork.__init__(self)
+        WeaveTest.__init__(self)
+
+        self.quiet = opts["quiet"]
+        self.node_id = opts["node_id"]
+        self.tap_if = options["tap_if"]
+        self.node_ip = options["node_ip"]
```
gerickson: See prior comments. We should to derive this from a prefix or just extract this from the topology.

.travis/before_install.sh
```diff
@@ -120,12 +120,23 @@ case "${BUILD_TARGET}" in
 
         cd $HOME
         git clone https://github.com/openweave/happy.git
-
+        echo "=========>copy happy/bin files======>"
```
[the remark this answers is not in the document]
jenniexie: it is because make install also does install happy package in system's python env and create some folder as root permission under happy folder.
when it comes to install happy in the virtual env, it will have permission issue for virtual env to install happy.

src/test-apps/happy/tests/standalone/inet/topology/two_nodes_on_wifi.sh
```diff
@@ -0,0 +1,34 @@
...
+happy-network-address --id Home --add 10.0.1.0
+happy-node-add node01
+happy-node-add --ap onhub
+happy-node-join onhub Home
+happy-node-join node01 Home
+happy-network-route --prefix 10.0.1.0 Home onhub
+# happy-internet --node onhub --interface eno1 --isp eno --seed 249
+happy-internet --node onhub --interface ${IFACE} --isp ${IFACE//[0-9]/} --seed 249
+happy-dns 8.8.8.8
+# happy-shell node01
```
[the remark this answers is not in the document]
jenniexie: OK, will do

src/test-apps/happy/tests/standalone/inet/topology/two_nodes_on_tap_wifi.sh
```diff
@@ -0,0 +1,33 @@
...
+happy-network-add Home wifi
+happy-network-address --id Home --add 10.0.1.0
+happy-node-add node01
+happy-node-add --ap onhub
+happy-node-join onhub Home
+happy-node-join --tap node01 Home
+happy-network-route --prefix 10.0.1.0 Home onhub
+# happy-internet --node onhub --interface eno1 --isp eno --seed 249
+happy-internet --node onhub --interface ${IFACE} --isp ${IFACE//[0-9]/} --seed 249
+# happy-shell node01
```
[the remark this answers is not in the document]
jenniexie: will do

.travis/before_install.sh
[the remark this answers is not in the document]
gerickson: Do we strictly need virtualenv? This all seems to run fine without virtualenv on my Ubuntu machine and, from what I can see, Travis CI instances are mostly Debian installations.

If this is necessary, it seems like this part of our before_install.sh script (or somewhere else) in our system needs some heavier documentation as to these nuanced exceptions from practice documented elsewhere. Otherwise, this all ends up seeming like black magic or tribal knowledge to the uninitiated.

src/test-apps/happy/tests/standalone/inet/test_inet_dns_01.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self.__run_inet_dns_test("node01")
+
+
+    def __run_inet_dns_test(self, node):
+        options = WeaveInetDNS.option()
+        options["quiet"] = False
+        options["node_id"] = node
+        options["tap_if"] = "wlan0"
+        options["node_ip"] = "10.0.1.3"
```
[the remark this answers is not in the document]
jenniexie: hmm??

src/test-apps/happy/tests/standalone/inet/test_inet_dns_01.py
```diff
@@ -0,0 +1,92 @@
...
+                "/topology/two_nodes_on_tap_wifi.sh"
+            output = subprocess.call([topology_shell_script])
+        else:
+            topology_shell_script = os.path.dirname(os.path.realpath(__file__)) + \
+                "/topology/two_nodes_on_wifi.sh"
+            output = subprocess.call([topology_shell_script])
+
+    def tearDown(self):
+        # cleaning up
+        subprocess.call(["happy-state-delete"])
```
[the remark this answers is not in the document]
jenniexie: I was thinking from test case perspective, after test done, we should do clean up. we can discuss this one

src/test-apps/happy/test-templates/WeaveInetDNS.py
jenniexie: RS talked about to get parameters for LWIP command to use some util that Jay had before
I will need to figure out that with Jay and RS
Is it OK that I create a issue for this one, and address this one in future?

src/test-apps/happy/test-templates/WeaveInetDNS.py
jenniexie: ??I will check with you for this one

src/test-apps/happy/test-templates/WeaveInetDNS.py
gerickson: My point was rather than I think we can get rid of the TODO comment here. I don't think there's any further refactoring work at this particular point.

src/test-apps/happy/tests/standalone/inet/test_inet_dns_01.py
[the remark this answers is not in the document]
gerickson: I agree that we should clean up. That's what tests that load the JSON file do at test conclusion; however, that happens in the framework; the individual unit tests don't have to do it themselves.

So, relative to our meeting a few weeks back about Happy changes, this is one place where we could probably add some changes to either the Weave-based Happy test infrastructure or to Happy itself. Does that make sense?

src/test-apps/happy/tests/standalone/inet/test_inet_dns_01.py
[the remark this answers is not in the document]
gerickson: We could specify the argument here as a prefix and then call getNodeAddressesOnPrefix or just call getNodePublicIPv4Address if we only expect one IPv4 address.

.travis/before_install.sh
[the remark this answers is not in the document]
jenniexie: I checked with @yunhanw-google yesterday, it is not strictly need virtualenv for travis-CI existing tests. but when it come to service tests, we need virtualenv
I agree it is black magic around here:), it took me a while to figure out the root cause of the "happy commands not found" error yesterday from Travis CI test run.
I am OK if we want to switch everything without using virtual env.
@yunhanw-google, could you explain why do we need virtualenv for service tests?

src/test-apps/happy/tests/standalone/inet/test_inet_dns_01.py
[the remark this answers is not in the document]
jenniexie: 1. getNodeAddressesOnPrefix will need prefix as argument
2. seems getNodePublicIPv4Address will not work for this case, because public ip need network type as "WAN"
seems getNodeInterfaceAddresses() will work better? I will test it out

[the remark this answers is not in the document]
gerickson: That `getNodeAddressesOnPrefix` needs an argument should not pose a problem. We can pass the prefix as an argument. The topology is known and fixed for this test and the prefix is fixed for the topology.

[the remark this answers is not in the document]
jenniexie: sure, will pass in prefix as argument then

jenniexie:
@gerickson should fixed all comments now, please take a look again
FYI: I removed BLUEZ test from scri

.travis/script.sh
```diff
@@ -40,7 +40,7 @@ case "${BUILD_TARGET}" in
 
     linux-auto-gcc-check-happy)
         # run happy test
-        sudo bash -c "source ${HOME}/ve/happy/bin/activate; make -f Makefile-Standalone DEBUG=1 TIMESTAMP=1 COVERAGE=1 BuildJobs=24 BLUEZ=1 check"
```
gerickson: Why did we remove BLUEZ here? Do the tests no longer work or run?

src/test-apps/happy/test-templates/WeaveInetDNS.py
```diff
@@ -0,0 +1,153 @@
...
+        status = (results['status'] == 0)
+
+        output = results['output']
+
+        return (status, output)
+
+    def __start_node_dnscheck(self):
+        """
+        lwip and socket use different command for now
+        TODO: fix lwip and socket to use similar command future.
```
gerickson: We should get rid of this "TODO" since, as we discussed, the problem is a general one to resolve across **all** of our tests and is not specific to this test.

src/test-apps/happy/tests/standalone/inet/test_inet_dns_01.py
```diff
@@ -0,0 +1,97 @@
...
+        options = WeaveInetDNS.option()
+        options["quiet"] = False
+        options["node_id"] = node
+        options["tap_if"] = "wlan0"
+        options["ipv4_gateway"] = "10.0.1.2"
+        options["dns"] = "8.8.8.8"
+        options["use_lwip"] = self.using_lwip
+
+        weave_inet_dns = WeaveInetDNS.WeaveInetDNS(options)
+        weave_inet_dns.node_ip = weave_inet_dns.getNodeAddressesOnPrefix("10.0.1", node)[0]
```
gerickson: I think, as we do with the other TestInet* tests, I'd pass prefix as an 'options["prefix"]' parameter and then perform this lookup / resolution in WeaveInetDNS.py.

src/test-apps/happy/test-templates/WeaveInetDNS.py
jenniexie: oops, sorry, missed this one:)

.travis/script.sh
jenniexie: The bluez test is failing after change to non virtual python env in setup part, I need to debug and fix it.
I have created an issue to track, I will address it in a different PR.
https://github.com/openweave/openweave-core/issues/166

src/test-apps/happy/tests/standalone/inet/test_inet_dns_01.py
jenniexie: OK, I see.

.travis/script.sh
gerickson: OK, thanks.

src/test-apps/happy/test-templates/WeaveInetDNS.py
```diff
@@ -0,0 +1,155 @@
...
+
+        return (status, output)
+
+    def __start_node_dnscheck(self):
+        """
+        lwip and socket use different command for now
+        """
+        cmd = "sudo "
+        cmd += self.getWeaveInetLayerDNSPath()
+        node_ip = self.getNodeAddressesOnPrefix(self.prefix, self.node_id)[0]
```
gerickson: We should probably check that this is a non-NULL, valid result as we do with other tests.

gerickson:
Please note that all checks are still failing.

PR #165, opened 2019-03-04
Redesign how TraitCatalog is keeping elements.

lbienczyk:
Redesign the way elements are removed. Instead of erasing an element from the catalog,
only set trait pointer to NULL, so DataHandle order would remain always the same.

### DIFF
--- a/src/lib/profiles/data-management/Current/NotificationEngine.cpp
+++ b/src/lib/profiles/data-management/Current/NotificationEngine.cpp
@@ -1680,13 +1680,15 @@ WEAVE_ERROR NotificationEngine::BuildSubscriptionlessNotification(PacketBuffer *
         TraitDataHandle traitHandle = currPath->mTraitDataHandle;
 
         // Get the max version from the datasource
-        err = pubCatalog->Locate(traitHandle, &dataSource);
-        SuccessOrExit(err);
+        // Locate() can return an error if the sink has been removed from the catalog. In that case,
+        // skip this path
+        if (pubCatalog->Locate(traitHandle, &dataSource) == WEAVE_NO_ERROR)
+        {
+            schemaVersion = dataSource->GetSchemaEngine()->GetMaxVersion();
 
-        schemaVersion = dataSource->GetSchemaEngine()->GetMaxVersion();
-
-        err = notifyRequest.WriteDataElement(traitHandle, kRootPropertyPathHandle, schemaVersion, NULL, 0, NULL, 0);
-        SuccessOrExit(err);
+            err = notifyRequest.WriteDataElement(traitHandle, kRootPropertyPathHandle, schemaVersion, NULL, 0, NULL, 0);
+            SuccessOrExit(err);
+        }
     }
 
     err = notifyRequest.MoveToState(kNotifyRequestBuilder_Idle);
--- a/src/lib/profiles/data-management/Current/SubscriptionClient.cpp
+++ b/src/lib/profiles/data-management/Current/SubscriptionClient.cpp
@@ -91,7 +91,6 @@ void SubscriptionClient::Reset(void)
 #if WEAVE_CONFIG_ENABLE_WDM_UPDATE
     mUpdateMutex                            = NULL;
     mUpdateInFlight                         = false;
-    mNumUpdatableTraitInstances             = 0;
     mMaxUpdateSize                          = 0;
     mUpdateRequestContext.Reset();
     mPendingSetState = kPendingSetEmpty;
@@ -143,7 +142,6 @@ WEAVE_ERROR SubscriptionClient::Init(Binding * const apBinding, void * const apA
 #if WEAVE_CONFIG_ENABLE_WDM_UPDATE
     mUpdateMutex                            = aUpdateMutex;
     mUpdateInFlight                         = false;
-    mNumUpdatableTraitInstances             = 0;
     mMaxUpdateSize                          = 0;
 
 #endif // WEAVE_CONFIG_ENABLE_WDM_UPDATE
@@ -156,10 +154,7 @@ WEAVE_ERROR SubscriptionClient::Init(Binding * const apBinding, void * const apA
     err = mUpdateClient.Init(mBinding, this, UpdateEventCallback);
     SuccessOrExit(err);
 
-    if (NULL != mDataSinkCatalog)
-    {
-        mDataSinkCatalog->Iterate(InitUpdatableSinkTrait, this);
-    }
+    ConfigureUpdatableSinks();
 
 exit:
 #endif // WEAVE_CONFIG_ENABLE_WDM_UPDATE
@@ -551,8 +546,12 @@ WEAVE_ERROR SubscriptionClient::SendSubscribeRequest(void)
                         outSubscribeParam.mSubscribeRequestPrepareNeeded.mPathList[i].mPropertyPathHandle;
                 }
 
-                err = mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink);
-                SuccessOrExit(err);
+                if (mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink) != WEAVE_NO_ERROR)
+                {
+                    // Locate() can return an error if the sink has been removed from the catalog. In that case,
+                    // continue to next entry
+                    continue;
+                }
 
                 // Start the TLV Path
                 err = writer.StartContainer(nl::Weave::TLV::AnonymousTag, nl::Weave::TLV::kTLVType_Path, dummyContainerType);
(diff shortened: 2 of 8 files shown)


src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -551,8 +546,11 @@ WEAVE_ERROR SubscriptionClient::SendSubscribeRequest(void)
                         outSubscribeParam.mSubscribeRequestPrepareNeeded.mPathList[i].mPropertyPathHandle;
                 }
 
-                err = mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink);
-                SuccessOrExit(err);
+                if (mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink) != WEAVE_NO_ERROR)
+                {
+                    // The application can remove Sinks from the catalog. Ignore this path.
```
pidarped: Nit: Rephrase as 
"Locate() can return an error if the sink has been removed from the catalog. In that case, continue to next entry".

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -561,6 +559,14 @@ WEAVE_ERROR SubscriptionClient::SendSubscribeRequest(void)
                 // Start, fill, and close the TLV Structure that contains ResourceID, ProfileID, and InstanceID
                 err = mDataSinkCatalog->HandleToAddress(versionedTraitPath.mTraitDataHandle, writer,
                                                         versionedTraitPath.mRequestedVersionRange);
+
+                if (err == WEAVE_ERROR_INVALID_ARGUMENT)
+                {
+                    // we should not reach this code, Locate would fail before
```
pidarped: Rephrase comment as:
" Ideally, this code will not be reached as HandleToAddress() should find the entry in the catalog as, otherwise, the earlier Locate() call would have continued. However, keeping this check here for consistency and code safety".

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -596,8 +602,11 @@ WEAVE_ERROR SubscriptionClient::SendSubscribeRequest(void)
                         outSubscribeParam.mSubscribeRequestPrepareNeeded.mPathList[i].mPropertyPathHandle;
                 }
 
-                err = mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink);
-                SuccessOrExit(err);
+                if (mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink) != WEAVE_NO_ERROR)
+                {
+                    // The application can remove Sinks from the catalog. Ignore this path.
```
pidarped: Same comment as line 551.

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -2186,7 +2194,14 @@ void SubscriptionClient::PurgeAndNotifyFailedPaths(WEAVE_ERROR aErr, TraitPathSt
 
             aPathStore.GetItemAt(j, traitPath);
             updatableDataSink = Locate(traitPath.mTraitDataHandle, mDataSinkCatalog);
-            VerifyOrDie(updatableDataSink != NULL);
+
+            if (updatableDataSink == NULL)
+            {
+                // Note: the application can remove sinks from the catalog
```
pidarped: nit: Rephrase as
"Locate() can return NULL if the datasink has been removed from the catalog. In that case, remove the stale item in the path store and continue to next entry".

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -2516,7 +2531,15 @@ void SubscriptionClient::OnUpdateResponse(WEAVE_ERROR aReason, nl::Weave::Profil
         mInProgressUpdateList.GetItemAt(j, traitPath);
 
         updatableDataSink = Locate(traitPath.mTraitDataHandle, mDataSinkCatalog);
-        VerifyOrExit(updatableDataSink != NULL, err = WEAVE_ERROR_WDM_SCHEMA_MISMATCH);
+
+        if (updatableDataSink == NULL)
+        {
+            WeaveLogDetail(DataManagement, "item: %zu, traitDataHandle: % potentially removed from the catalog" PRIu16 ", pathHandle: %" PRIu32 "",
+                    j, traitPath.mTraitDataHandle, traitPath.mPropertyPathHandle);
+            // The application can remove Sinks from the catalog. Ignore this path.
```
pidarped: Similar comment as line 2200.

src/lib/profiles/data-management/Current/SubscriptionEngine.cpp
```diff
@@ -499,11 +499,22 @@ WEAVE_ERROR SubscriptionEngine::ProcessDataList(nl::Weave::TLV::TLVReader & aRea
         SchemaVersionRange versionRange;
 
         err = aCatalog->AddressToHandle(pathReader, handle, versionRange);
-        SuccessOrExit(err);
 
-        err = aCatalog->Locate(handle, &dataSink);
+        if (err == WEAVE_ERROR_INVALID_PROFILE_ID)
+        {
+            // Note: the application can remove sinks from the catalog
```
pidarped: nit: Rephrase as
"AddressToHandle() can return an error if the datasink for the handle has been removed from the catalog. In that case, continue to next entry".

src/lib/profiles/data-management/Current/NotificationEngine.cpp
```diff
@@ -1680,13 +1680,15 @@ WEAVE_ERROR NotificationEngine::BuildSubscriptionlessNotification(PacketBuffer *
         TraitDataHandle traitHandle = currPath->mTraitDataHandle;
 
         // Get the max version from the datasource
-        err = pubCatalog->Locate(traitHandle, &dataSource);
-        SuccessOrExit(err);
-
-        schemaVersion = dataSource->GetSchemaEngine()->GetMaxVersion();
+        // Locate() can return an error if the sink has been removed from the catalog. In that case,
+        // skip this path
+        if (pubCatalog->Locate(traitHandle, &dataSource) == WEAVE_NO_ERROR)
```
jaylogue: I would prefer that this code not ignore all errors, only the error indicating that the data source wasn't found.

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -551,8 +546,12 @@ WEAVE_ERROR SubscriptionClient::SendSubscribeRequest(void)
                         outSubscribeParam.mSubscribeRequestPrepareNeeded.mPathList[i].mPropertyPathHandle;
                 }
 
-                err = mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink);
-                SuccessOrExit(err);
+                if (mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink) != WEAVE_NO_ERROR)
+                {
```
jaylogue: Prefer better error handling.

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -561,6 +560,16 @@ WEAVE_ERROR SubscriptionClient::SendSubscribeRequest(void)
                 // Start, fill, and close the TLV Structure that contains ResourceID, ProfileID, and InstanceID
                 err = mDataSinkCatalog->HandleToAddress(versionedTraitPath.mTraitDataHandle, writer,
                                                         versionedTraitPath.mRequestedVersionRange);
+
+                if (err == WEAVE_ERROR_INVALID_ARGUMENT)
```
jaylogue: If this error is truly unexpected, then this test should really be replaced with a SuccessOrExit().

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -596,8 +605,12 @@ WEAVE_ERROR SubscriptionClient::SendSubscribeRequest(void)
                         outSubscribeParam.mSubscribeRequestPrepareNeeded.mPathList[i].mPropertyPathHandle;
                 }
 
-                err = mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink);
-                SuccessOrExit(err);
+                if (mDataSinkCatalog->Locate(versionedTraitPath.mTraitDataHandle, &pDataSink) != WEAVE_NO_ERROR)
```
jaylogue: Prefer better error handling.

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -2085,13 +2098,17 @@ WEAVE_ERROR SubscriptionClient::MoveInProgressToPending(void)
...
         if ( ! mInProgressUpdateList.AreFlagsSet(i, kFlag_Private))
         {
-            err = mDataSinkCatalog->Locate(traitPath.mTraitDataHandle, &dataSink);
-            SuccessOrExit(err);
-            err = AddItemPendingUpdateSet(traitPath, dataSink->GetSchemaEngine());
-            SuccessOrExit(err);
-            mInProgressUpdateList.RemoveItemAt(i);
+            // Locate() can return an error if the sink has been removed from the catalog. In that case,
+            // skip this path
+            if (mDataSinkCatalog->Locate(traitPath.mTraitDataHandle, &dataSink) == WEAVE_NO_ERROR)
```
jaylogue: Prefer better error handling.

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
pidarped: The only reason I saw in keeping it the way it is now is this piece of error check is independent of the previous Locate() error handling. Otherwise, if it were replaced by a SuccessOrExit() and later, for whatever reason, if the earlier Locate() was refactored/moved then the error handling here would break. Not sure what the chances are for that. So, to keep it simple and less code change for now, we could replace it with SuccessOrExit().

Issue #172, opened 2019-03-12
Standalone Tunnel establishment should disable border gateway configuration at platform level


pidarped:
Standalone tunnel establishment should disable the platform-level border gateway configuration directly instead of relying on the restricted routing flag sent as part of the StatusReport in response to a TunnelOpen.

gerickson:
@pidarped, can this issue be closed now that 173 and 174 have been closed and merged?

PR #174, opened 2019-03-12
Standalone Tunnel establishment should disable border gateway configuration at the platform level.

pidarped:
Fix for Issue #172 

This was being achieved by the routingRestricted flag sent in the
StatusReport in response to the TunnelOpen. Instead, the
WeaveTunnelAgent should consult its configured role(Standalone) to
disable platform-level border gateway function.

### DIFF
--- a/src/test-apps/TestWeaveTunnelBR.cpp
+++ b/src/test-apps/TestWeaveTunnelBR.cpp
@@ -2310,6 +2310,75 @@ exit:
     gTunAgent.Shutdown();
 }
 
+static void TestTunnelRestrictedRoutingOnStandaloneTunnelOpen(nlTestSuite *inSuite, void *inContext)
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+
+    Done = false;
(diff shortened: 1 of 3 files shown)


src/test-apps/TestWeaveTunnelBR.cpp
```diff
@@ -2310,6 +2310,75 @@ static void TestTunnelRestrictedRoutingOnTunnelOpen(nlTestSuite *inSuite, void *
     gTunAgent.Shutdown();
 }
 
+static void TestTunnelRestrictedRoutingOnStandaloneTunnelOpen(nlTestSuite *inSuite, void *inContext)
```
robszewczyk: What is it that we are testing here? I'm not exactly seeing the tests that verify that the semantics of the standalone device are enforced, is that embedded elsewhere in the code?

robszewczyk: nevermind, I see that adding the case statement in line 2868 implements the actual test.

PR #175, opened 2019-03-13
Fix the logic to determine whether the LwIP interface supports multicast

robszewczyk:


jenniexie:
I have verified at my side with this change, it worked for both weave_time_local and weave_time_auto tests.

PR #178, opened 2019-03-14
Weave tool print-sig command

jaylogue:
-- Added a command to the weave tool for printing the contents of a Weave signature object.  Similar to print-cert, the print-sig command takes a file containing an encoded Weave signature in either raw TLV or base-64 format, and prints the fields within the signature in a human readable format.

-- Added utility functions for decoding Weave certificates and certificate-related structures without needing to load them into a WeaveCertificateSet object.

-- Added utility functions for printing various security-related constructs such as Weave signatures, distinguished names, TLV arrays of certificates, ECDSA signatures, and certificate references.

-- Simplified the implementation of the weave tool’s print-cert command by using the new utility functions described above.

-- Modified the ASN1Writer class to support a ‘null-writer’ mode.  When initialized in this mode, the user is not required to supply an output buffer, and all calls to encode ASN1 constructs are silently ignored.  This is useful in cases in cases where one must call a function that requires an ASN1Writer but the output of the writer is unneeded.

-- Enhanced the TestWeaveTool unit test script to test the print-cert command.

-- Enhanced the TestASN1 unit test to test the new null-writer functionality.


jaylogue:
Happy pairing test is failing, presumably for unrelated reasons.

src/tools/weave/Cmd_PrintSig.cpp
```diff
@@ -0,0 +1,131 @@
+/*
+ *
+ *    Copyright (c) 2013-2017 Nest Labs, Inc.
```
gerickson: This should be "Copyright (c) 2019 Google LLC.".

src/tools/weave/weave-tool.h
```diff
@@ -54,7 +54,7 @@ using namespace nl::ArgParser;
 
 using nl::Weave::ASN1::OID;
 
-#define COPYRIGHT_STRING "Copyright (c) 2013-2017 Nest Labs, Inc.\nAll rights reserved.\n"
+#define COPYRIGHT_STRING "Copyright (c) 2013-2019 Nest Labs, Inc.\nAll rights reserved.\n"
```
gerickson: This should probably be changed to:

```
#define COPYRIGHT_STRING "Copyright (c) 2019 Google LLC.\nCopyright (c) 2013-2017 Nest Labs, Inc.\nAll rights reserved.\n"
```

PR #179, opened 2019-03-15
Add cros ble indication change in weave

yunhanw-google:
Add cros ble indication change in cros since its ble indication args are different from bluez upstream.


yunhanw-google:
> Where does BLE_CONFIG_CROS_BLUEZ_FEATURE come from? I believe it should be put into an appropriate config file, documented and provided with a reasonable default value.

Thanks, just set BLE_CONFIG_CROS_BLUEZ_FEATURE as default 0 in bleconfig file.

src/ble/BleConfig.h
```diff
@@ -112,6 +112,16 @@
 #define BLE_CONFIG_BLUEZ_MTU_FEATURE 0
 #endif // BLE_CONFIG_BLUEZ_MTU_FEATURE
 
+/**
+ *  @def BLE_CONFIG_CROS_BLUEZ_FEATURE
+ *
+ *  @brief
+ *    This define if CROS BLUEZ FEATURE is enabled or not
```
jaylogue: When you say "CROS BLUEZ" do you mean ChromeOS's implementation of BlueZ?  If so, please make that clear in the description.

yunhanw-google: updated description, thanks,

PR #189, opened 2019-03-28
Fix for WEAV-3231.

pidarped:
Make sure that while processing the DataList in a Notify starting from
the root, filter checks are correctly performed while descending into
dictionaries and DictionaryReplace/Modify pre-processing functions
are not called for dictionary paths that also exist in a pending WDM
Update.


robszewczyk:
Please add test cases that test the demonstrate the erroneous behavior.

PR #190, opened 2019-03-29
On LwIP: count all interfaces that support broadcast as supporting mu…

robszewczyk:
…lticast

Rationale: The callers of this interface use it to determine whether
an interface is capable of multicast, with the common case being
link-local all-nodes.  Per spec, the MLD support is not required if the
node only supports all-nodes addess; consequently, checking for BROADCAST
support seems appropriate.


gerickson:
Is there / would there be a similar issue on BSD sockets (i.e., IFF_BROADCAST)?

robszewczyk:
@gerickson I don't know about BSDs.  This problem was surfaced by interaction with OpenThread on LwIP platforms.  SHould this problem arise, I will fix it again.

PR #194, opened 2019-04-13
Generalized Weave error-to-string facility

jaylogue:
-- Re-implemented Weave’s ErrorStr() function to use a modular design where the work of translating an error to string form is delegated to a series of formatting functions which can be registered at runtime.  The new design makes it possible for error formatting functionality to be implemented separately for each category of error.  Additionally, adapters of Weave code can install their own error-to-string translations for errors that are unique to a particular platform or execution environment.

Once landed, this feature will be used to provide SDK-specific error translations for the nRF52840 and ESP32 platforms.

-- Using the above facility, error translations for the various OpenWeave-defined error categories were moved out of ErrorStr.cpp and into their own files.

-- Minor adjustments were made to the output format of error strings to make them more consistent and more informative.  Notably, long-format error strings now include the error code in both decimal and hex format.

-- Reduced memory usage when using long-format error strings by decreasing the size of the static buffer used to error strings from 1000 to 256 bytes.  The new size seems to be more than adequate for the existing crop of error translations.


src/ble/BleError.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#if CONFIG_NETWORK_LAYER_BLE
+
+#include <BleLayer/BleLayer.h>
+#include <BleLayer/BleError.h>
+
+#include <Weave/Support/ErrorStr.h>
+
+
+namespace nl {
```
gerickson: This gets copy-and-pasted in each *Error.cpp file. Any reason to not create a header declaration for this in ErrorStr.h?

src/inet/InetError.cpp
```diff
@@ -0,0 +1,112 @@
...
+ *      This file contains functions for working with Inet Layer errors.
+ */
+
+#include <InetLayer/Inet.h>
+#include <InetLayer/InetError.h>
+
+#include <Weave/Support/ErrorStr.h>
+
+
+namespace nl {
```
gerickson: See previous.

gerickson:
One minor bit of feedback; however, overall, this looks good.

robszewczyk:
The main pitfall with this approach is that the application has to remember to register all the error formatters that it needs to use.

jaylogue:
> The main pitfall with this approach is that the application has to remember to register all the error formatters that it needs to use.

I think I have that mostly covered for all the existing formatters.  Weave and ASN1 formatters are "pre-registered".  Ble Layer, Inet Layer, System Layer, LwIP and POSIX formatters get registered when you initialize the respective layer objects.

jaylogue:
Review comments addressed.

robszewczyk:
> > The main pitfall with this approach is that the application has to remember to register all the error formatters that it needs to use.
> 
> I think I have that mostly covered for all the existing formatters. Weave and ASN1 formatters are "pre-registered". Ble Layer, Inet Layer, System Layer, LwIP and POSIX formatters get registered when you initialize the respective layer objects.

I caught some of those, and missed others, that's what inspired the comment.  I then realized that even if this were left to the app developer, it would still work even in the presence of multiple registration of the same formatter (though at an expense of a few extra checks).

jaylogue:
>  it would still work even in the presence of multiple registration of the same formatter (though at an expense of a few extra checks

Yep, and the "extra" check occurs at registration time, as RegisterErrorFormatter() won't add a formatter to the list that points to the same format function as an existing one.

PR #195, opened 2019-04-15
Give event ids from subscription request to the app

Glen-Weaver:
Weave provides to the application the event ids send down as part of the
subscription request.

A new method is added to PersistentCounter that allows the counter to be
advanced past a given value (event id).


robszewczyk:
Minor comment: it is useful to keep the title of the PR short so that the `git log --oneline` looks good and the pull request is formatted correctly.

PR #196, opened 2019-04-16
Enhance the API of the ServiceDirectory to produce a HostPortList

robszewczyk:
The previous `lookup` method was not terribly useful -- it exposed the
internal structures in the ServiceDirectory, and it was not entirely
clear how to use it properly.  This commit refactors the code such
that:

* we provide a variant of the `lookup` method that populates a
  HostPortList.

* we use the code internally to implement the code paths inside
  `lookupAndConnect`

### DIFF
--- a/src/lib/profiles/service-directory/ServiceDirectory.cpp
+++ b/src/lib/profiles/service-directory/ServiceDirectory.cpp
@@ -61,6 +61,19 @@
 
 #if WEAVE_CONFIG_ENABLE_SERVICE_DIRECTORY
 
+#if HAVE_NEW
+#include <new>
+#else
+inline void * operator new(size_t, void * p) throw()
+{
+    return p;
+}
+inline void * operator new[](size_t, void * p) throw()
+{
+    return p;
+}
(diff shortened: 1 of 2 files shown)


src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -538,13 +551,61 @@ WEAVE_ERROR WeaveServiceManager::connect(uint64_t aServiceEp,
     return err;
 }
 
+/**
+ * @brief This method looks up directory information for a service endpoint.
+ *
+ * If the service directory has been resolved, i.e. if there has been a
+ * successful connect() operation, then this method will populate the supplied
+ * HostPortList object.
+ *
```
jaylogue: Maybe add a note that the returned HostPortList contains pointers into the service directory cache, and thus only remains valid until the cache is cleared or another service directory lookup occurs.

robszewczyk: Done.

PR #201, opened 2019-04-22
Fixed a bug in ServiceProvisioningServer that was causing timeouts.

suryanshup:
-- Before a pair device to account request can be sent out, the service
   provisioning server would check to see if a tunnel has been established
   with the service. Service Provisioning would always fail on platforms
   such as the nRF5 since it doesn't have the capability to establish a tunnel
   directy with the service.

-- Solution: Add logic to gate the provisioning request only when service
   connectivity AND service tunnel connection have NOT been established.


suryanshup:
Tested service provisioning on nrf52840 and esp32.

PR #203, opened 2019-04-23
Improved support for project-specific configuration files

jaylogue:
-- Modified the behavior of the --with-weave-...-project-includes configure options to allow application developers to specify a full header file name, rather than just a directory name.  This makes it possible, for example, to use a single project-specific config file to override configuration for multiple OpenWeave components.

As part of this change, the underlying NL_CHECK_PROJECT_CONFIG_INCLUDES macro was modified to NOT use AC_CHECK_HEADERS when checking for the specified header file.  This reflects the fact that the supplied project header file may not be compilable outside of the context of a full OpenWeave build.


jaylogue:
Hey @robszewczyk, here's the change that I need tested.  Thanks.

jaylogue:
Commit updated with new bootstrapped files.

PR #205, opened 2019-04-27
Make LoggingManagent::CreateLoggingManagement() API clearer and more …

robszewczyk:
…robust

Previously, the API (and associated set of constructors) took a series
of arrays to initialize the LoggingManagement object.  All arrays were
supposed to be provided in the order from Debug to CriticalProduction
importance.  The code required 5 different arrays.  The ordering
required was the opposite of the numerical ordering of the importance
enum.  This led to confusion.  The new API replaces the prior arrays
with a single array of objects, each object specifies all required
storage and associated objects per importance level.  The API
documentation unabiguously states that the objects are passed in the
numerical order of importance, and the API removes the flexibility to
omit buffers per importance levels.  Finally, the API is friendly
towards flash based structures, which should allow constrained
platforms to create the initialization arguments in flash.

### DIFF
--- a/src/lib/profiles/data-management/Current/LoggingManagement.cpp
+++ b/src/lib/profiles/data-management/Current/LoggingManagement.cpp
@@ -62,7 +62,7 @@ namespace WeaveMakeManagedNamespaceIdentifier(DataManagement, kWeaveManagedNames
 
 // Static instance: embedded platforms not always implement a proper
 // C++ runtime; instead, the instance is initialized via placement new
-// in CreateLoggingManangement.
+// in CreateLoggingManagement.
 
 static LoggingManagement sInstance;
 
@@ -400,52 +400,29 @@ void LoggingManagement::SkipEvent(EventLoadOutContext * aContext)
 }
 
 /**
- * @brief Create and initialize the logging management buffers. Must
- *   be called prior to the logging being used.
+ * @brief Create LoggingManagement object and initialize the logging management
+ *   subsystem with provided resources.
+ *
+ * Initialize the LoggingManagement with an array of LogStorageResources.  The
+ * array must provide a resource for each valid importance level, the elements
+ * of the array must be in increasing numerical value of importance (and in
+ * decreasing importance); the first element in the array corresponds to the
+ * resources allocated for the most critical events, and the last element
+ * corresponds to the least important events.
  *
  * @param[in] inMgr         WeaveExchangeManager to be used with this logging subsystem
  *
- * @param[in] inNumBuffers  Number of buffers to use for event storage
+ * @param[in] inNumBuffers  Number of elements in inLogStorageResources array
  *
- * @param[in] inBufferLengths Description of inBufferLengths
+ * @param[in] inLogStorageResources  An array of LogStorageResources for each importance level.
  *
- * @param[in] inBuffers     The buffers to use for actual event logging.
- * @param[in] inCounterKeys Keys naming persisted counters
- *
- * @param[in] inCounterEpochs An array of epochs for each of the persisted counters
- *
- * @param[in] inCounterStorage Application-provided storage for the persistent counters
+ * @note This function must be called prior to the logging being used.
  */
-void LoggingManagement::CreateLoggingManagement(nl::Weave::WeaveExchangeManager * inMgr, size_t inNumBuffers,
-                                                size_t * inBufferLengths, void ** inBuffers,
-                                                nl::Weave::Platform::PersistedStorage::Key * inCounterKeys,
-                                                const uint32_t * inCounterEpochs, nl::Weave::PersistedCounter ** inCounterStorage)
+void LoggingManagement::CreateLoggingManagement(nl::Weave::WeaveExchangeManager * inMgr,
+                                                size_t inNumBuffers,
+                                                const LogStorageResources * const inLogStorageResources)
 {
-    new (&sInstance)
-        LoggingManagement(inMgr, inNumBuffers, inBufferLengths, inBuffers, inCounterKeys, inCounterEpochs, inCounterStorage);
-}
-
-/**
- * @brief Create and initialize the logging management buffers. Must
- *   be called prior to the logging being used.
- *
- * @param[in] inMgr         WeaveExchangeManager to be used with this logging subsystem.
- *
- * @param[in] inNumBuffers  Number of buffers to use for event storage.
- *
- * @param[in] inBufferLengths Description of inBufferLengths.
- *
- * @param[in] inBuffers     The buffers to use for actual event logging.
- *
- * @param[in] nWeaveCounter The array of counter pointers must contain the initialized counters, and has to contain inNumBuffers of
- * counters.
- *
- */
-void LoggingManagement::CreateLoggingManagement(nl::Weave::WeaveExchangeManager * inMgr, size_t inNumBuffers,
-                                                size_t * inBufferLengths, void ** inBuffers,
-                                                nl::Weave::MonotonicallyIncreasingCounter ** nWeaveCounter)
-{
-    new (&sInstance) LoggingManagement(inMgr, inNumBuffers, inBufferLengths, inBuffers, nWeaveCounter);
+    new (&sInstance) LoggingManagement(inMgr, inNumBuffers, inLogStorageResources);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -475,59 +447,57 @@ WEAVE_ERROR LoggingManagement::SetExchangeManager(nl::Weave::WeaveExchangeManage
...
+
+LoggingManagement::LoggingManagement(nl::Weave::WeaveExchangeManager * inMgr,
+                                     const LogStorageResources * const inLogStorageResources)
 {
     CircularEventBuffer * current = NULL;
     CircularEventBuffer * prev    = NULL;
     CircularEventBuffer * next    = NULL;
-    size_t i;
+    size_t i, j;
+
```
gerickson: Style: spurious new line?

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -1783,8 +1699,7 @@ void LoggingManagement::FlushHandler(System::Layer * inSystemLayer, INET_ERROR i
 
     switch (mState)
     {
-    case kLoggingManagementState_Idle:
-    {
+    case kLoggingManagementState_Idle: {
```
gerickson: Style: new line before this line, consistent with line 1751.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -1817,8 +1732,7 @@ void LoggingManagement::FlushHandler(System::Layer * inSystemLayer, INET_ERROR i
 
         break;
     }
-    case kLoggingManagementState_Holdoff:
-    {
+    case kLoggingManagementState_Holdoff: {
```
gerickson: Style: new line before this line, consistent with line 1751.

src/lib/profiles/data-management/Current/LoggingManagement.h
```diff
@@ -146,6 +146,36 @@ enum LoggingManagementStates
...
+ */
+
+struct LogStorageResources
+{
+    constexpr LogStorageResources(void * inBuffer, size_t inBufferSize, nl::Weave::Platform::PersistedStorage::Key * inCounterKey,
+                                  uint32_t inCounterEpoch, nl::Weave::PersistedCounter * inCounterStorage) :
+        mBuffer(inBuffer),
+        mBufferSize(inBufferSize), mCounterKey(inCounterKey), mCounterEpoch(inCounterEpoch), mCounterStorage(inCounterStorage) { };
+    void * mBuffer;     //< Buffer to be used as a storage at the particular importance level and shared with more important events.
+                        //Must not be NULL.  Must be large enough to accomodate the largest event emitted by the system.
```
gerickson: Do we have tooling to check what the largest event emitted by the system might be or an assertion that fires when this "largest event emitted..." fails?

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -475,59 +447,57 @@ WEAVE_ERROR LoggingManagement::SetExchangeManager(nl::Weave::WeaveExchangeManage
...
-        next = ((i + 1) < inNumBuffers) ? static_cast<CircularEventBuffer *>(inBuffers[i + 1]) : NULL;
+        i = (kImportanceType_Last - kImportanceType_First) - j;
+
+        next = (i > 0) ? static_cast<CircularEventBuffer *>(inLogStorageResources[i - 1].mBuffer) : NULL;
 
-        new (inBuffers[i]) CircularEventBuffer(static_cast<uint8_t *>(inBuffers[i]) + sizeof(CircularEventBuffer),
-                                               inBufferLengths[i] - sizeof(CircularEventBuffer), prev, next);
+        new (inLogStorageResources[i].mBuffer)
+            CircularEventBuffer(static_cast<uint8_t *>(inLogStorageResources[i].mBuffer) + sizeof(CircularEventBuffer),
+                                inLogStorageResources[i].mBufferSize - sizeof(CircularEventBuffer), prev, next);
```
emargolis: Should we check appropriate buffer size:
    VerifyOrExit(inLogStorageResources[i].mBufferSize > sizeof(CircularEventBuffer), err = WEAVE_ERROR_BUFFER_TOO_SMALL)

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -475,59 +447,57 @@ WEAVE_ERROR LoggingManagement::SetExchangeManager(nl::Weave::WeaveExchangeManage
...
- *
- * @param[in] inCounterStorage Application-provided storage for the persistent counters
- *
- * @return LoggingManagement
  */
-LoggingManagement::LoggingManagement(WeaveExchangeManager * inMgr, size_t inNumBuffers, size_t * inBufferLengths, void ** inBuffers,
-                                     nl::Weave::Platform::PersistedStorage::Key * inCounterKeys, const uint32_t * inCounterEpochs,
-                                     nl::Weave::PersistedCounter ** inCounterStorage)
+
+LoggingManagement::LoggingManagement(nl::Weave::WeaveExchangeManager * inMgr,
```
emargolis: This implementation assumes that number of input buffers is always kImportanceType_Last, which is not always correct. Some platform configurations can disable Info and Debug buffers.
size_t inNumBuffers   should be added as an input to this function.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -475,59 +447,57 @@ WEAVE_ERROR LoggingManagement::SetExchangeManager(nl::Weave::WeaveExchangeManage
...
+                                inLogStorageResources[i].mBufferSize - sizeof(CircularEventBuffer), prev, next);
 
-        current = prev                          = static_cast<CircularEventBuffer *>(inBuffers[i]);
+        current = prev                          = static_cast<CircularEventBuffer *>(inLogStorageResources[i].mBuffer);
         current->mBuffer.mProcessEvictedElement = AlwaysFail;
         current->mBuffer.mAppData               = NULL;
-        current->mImportance                    = static_cast<ImportanceType>(inNumBuffers - i);
-
-        if ((inCounterStorage != NULL) && (inCounterStorage[i] != NULL))
+        current->mImportance                    = static_cast<ImportanceType>(kImportanceType_Last - j);
```
emargolis: Why don't we add importance level to the LogStorageResources structure so we don't have to guess/assume specific order of the input buffers. It will also simplify things when some of the logging buffers are disabled.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: Thanks, fixed.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: done

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: done

src/lib/profiles/data-management/Current/LoggingManagement.h
robszewczyk: We do not have tooling that would check that at compile time (not sure even if it is possible). For runtime checks -- we will fail gracefully if the event is too large to fit into the first buffer, we don't handle the case where the event is too large to fit into subsequent buffers reserved for higher importance levels.  I see a way to add that check relatively easily, will add that check.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: Excellent idea, Done.  One small change I went with is use `VerifyOrDie` as the function returns no error, and there is no reasonable recovery from such an error

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: Agreed, I'm adding that to the PR.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: Agreed.   Thank you for the suggestion, adding to PR.

src/lib/profiles/data-management/Current/LoggingManagement.h
robszewczyk: Nevermind, I see I already added that check some time ago

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: done

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: done

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -400,52 +400,29 @@ void LoggingManagement::SkipEvent(EventLoadOutContext * aContext)
...
-        LoggingManagement(inMgr, inNumBuffers, inBufferLengths, inBuffers, inCounterKeys, inCounterEpochs, inCounterStorage);
-}
-
-/**
- * @brief Create and initialize the logging management buffers. Must
- *   be called prior to the logging being used.
- *
- * @param[in] inMgr         WeaveExchangeManager to be used with this logging subsystem.
+ * Initialize the LoggingManagement with an array of LogStorageResources.  The
+ * array must provide a resource for each valid importance level, the elements
```
emargolis: The function description should be updated because elements order is not important in current implementation.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -475,59 +452,64 @@ WEAVE_ERROR LoggingManagement::SetExchangeManager(nl::Weave::WeaveExchangeManage
...
+    VerifyOrDie(inNumBuffers > 0);
 
     mThrottled   = 0;
     mExchangeMgr = inMgr;
 
-    for (i = 0; i < inNumBuffers; i++)
+    for (j = 0; j < inNumBuffers; j++)
     {
-        next = ((i + 1) < inNumBuffers) ? static_cast<CircularEventBuffer *>(inBuffers[i + 1]) : NULL;
+        i = inNumBuffers - 1 - j;
```
emargolis: If the order of elements is not important then "i' can be used directly and "j' is not needed.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -538,68 +520,15 @@ LoggingManagement::LoggingManagement(WeaveExchangeManager * inMgr, size_t inNumB
 
         current->mFirstEventID = current->mEventIdCounter->GetValue();
     }
-    mEventBuffer = static_cast<CircularEventBuffer *>(inBuffers[0]);
+    mEventBuffer = static_cast<CircularEventBuffer *>(inLogStorageResources[kImportanceType_Last - kImportanceType_First].mBuffer);
```
emargolis: mEventBuffer = static_cast<CircularEventBuffer *>(inLogStorageResources[0].mBuffer);

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: Order remains important, and the function expects the elements to be ordered.  By providing the importance, we admit to gaps and non-contiguous importance enums.  Documentation is correct.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: Order remains important, and the function expects the elements to be ordered.  By providing the importance, we admit to gaps and non-contiguous importance enums. documentation remains correct.

PR #207, opened 2019-04-30
fix internal fault injection test fail CIEC-3526

robszewczyk:
Previous pull request added a new fault label for testing a particular scenario around WDM Update in the presence of faults.  This label needs to be added to the "not_required_server_faults" set in the scripts to ensure proper function. The fault injection tests are currently not a part of GitHub pull request validation, as at the moment they are too time consuming.


jenniexie:
Thanks @robszewczyk for helping out this PR quickly, and added me in the openweave members so I can use openweave branches directly future for my PRs.
Yay, I am openweave member now and finally :)

PR #209, opened 2019-05-01
Update nlbuild-autotools to 1.6.9

gerickson:
This pulls in an update to nlbuild-autotools-1.6.9 which, most importantly, addresses an issue where modern versions of m4 crash on Mac OS X High Sierra and Mojave by down-revving m4 to an older 1.4.5.


gerickson:
> I would prefer we wait with the merge until @jaylogue is able to land his `feature/nrf52840-device-layer` branch. If other things are simplified by merging this right away, I would strongly request we do not regenerate the automake tools

This should have zero impact to that branch and should have zero conflicts. In the interim, the m4 issue will fix a real problem that people may be running into. Your call at the end of the day.

robszewczyk:
> > I would prefer we wait with the merge until @jaylogue is able to land his `feature/nrf52840-device-layer` branch. If other things are simplified by merging this right away, I would strongly request we do not regenerate the automake tools
> 
> This should have zero impact to that branch and should have zero conflicts. In the interim, the m4 issue will fix a real problem that people may be running into. Your call at the end of the day.

OK, lets merge this now, and we will regenerate the `*.in` files later on

PR #212, opened 2019-05-03
use maes for x86_64 arch only

yunhanw-google:
maes don't work in arm.
update uname -m for ARCH check since uname -i show unknown in our linux
setup and -i don't work in macbook.

### DIFF
--- a/Makefile-Standalone
+++ b/Makefile-Standalone
@@ -40,7 +40,7 @@ BLUEZ                          ?= 0
 USE_FUZZING                    ?= 0
 
 HOSTOS                          = $(shell uname -s |tr [:upper:] [:lower:])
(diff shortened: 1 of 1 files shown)


robszewczyk:
Please do the following: 
* close this pull request
* resubmit the commit to a branch of openweave-core and open a new pull request
* Reword the commit message to read 
 
    Properly conditionalize the use of -maes option

    The compilation flag -maes is applicable only to x86_64 processors that support AES-NI instructions.  Use the flag only in that context.  
    
    Switch to using `uname -m` to determine the machine architecture.  This call is POSIX-postable, whereas `uname -i` and `uname -p` are non-portable and produce unexpected results on Linux ARM, Mac OS x86_64, etc.

PR #223, opened 2019-05-10
UpdateRequestBadMsgType fault injection added

pkowalewski2:
Fault injection added that allows forcing cloud to respond with faulty status code.

### DIFF
--- a/src/lib/profiles/data-management/Current/UpdateClient.cpp
+++ b/src/lib/profiles/data-management/Current/UpdateClient.cpp
@@ -158,6 +158,11 @@ WEAVE_ERROR UpdateClient::SendUpdate(bool aIsPartialUpdate, PacketBuffer *aBuf, 
                 nl::Weave::FaultInjection::kFault_WRMSendError,
                 0, 1));
 
+    // Fault injection used to force faulty status code response from Phoenix
+    WEAVE_FAULT_INJECT(FaultInjection::kFault_WDM_UpdateRequestBadMsgType,
+            msgType = 0;
+            );
+
     err = mEC->SendMessage(nl::Weave::Profiles::kWeaveProfile_WDM, msgType, aBuf,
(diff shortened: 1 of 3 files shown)


pkowalewski2:
I signed it!

src/lib/profiles/data-management/Current/UpdateClient.cpp
```diff
@@ -158,6 +158,11 @@ WEAVE_ERROR UpdateClient::SendUpdate(bool aIsPartialUpdate, PacketBuffer *aBuf,
                 nl::Weave::FaultInjection::kFault_WRMSendError,
                 0, 1));
 
+    // Fault injection used to force faulty status code response from Phoenix
+    WEAVE_FAULT_INJECT(FaultInjection::kFault_WDM_UpdateRequestBadMsgType,
+            msgType = 0;
```
jaylogue: WDM message type 0 is not an invalid message type.  It is a deprecated form of ViewRequest used by an old version of the WDM protocol.

I'm not certain what the purpose of this code is, but generally speaking, fault injection should not be used to inject faults into other systems, with the expectation that they will respond in a specific way, because they may not always do that.  For example, in the future, the service team may decide that, if a client can't even get the message type correct, they then shouldn't respond at all.  Thus this fault could result in a timeout, rather than a specific Status Report.

A better use of fault injection is to alter the internal state or inputs to the local system and see how it reacts.  In this case, when a response is received from the service, the fault injection test would transform it into the desired Status Report and confirm the system reacted appropriately.

pkowalewski2: Thanks for this comment. The purpose is to reproduce test scenario when service responds to UpdateRequest with fault and then sends notification with wrong setting value. I abandon this PR and think of some other solution.

PR #226, opened 2019-05-14
Make header importable by Objective-C source files

Alarantalara:
constexpr is not available in Objective-C, only Objective-C++
Using the older form of constness allows this file to be directly imported by Objective-C sources


robszewczyk:
verified it builds for iOS.  merging

PR #228, opened 2019-05-16
Added a few of the missing status codes to improve error reporting

robszewczyk:


src/lib/profiles/data-management/Current/MessageDef.h
```diff
@@ -33,6 +33,56 @@
 namespace nl {
 namespace Weave {
 namespace Profiles {
+namespace DataManagement_Legacy {
```
gerickson: Wrong namespace?

robszewczyk: No, that's the required namespace.  The idea is that we need the legacy namespace so that we can pull in the old status codes

gerickson: But aren't we re-importing or re-instantiating those legacy status code in new ("current") headers for the new ("current") implementation? If so, this Legacy namespace usage seems superfluous at best and incorrect and confusing at worst.

robszewczyk: The core problem I ran into is that I *cannot* include both the legacy headers and current headers in the same file.  If I were able to do so, I would happily, the current structure does not permit it.  Perhaps the answer is to fix the managed namespace headers, I will take a look at that.

Into the rabbit hole we go...

gerickson: It might be worth reconciling WDM against what was documented here: src/lib/support/ManagedNamespace.hpp. Happy to sit down and review where things might have gone off the rails relative to the intent of managed namespaces.

PR #235, opened 2019-05-20
simplify wdm tests

jenniexie:
1. removed sanity test because sanity is covered in stress test.
2. use count in wdm_next_args['client_log_check'] for easy
understanding.
3. use function for common code block in weave_wdm_next_test_base.py

### DIFF
--- a/src/test-apps/happy/tests/standalone/wdmNext/test_weave_wdm_next_application_key_01.py
+++ b/src/test-apps/happy/tests/standalone/wdmNext/test_weave_wdm_next_application_key_01.py
@@ -66,8 +66,7 @@ class test_weave_wdm_next_application_key_01(weave_wdm_next_test_base):
                                              ('Client\[0\] moving to \[ FREE\] Ref\(0\)', wdm_next_args['test_client_iterations']),
                                              ('Handler\[0\] Moving to \[ FREE\] Ref\(0\)', wdm_next_args['test_client_iterations'])]
         wdm_next_args['test_tag'] = self.__class__.__name__[35:].upper()
-        wdm_next_args['test_case_name'] = ['A01: Mutual Subscribe: Application key: Key distribution',
-                                           'B01: Stress Mutual Subscribe: Application key: Key distribution']
+        wdm_next_args['test_case_name'] = ['B01: Stress Mutual Subscribe: Application key: Key distribution']
         print 'test file: ' + self.__class__.__name__
         print "test_weave_wdm_next_application_key test A01 and B01"
         super(test_weave_wdm_next_application_key_01, self).weave_wdm_next_test_base(wdm_next_args)
--- a/src/test-apps/happy/tests/standalone/wdmNext/weave_wdm_next_test_base.py
+++ b/src/test-apps/happy/tests/standalone/wdmNext/weave_wdm_next_test_base.py
@@ -135,20 +135,61 @@ class weave_wdm_next_test_base(unittest.TestCase):
             self.__process_result("node01", "node02", value, data)
             self.result_data = data
 
+    def __validate_log_line(self, data, node_type, wdm_stress_check, stress_failures):
+        """
+        check expected log lines showed up in node logs with expected count
+        """
+        if node_type == "client":
+            node_log = data["client_output"]
+            log_check_list = self.client_log_check
+        else:
+            node_log = data["server_output"]
+            log_check_list = self.server_log_check
+        for expect_line_count in log_check_list:
+            item_output_cnt = len(re.findall(expect_line_count[0], node_log))
+            if item_output_cnt != expect_line_count[1]:
+                wdm_stress_check = False
+                error = "missing '%s' in %s_output, need %d, actual %d" % (expect_line_count[0], node_type, expect_line_count[1], item_output_cnt)
+                stress_failures.append(error)
+        return (wdm_stress_check, stress_failures)
+
+    def __validate_node_events(self, wdm_stress_check, stress_failures, data, client_gen):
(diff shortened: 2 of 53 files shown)


src/test-apps/happy/tests/standalone/wdmNext/test_weave_wdm_next_application_key_01.py
```diff
@@ -42,7 +42,7 @@ def test_weave_wdm_next_application_key_01(self):
         wdm_next_args['total_client_count'] = 2
         wdm_next_args['final_client_status'] = 0
         wdm_next_args['timer_client_period'] = 5000
-        wdm_next_args['test_client_iterations'] = 5
+        count = self.test_client_iterations
```
gerickson: Where does this get set the first time. In some instances, I see this explicitly set to '1'; however, in other cases, like this one, it appears to be set out of scope of this review.

jenniexie: Thanks Grant for the feedback!
You are right, there are 2 ways of setting the test count:
1. explicitly set to 1
2. setting it from the wdm_next_test_base class in below code, from setUp() function, every test script will call the setUp() function first. 
class weave_wdm_next_test_base(unittest.TestCase):
    def setUp(self):
          self.test_client_iterations = self.__read_wdm_test_config(
               os.path.dirname(os.path.abspath(__file__))+"/wdm_test_config.json", 
               "test_client_iterations")

Ideally, I would like to use 1 for all tests in Travis CI test to run Sanity test only, but some log check required to have 5 test iterations. I did not change any test logic for now.

I will have meeting with @yunhanw-google tomorrow to discuss if we can change the log check requirement, and use same test count for all tests.

jenniexie:
Based on today's meeting with Yunhan and Qilin, the feedback is to keep wdm_next_args['test_client_iterations'] in the test script file for this PR.
In future PRs, I will address the parameters' concern and make it more simplified.

src/test-apps/happy/tests/standalone/wdmNext/weave_wdm_next_test_base.py
```diff
@@ -135,106 +135,68 @@ def weave_wdm_next_test_base(self, wdm_next_args):
...
+            log_check_list = self.server_log_check
+        for expect_line_count in log_check_list:
+            item_output_cnt = len(re.findall(expect_line_count[0], node_log))
+            if item_output_cnt != expect_line_count[1]:
+                wdm_stress_check = False
+                error = "missing '%s' in %s_output, need %d, actual %d" % (expect_line_count[0], node_type, expect_line_count[1], item_output_cnt)
+                stress_failures.append(error)
+        return (wdm_stress_check, stress_failures)
+
+    def __validate_node_events(self, wdm_stress_check, stress_failures, data, client_gen=False, server_gen=False):
```
robszewczyk: This seems like a failure-prone API: exactly one of `client_gen` and `server_gen` MUST be true or the method will fail.  It would be better if the API took a single parameter, e.g. `client_gen`, and lines 162..173 were a `if...else` statement.  The calls to the __validate_node_events below will also need adjusting.

robszewczyk: def __validate_node_events(self, wdm_stress_check, stress_failures, data, use_client_gen):
    ...
    if use_client_gen:
        client_output = data["client_output"]
        ....
    else 
        client_out = data["server_output"]

jenniexie: Thanks RS for the feedback!
It is much clear after the discussion, I will make changes now.

jenniexie:
@robszewczyk I have addressed both comments, and verified the wdm tests pass at my local PC.
Please take a quick look and let me know, Thanks!

src/test-apps/happy/tests/standalone/wdmNext/weave_wdm_next_test_base.py
```diff
@@ -135,106 +135,68 @@ def weave_wdm_next_test_base(self, wdm_next_args):
...
+        if client_gen:
+            client_output = data["client_output"]
+            publisher_output = data["server_output"]
+            node_role = {"client":"client",
+                         "publisher":"server"
+                        }
+        if server_gen:
+            client_output = data["server_output"]
+            publisher_output = data["client_output"]
+            node_role = {"client":"sever",
```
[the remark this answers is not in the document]
jenniexie: done

src/test-apps/happy/tests/standalone/wdmNext/weave_wdm_next_test_base.py
jenniexie: done

src/test-apps/happy/tests/standalone/wdmNext/weave_wdm_next_test_base.py
```diff
@@ -135,106 +135,68 @@ def weave_wdm_next_test_base(self, wdm_next_args):
             self.__process_result("node01", "node02", value, data)
             self.result_data = data
 
+    def __validate_log_line(self, data, node_type, wdm_stress_check, stress_failures):
+        """
+        check expected log lines showed up in node logs with expected count
+        """
+        if node_type is "client":
```
robszewczyk: Please replace `is` with `==`

jenniexie: good point, done with the change

PR #242, opened 2019-05-25
Clean up build support for OS X

jaylogue:
-- Revised Makefile-Standalone to automatically detect and use the standard OpenSSL package installed on either Linux or OS X, when available.  On Linux, the logic uses pkg-config to determine if the standard OpenSSL development package (typically libssl-dev) has been installed.  On OS X, the logic uses homebrew to look for the OpenSSL formula.  If a standard package is not detected, the build falls-back to using the internal version of OpenSSL located in third_party/openssl.

-- Added OPENSSL argument to Makefile-Standalone that allows the developer to override the above logic.

-- Updated Travis build script for OS X to use Makefile-Standalone, rather that calling configure and make directly.

-- Added prebuild steps to Travis OS X builds to install homebrew packages for openssl, automake and libtool.

-- Generally cleaned up the logic in Makefile-Standalone for handling OpenSSL options.

-- Cleaned up the Makefile-Standalone help text.

-- Fixed a small sed expression incompatibility in nrf5-app.mk that preventing building nRF5 SDK applications on OS X.

### DIFF
--- a/Makefile-Standalone
+++ b/Makefile-Standalone
@@ -64,7 +64,11 @@ AbsTopResultDir                 = $(CURDIR)/$(TopResultDir)
 
 TargetTuple                     = $(shell ${AbsTopSourceDir}/third_party/nlbuild-autotools/repo/third_party/autoconf/config.guess | sed -e 's/[[:digit:].]*$$//g')
 
+ifeq ($(HOSTOS),darwin)
+ProjectConfigDir               ?= $(AbsTopSourceDir)/build/config/standalone/darwin
+else
 ProjectConfigDir               ?= $(AbsTopSourceDir)/build/config/standalone
+endif
 
 ifeq ($(ARCH),x86_64)
 configure_OPTIONS               = CPPFLAGS="-maes"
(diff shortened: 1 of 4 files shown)


Makefile-Standalone
```diff
@@ -437,46 +455,59 @@ clean:
...
-	$(ECHO) "  HAPPY                   When set to 0, skip the functional tests that require"
-	$(ECHO) "                          Happy orchestration tool.  When set to 1, pick the"
-	$(ECHO) "                          default location to the Happy tool.  When set to any"
-	$(ECHO) "                          other value, the value is treated as the location of"
-	$(ECHO) "                          the Happy tool."
-	$(ECHO) ""
+	@echo "$${HelpText}"
+
+export HelpText
+define HelpText
```
robszewczyk: This is sweet.   Thanks!

Makefile-Standalone
```diff
@@ -87,22 +91,36 @@ ifeq ($(HOSTOS),darwin)
 configure_OPTIONS              += ac_cv_func_clock_gettime=no  ac_cv_have_decl_clock_gettime=no
 endif
 
-# By default, build using the system's OpenSSL library, except on OSX which needs
-# the Weave internal OpenSSL.
-# However, if NO_OPENSSL = 1, build with an alternate configuration that avoids the
-# use of OpenSSL.  Note that building with NO_OPENSSL = 1 automatically suppresses
-# the building of various command line tools that depend on OpenSSL (e.g. the weave tool).
+# By default, attempt use the system's OpenSSL library, falling back to building the internal
```
gerickson: Any reason to have not included this in configure?

PR #249, opened 2019-05-31
WEAV-3246: Dockerfile for weave device manager

NLAVogel:
Also add __init__.py so when files are installed in dist-packages they
can be imported.

### DIFF
--- a/build/contrib/docker/Dockerfile
+++ b/build/contrib/docker/Dockerfile
@@ -1,0 +1,20 @@
+FROM ubuntu:14.04
+
+RUN \
+    apt-get update && \
+    apt-get -y upgrade && \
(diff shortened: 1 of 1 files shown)


build/contrib/docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    pip install --user grpc google-cloud protobuf && \
+    sudo cpan -i Text::Template && \
+    mkdir weave
+
+COPY . weave
+
+RUN cd weave && \
+    ./configure --disable-systemd && \
+    make all && \
+    make install
```
jaylogue: I think we should be using Makefile-Standalone whenever we build any of the desktop tools, rather than invoking ./configure with default args.  The former employs project configuration files that configures OpenWeave specifically for desktop use.

PR #255, opened 2019-05-31
Group test instances run within happy by PID of the makefile

robszewczyk:
The changes in this commit add a PID-based prefix to all happy tests
run within a particular instance of the build.  This enables a
non-conflicting way to run multiple simultaneous execution of Weave
tests on the same machine, as the "happy" simulator instances are
non-overlapping.

### DIFF
--- a/src/test-apps/Makefile.am
+++ b/src/test-apps/Makefile.am
@@ -23,6 +23,8 @@
 #
 
 include $(abs_top_nlbuild_autotools_dir)/automake/pre.am
+SHELL := /bin/bash
+AM_LOG_DRIVER_FLAGS := --build-num $(shell echo $$PPID|tail -c5)
 
 #
 # Other files we do want to distribute with the Weave source SDK.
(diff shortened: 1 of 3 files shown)


src/test-apps/Makefile.am
```diff
@@ -24,6 +24,8 @@
 
 include $(abs_top_nlbuild_autotools_dir)/automake/pre.am
 
+AM_LOG_DRIVER_FLAGS := --build-num $(shell echo $$PPID)
```
gerickson: I suspect this may blow up on shells in which SHELL is /bin/sh but in which /bin/sh is a symlink to /bin/dash or /bin/zsh rather than /bin/bash since I believe PPID is a bashism.

Specifying SHELL := `which bash` may work around this.

robszewczyk: tested with `echo $PPID` on dash, zsh, ash (which should be available on busybox systems), ksh, and multiple versions of bash. Worked in each case. Fails as might be expected in csh and tcsh, but on csh-based shells other things might fail as well (e.g. redirects that we use extensively).

robszewczyk: Changed the shell to `/bin/bash` -- there are a number of other components of the build system that depend on bash being installed in `/bin/bash` rather than in an arbitrary location.  I have also added a restriction on the length of the PID -- I now only take the last 5 characters plus a newline; this ensures that the overall network namespace ID continues to fit in 15 characters.

PR #262, opened 2019-06-05
Fix a test edge case in working with non-persistent counters

robszewczyk:
When using the non-persistent counters, the counters, by default,
start at 0; that behavior is exploited in a number of unit tests to
verify the event ID progression.  However, this leads to a delay in
event delivery in a case where only a single event for a particular
importance level has been logged; a number of tests that use fault
injection depend on a single event being delivered immediately.  Prior
to the refactoring of the initialization of the logging subsystem this
issue was addressed by passing an array of non-persistent counters
initialized to the current system time.  This patch restores that
functionality though at a different level: we add a method to the
LoggingManagement class to re-initialize the non-persistent counters
to the current system time.  This method is intended for use only in
test code.

### DIFF
--- a/src/lib/profiles/data-management/Current/LoggingManagement.cpp
+++ b/src/lib/profiles/data-management/Current/LoggingManagement.cpp
@@ -531,6 +531,34 @@ LoggingManagement::LoggingManagement(nl::Weave::WeaveExchangeManager * inMgr,
 
 /**
  * @brief
+ *  Reinitialize the non-persistent counters to values other than 0
+ *
+ * This function, much like the non-persistent counters themselves, should only
+ * be used in test code.
+ *
+ * The function will reinitialize the event ID counters to the current system
+ * time for all importance levels that use the non-persistent counters.  The
+ * counters will get reinitialized only if there were no events logged at that
+ * importance level.
+ */
+void LoggingManagement::ReinitializeMonotonicEventCounters(void)
+{
+    CircularEventBuffer *current = mEventBuffer;
+    while (current != NULL)
+    {
+        if ((current->mEventIdCounter == &(current->mNonPersistedCounter)) &&
+            (current->mFirstEventID == 0) &&
+            (current->mLastEventID == 0))
+        {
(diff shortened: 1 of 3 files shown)


src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -529,6 +529,34 @@ LoggingManagement::LoggingManagement(nl::Weave::WeaveExchangeManager * inMgr,
...
+ *
+ * This function, much like the non-persistent counters themselves, should only
+ * be used in test code.
+ *
+ * The function will reinitialize the event ID counters to the current system
+ * time for all importance levels that use the non-persistent counters.  The
+ * counters will get reinitialized only if there were no events logged at that
+ * importance level.
+ */
+void LoggingManagement::RandomizeMonotonicEventCounters(void)
```
gerickson: Given the implementation, it might be better to rename this to "ReinitializeMonotonic..." or some such.

src/lib/profiles/data-management/Current/LoggingManagement.h
```diff
@@ -245,6 +245,7 @@ class LoggingManagement
 #if WEAVE_CONFIG_EVENT_LOGGING_WDM_OFFLOAD
     bool CheckShouldRunWDM(void);
 #endif
+    void RandomizeMonotonicEventCounters(void);
```
gerickson: See prior comment.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
robszewczyk: done

src/lib/profiles/data-management/Current/LoggingManagement.h
robszewczyk: done

robszewczyk:
After a conversation with @jaylogue we decided on a different course of action.  Rather than introducing an odd, test-only API, I am modifying the TestEventLogging test cases to assume that the non-persistent counters are initialized to `1`.  This results in removing a bunch of related one-off code, and in general makes a bit more sense.

PR #268, opened 2019-06-10
Weave Device Manager clean-up

jaylogue:
-- Build and package the Weave Device Manager C++ object as part of the main Weave static library (libWeave.a), rather than as a separate shared library.  This eliminates load-time dependencies on a weave-specific shared library (libWeaveDeviceManager.so) for components which incorporate the Weave Device Manager.

-- Changed the meaning of the --enable-device-manager/--disable-device-manager configure options.  These options now control whether or not the various Device Manager wrapper modules (i.e. the Java, Python and Cocoa wrappers) are built.  Previously these also controlled whether the core Device Manager C++ code was built.  However this code is now built and included in the main Weave static library in all cases.

-- Restructured the build of the Java Device Manager wrapper such that it creates a dedicated shared library specifically for use by the Java code.  This library is built as a shared “module” (i.e. a shared library which is only capable of being loaded dynamically at runtime).  To avoid any changes to the Java code, the library retains its original name of libWeaveDeviceManager.so.

-- Restructured the build of the Python Device Manager wrapper such the associated shared library (_WeaveDeviceMgr.so) includes all of Weave, rather than having a load-time dependency on libWeaveDeviceManager.so.  Like the Java wrapper, the Python wrapper shared library is built as a shared module.

-- Restructured the Device Manager Python code into a Python package named “openweave”.  The APIs to the existing Python modules (WeaveDeviceMgr, WeaveBluezMgr, etc.) remain the same, however the modules must now be imported using a package relative syntax; e.g:

```
from openweave import WeaveDeviceMgr
```

As part of this change, the installed location of the Python wrapper shared library (_WeaveDeviceMgr.so) was moved into the root directory of the “openweave” package.

-- Rewrote the logic that the WeaveDeviceMgr Python class uses to locate the Python wrapper shared library.  By default the library is expected to be found in the root directory of the package in which executing WeaveDeviceMgr python class was found.  If not found there, the code will search up the directory hierarchy for an OpenWeave build directory containing the library in the expected build location.  This behavior makes it possible to import and use the WeaveDeviceMgr module directly from a built copy of the OpenWeave source tree.

-- Modified the Weave Device Manager Shell (weave-device-mgr.py) to search for the “openweave” python package in directories relative to the location of the script itself. This makes it possible to run the device manager shell from a non-standard install location, as well as directly from its location the OpenWeave source tree.  Relative locations are given priority over standard install locations so as to ensure the correct version of the package is used.

-- Eliminated the use of the weave-run helper script for executing the Weave Device Manager Shell.  The helper script was no longer necessary after the elimination of the generic Weave Device Manager shared library.

-- Added logic to configure.ac to disable building of the Device Manager wrapper modules when the target network type is LwIP only.  Although building this configuration is technically feasible, the resultant code is not usable.  Note that this does not disable building the core Device Manager C++ class.


gerickson:
Do embedded platforms appropriately dead-strip out device manager code from libWeave.a?

jaylogue:
> Do embedded platforms appropriately dead-strip out device manager code from libWeave.a?

The linker won't bring in any of the device manager code unless its expressly referenced by the device code.

yunhanw-google:
Hi, Jay

Previously weave-device-mgr in build folder is working, with your change, it looks broken.

/workspace/openweave-core/build/x86_64-unknown-linux-gnu/src/device-manager/python$ ls
Makefile
weave-device-mgr
_WeaveDeviceMgr.la
_WeaveDeviceMgr_la-WeaveDeviceManager-BleApplicationDelegate.lo
_WeaveDeviceMgr_la-WeaveDeviceManager-BleApplicationDelegate.o
_WeaveDeviceMgr_la-WeaveDeviceManager-BlePlatformDelegate.lo
_WeaveDeviceMgr_la-WeaveDeviceManager-BlePlatformDelegate.o
_WeaveDeviceMgr_la-WeaveDeviceManager-ScriptBinding.lo
_WeaveDeviceMgr_la-WeaveDeviceManager-ScriptBinding.o

/workspace/openweave-core/build/x86_64-unknown-linux-gnu/src/device-manager/python$ ./weave-device-mgr 
Traceback (most recent call last):
  File "./weave-device-mgr", line 62, in <module>
    from openweave import WeaveDeviceMgr
ImportError: No module named openweave

Thanks
Best wishes
Yunhan

jaylogue:
> Previously weave-device-mgr in build folder is working, with your change, it looks broken.

Please try running weave-device-mgr.py directly from the source directory; e.g. /workspace/openweave-core/src/device-manager/python/weave-device-mgr.py

yunhanw-google:
/workspace/openweave-core/src/device-manager/python/weave-device-mgr.py, this looks good, are you going to add support to make weave-device-mgr.py in build folder work?

jaylogue:
>are you going to add support to make weave-device-mgr.py in build folder work?

I could but I would prefer not to.  Would it be possible to change your code to run it from the source directory?

yunhanw-google:
When directly running the test, python src/test-apps/happy/tests/standalone/pairing/test_weave_pairing_01.py or test_weave_ble_01.py

I need import openweave.WeaveDeviceMgr

When running coverage test using "sudo make -f Makefile-Standalone COVERAGE=1 TIMESTAMP=1 SUITE=pairing BLUEZ=1 coverage"
I need import WeaveDeviceMgr 

Therefore, I made the below tentative change to try openweave.WeaveDeviceMgr firstly, if failing, try import WeaveDeviceMgr directly, the same applies to WeaveBluezMgr and WeaveBleUtility. It seems openweave scope does not take effective when directly running test.


diff --git a/src/test-apps/happy/lib/WeaveDeviceManager.py b/src/test-apps/happy/lib/WeaveDeviceManager.py
index 1762961c..63222fdb 100644
--- a/src/test-apps/happy/lib/WeaveDeviceManager.py
+++ b/src/test-apps/happy/lib/WeaveDeviceManager.py
@@ -43,9 +43,12 @@ if os.path.exists(weaveDeviceMgrLibPath):
     sys.path.append(weaveDeviceMgrLibPath)
 
 try:
-    import WeaveDeviceMgr
+    from openweave import WeaveDeviceMgr
 except Exception as ex:
-    print ("Could not import the WeaveDeviceMgr module: %s" % (str(ex)))
+    try:
+        import WeaveDeviceMgr
+    except Exception as ex:
+        print ("Could not import the WeaveDeviceMgr module: %s" % (str(ex)))
 
 # Dummy Access Token
 #
@@ -162,8 +165,15 @@ if __name__ == '__main__':
 
     try:
         if options.useBle:
-            from WeaveBluezMgr import BluezManager as BleManager
-            from WeaveBleUtility import FAKE_CONN_OBJ_VALUE
+            try:
+                from openweave.WeaveBluezMgr import BluezManager as BleManager
+                from openweave.WeaveBleUtility import FAKE_CONN_OBJ_VALUE
+            except:
+                try:
+                    from WeaveBluezMgr import BluezManager as BleManager
+                    from WeaveBleUtility import FAKE_CONN_OBJ_VALUE
+                except:
+                    print "cannot find WeaveBluezMgr and WeaveBleUtility"
             bleManager = BleManager(devMgr)
 
             try:

yunhanw-google:
issue resolved.
thanks

PR #269, opened 2019-06-10
Fix install-headers target

robszewczyk:
Some builds may desire to run the install-headers target prior to
building the entire tree.  Add dependencies and rules to generate the
ASN1OID.h header whenever appropriate.

### DIFF
--- a/src/include/Makefile.am
+++ b/src/include/Makefile.am
@@ -469,6 +469,12 @@ $(filter %.hpp,$(nl_public_WeaveSupport_header_destinations)): $(nl_public_Weave
 # from the build directory which may not be the same as the source
 # directory in a non-colocated build.
 
+$(abs_builddir)/../$(nl_public_WeaveSupport_source_dirstem)/ASN1OID.h:
+	(cd $(abs_builddir)/../$(nl_public_WeaveSupport_source_dirstem) && $(MAKE) $(AM_MAKEFLAGS) ASN1OID.h)
+
+$(addprefix ../,$(nl_public_WeaveSupport_ASN1OID_base_header_sources)):
+	(cd $(abs_builddir)/../$(nl_public_WeaveSupport_source_dirstem) && $(MAKE) $(AM_MAKEFLAGS) ASN1OID.h)
+
 $(nl_public_WeaveSupport_destination_dirstem)/ASN1OID.h: $(abs_builddir)/../$(nl_public_WeaveSupport_source_dirstem)/ASN1OID.h | $(nl_public_WeaveSupport_ASN1OID_header_dirs)
 	$(call create-link)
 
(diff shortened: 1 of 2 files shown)


src/include/Makefile.am
```diff
@@ -469,6 +469,12 @@ $(filter %.hpp,$(nl_public_WeaveSupport_header_destinations)): $(nl_public_Weave
 # from the build directory which may not be the same as the source
 # directory in a non-colocated build.
 
+$(abs_builddir)/../$(nl_public_WeaveSupport_source_dirstem)/ASN1OID.h:
```
gerickson: make -C ?

src/include/Makefile.am
```diff
@@ -469,6 +469,12 @@ $(filter %.hpp,$(nl_public_WeaveSupport_header_destinations)): $(nl_public_Weave
...
+	(cd $(abs_builddir)/../$(nl_public_WeaveSupport_source_dirstem) && $(MAKE) $(AM_MAKEFLAGS) ASN1OID.h)
+
+$(addprefix ../,$(nl_public_WeaveSupport_ASN1OID_base_header_sources)):
```
gerickson: make -C ?

gerickson:
I need to give it some thought myself; however, have you had any thoughts on a good unit / functional test for this?

jaylogue:
I'm concerned that this pattern of use is not something we can support long-term.  If we end up adding other code generation tools to our build it seems likely that the install-headers target is not going to work.

Why is install-headers being used?

robszewczyk:
> I need to give it some thought myself; however, have you had any thoughts on a good unit / functional test for this?

I'm not sure I've ever seen unit tests for a specific instantiation of a build system/makefile.

robszewczyk:
> I'm concerned that this pattern of use is not something we can support long-term. If we end up adding other code generation tools to our build it seems likely that the install-headers target is not going to work.
> 
> Why is install-headers being used?

I think in general, the makefiles in src/include have been problematic, they certainly have not been particularly easy to work with. 

On systems that heavily parallelize builds, including a number of Google Nest products, there is a usecase for running install-headers prior to the full build -- the rationale is that once the build system installs the headers, all other modules that depend on Weave being built can compile.  The link step then has a separate dependency on the Weave library being built.

You are right that the solution presented here is not scalable.  Perhaps a better approach would be to create a custom dependency target that just builds all `BUILT_SOURCES` prior to any work on installing headers. The primary challenge with is that autotools do not define a single action across the board that would support `BUILT_SOURCES`, but that is something we could likely support with nlbuild-autotools.  I will prototype this solution. 

A different mechanism would be to generate the sources during the configure step; such approach meshes well with automated build systems, but has some pitfalls with partial builds.

Finally, we could entirely subvert the intentions of the build system, and make install-headers depend on the overall compilation step.  Not a great solution, but perhaps a defensible one:  in practice, the `configure` step takes comparable time to the actual baseline compilation step (no docs, no tests).  As a result, the rough order of magnitude for build times would remain the same.

PR #275, opened 2019-06-19
Implement new Changelog format

### LINKED ISSUE #273: Implement new Changelog format
We have an existing changelog, but it's had only minimal updates:
https://github.com/openweave/openweave-core/blob/master/CHANGELOG

Since we don't have a formal release process with versioning and notes, we should reformat the changelog to be a bit more comprehensive, and use it for all substantial feature and dependency changes that should be highlighted for end users.

Taking inspiration from:
https://keepachangelog.com/en/1.0.0/
https://github.com/olivierlacan/keep-a-changelog/blob/master/CHANGELOG.md

I can get it started with some of the recent updates made to the project.

Vyrastas:
Reformats the existing Changelog into a Markdown format to be more informative, serve as a place to summarize all substantial changes to OpenWeave Core we want users to be aware of, and provide quick access to the diffs between versions.  The versioning here is based on the release tags that exist in the repo currently.

I've added a few items to the most recent versions to give an idea of what I think we should be calling out.

I'm assuming we're adhering to semantic versioning, but I'm not sure if there is any significance to the versioning right now beyond internal Google usage.

Inspired by https://github.com/olivierlacan/keep-a-changelog/blob/master/CHANGELOG.md

Fixes #273


gerickson:
> I don't think this is what we want.
> 
> Recently we agreed that openweave-core will not be adhering to semantic versioning for the foreseeable future.

Can you provide more context? Who does "we" comprise?

> Rather, master will be considered always stable and suitable for general use. As necessary, downstream consumers will pin to particular commits (typically using git submodule) to avoid accepting any destabilizing changes.

I don't think semantic versioning or what @Vyrastas is proposing here is incompatible with that stability mandate. There is certainly no harm in creating semantic versions in such a world and, I'd argue, creating semantic versions at appropriate points is mnemonically more useful to consumers of openweave-core.

> We also agreed to document (in some sort of CHANGELOG or NEWS file) only major feature enhancements and/or large-impact changes to existing functionality. Otherwise, the commit history for the master branch will serve as the log of detailed changes for a downstream consumer.
> 
> I would prefer the entries in the CHANGELOG/NEWS file to be roughly a paragraph or two of descriptive text giving a summary of the feature/enhancement, why its interesting, and pointing readers to additional documentation if available. Ideally this should read as a "latest news" blog, rather than a point-by-point list of changes. I think a somewhat conversational style would be useful here.
> 
> The purpose of all this is to avoid the overhead of a formal release process, and the associated detailed documentation, while still providing a means to communicate our efforts to OpenWeave's users.

Who does "we" comprise?

Fully agreed here. What's in CHANGELOG or NEWS should be up-leveled from the minutiae and detail of commit history. However, again, I don't think that incompatible with semantic versioning and I do not believe that semantic versioning warrants or mandates any overhead, formality, or overly-detailed documentation. It's simply a way to create a user-visible impact statement in conversational form (as you advocate) and give it a well-known and somewhat-more memorable name rather than a cryptic git hash. Some commits will be worthy of such a name and entry; others will not. Developers and maintainers get to decide when the threshold is exceeded.

Vyrastas:
Revised the format based on offline discussions.  This update:

- Clarifies that openweave-core does not use semantic versioning and that master is stable
- Changes headers to date of update
- Adds an update around docs and this changelog as an example of the format going forward
- Preserves the original changelog entries
- Updates Makefiles etc since the filename changed

PTAL and let me know if this aligns with what you want.  I put the single doc update in there in lieu of any of your recent features to make it easier to illustrate the format.

Vyrastas:
@gerickson Will it work for you to have this format as each header?

2019-05-24 (4.1.2)

Without strict versioning for all updates or releases, leading with the date would be the most consistent format.

PR #278, opened 2019-06-21
Fixed hang in Device Manager on unexpected connection close

jaylogue:
-- Fixed a logic bug in the WeaveDeviceManager C++ class that resulted in the application’s on-error callback never being called when a connection to a device closed unexpectedly during a connect or rendezvous operation.  In the observed scenario, a spontaneous close of a newly established Weave connection just as a security session was starting would result in the Device Manager ignoring the HandleSessionError() callback from the WeaveSecurityManager, but only in the case where it had received a HandleConnectionClosed() callback prior to the HandleSessionError().

The ordering of the two events depends very much on the behavior of the underlying TCP stack.  Indeed, the situation may only be prevalent on mobile devices, where the network stack immediately invalidates TCP sockets when there is a change in WiFi/Cellular connectivity.

The fix was loosen a state check such that the session establishment error is process correctly regardless of the ordering of the two events.

This fix relates to the problem reported in buganizer issue 135032367.

### DIFF
--- a/src/device-manager/WeaveDeviceManager.cpp
+++ b/src/device-manager/WeaveDeviceManager.cpp
@@ -3394,15 +3394,7 @@ void WeaveDeviceManager::HandleSessionError(WeaveSecurityManager *sm, WeaveConne
     DeviceStatus devStatus;
     DeviceStatus *devStatusArg = NULL;
 
-    // Bail immediately if not in the correct state. May occur if the connection closes abruptly and the
-    // SecurityManager's HandleConnectionClosed callback fires _after_ the DeviceManager's own callback.
-    // In this case, con is already closed and mOnError has already been called, so we should just exit.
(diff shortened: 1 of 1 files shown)


jaylogue:
@mburshteyn1 Actually, no.  The root problem in OBS-5981 was that the code in HandleSessionError() 
 called con->Closed() in the case where mConState == kConnectionState_NotConnected.  Since con->Close() had already been called in HandleConnectionClosed(), this caused the connection reference count to go below zero.

Note that, even though HandleSessionError() can now go on to call devMgr->Close() in the case where mConState == kConnectionState_NotConnected, that path will never result in the connection object being double-closed because mDeviceCon has been set to NULL in HandleConnectionClosed().

mburshteyn1:
@jaylogue got it, thanks for the explanation!

PR #286, opened 2019-07-02
remove copy log from tmp

jenniexie:
Given the changes from openweave-core PR#265 and happy PR#31 happy no longer puts the logs in /tmp and thus the copying logs from /tmp is no longer required.


jenniexie:
> I would encourage you to write more in the commit messages -- in this case, an appropriate thing to say is that given the changes from PR#XXX happy no longer puts the logs in /tmp and thus the copying logs from /tmp is no longer required.

Thanks RS for the feedback, I made changes for the commit messages

PR #298, opened 2019-07-09
Fixed Typo In The ConvertExtension() Function.

emargolis:
Replaced kTag_SubjectKeyIdentifier_Critical with kTag_KeyUsage_Critical
when converting Key Usage Extension.
This is not a functional bug because both enums have the same value "1".

### DIFF
--- a/src/lib/profiles/security/X509ToWeave.cpp
+++ b/src/lib/profiles/security/X509ToWeave.cpp
@@ -434,7 +434,7 @@ static WEAVE_ERROR ConvertExtension(ASN1Reader& reader, TLVWriter& writer)
 
                 if (critical)
                 {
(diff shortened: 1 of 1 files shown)


src/lib/profiles/security/X509ToWeave.cpp
```diff
@@ -434,7 +434,7 @@ static WEAVE_ERROR ConvertExtension(ASN1Reader& reader, TLVWriter& writer)
 
                 if (critical)
                 {
-                    err = writer.PutBoolean(ContextTag(kTag_SubjectKeyIdentifier_Critical), critical);
+                    err = writer.PutBoolean(ContextTag(kTag_KeyUsage_Critical), critical);
```
robszewczyk: Note to other reviewers: this was not an actual problem in running binaries, because both context tags equaled `1`.

PR #301, opened 2019-07-11
add wdm update integration tests

yunhanw-google:
-- add wdm conditional update integration test and unconditional update integration
test along with subscription.
-- add standalone wdm unconditional integration test without
subscription.
-- remove smoke check for wdm test since we should only rely on the log-check
in test application script, and reduce the log check confusion.
-- add NoSub enum value in WdmUpdateTiming, currently it include before
sub, during sub, and after sub, possibly it is good place to add no sub.
-- during no sub, MonitorClientCurrentState would be triggered when all
update iteracton completes.
-- with sub, update shutdown would be handled in existing sub shutdown
codepth.
-- delete the dead code in MockWdmSubscriptionInitiator.cpp and
MockWdmSubscriptionResponder.cpp.


jenniexie:
FYI: Qilin and Allen are going to help with WDM tests refactoring, I will let them know this PR.

PR #302, opened 2019-07-12
Software Update Manager Implementation

suryanshup:
-- Added new SoftwareUpdateManager class that defines an abstract
   interface for managing software updates.
-- Added new GenericSoftwareUpdateManagerImpl class that provides a
   default implementation of software update query. This class also
   implements a state machine that drives a software update through
   it's different stages such as query, download, integritycheck and
   image install.
-- Added new GenericSoftwareUpdateManagerImpl_BDX class that provides a
   generic implementation of software update file download for platforms
   that download software updates using the Weave Bulk Data Transfer
   Protocol.
-- Added a SoftwareUpdateManagerImpl class for nrf5 platform that
   implements part of the software update manager abstract interface
   specific to the nrf5 platform.
-- Software Update Manager can be disabled by setting
   WEAVE_DEVICE_CONFIG_ENABLE_SOFTWARE_UPDATE_MANAGER to 0 in
   WeaveDeviceConfig.h or in the project specific Weave Config file.


src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl_BDX.ipp
```diff
@@ -0,0 +1,264 @@
+/*
+ *
+ *    Copyright (c) 2018 Nest Labs, Inc.
```
suryanshup: Use new header.

src/adaptations/device-layer/ESP32/ESP32Config.cpp
```diff
@@ -55,6 +55,7 @@ const ESP32Config::Key ESP32Config::kConfigKey_GroupKeyIndex           = { kConf
 const ESP32Config::Key ESP32Config::kConfigKey_LastUsedEpochKeyId      = { kConfigNamespace_WeaveConfig,  "last-ek-id"         };
 const ESP32Config::Key ESP32Config::kConfigKey_FailSafeArmed           = { kConfigNamespace_WeaveConfig,  "fail-safe-armed"    };
 const ESP32Config::Key ESP32Config::kConfigKey_WiFiStationSecType      = { kConfigNamespace_WeaveConfig,  "sta-sec-type"       };
+const ESP32Config::Key ESP32Config::kConfigKey_CertBodyId              = { kConfigNamespace_WeaveConfig,  "cert-body-id"       };
```
jaylogue: I think we should remove this.

PR #304, opened 2019-07-12
Populate 802.15.4 MAC address on Nordic

robszewczyk:
On Nordic, populate 802.15.4 extended address. This address may be a
meaningless set of numbers -- there is no guarantee that the address
populated will be the same as the address used in joining (or
re-attaching) to any particular Thread network or that it will
correspond to the factory provisioned EUI64.

### DIFF
--- a/src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
+++ b/src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
@@ -807,6 +807,14 @@ exit:
 }
 
 template<class ImplClass>
+WEAVE_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::_GetPrimary802154MACAddress(uint8_t *buf)
+{
+    const otExtAddress *extendedAddr = otLinkGetExtendedAddress(mOTInst);
+    memcpy(buf, extendedAddr, sizeof(otExtAddress));
+    return WEAVE_NO_ERROR;
+};
+
+template<class ImplClass>
 WEAVE_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::DoInit(otInstance * otInst)
 {
     WEAVE_ERROR err = WEAVE_NO_ERROR;
@@ -870,7 +878,6 @@ bool GenericThreadStackManagerImpl_OpenThread<ImplClass>::IsThreadAttachedNoLock
     otDeviceRole curRole = otThreadGetDeviceRole(mOTInst);
     return (curRole != OT_DEVICE_ROLE_DISABLED && curRole != OT_DEVICE_ROLE_DETACHED);
 }
-
 
 } // namespace Internal
 } // namespace DeviceLayer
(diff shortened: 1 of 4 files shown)


g-coppock:
Does this value get surfaced to the apps anywhere? Curious what the implications are for us.

robszewczyk:
@g-coppock Apparently Nest app decides to provision Thread networks if the Identify response contains the Thread MAC address.  I've talked briefly with @jaylogue about that: this is not a great signal (after all, we're spending 8+ bytes to show that 802.15.4 radio is present; the MAC address itself is not really used anywhere else) and ultimately the pairing/OOB is a product-specific decision (and thus should be driven by Vendor ID/Product ID fields).  However, the existing code drives that decision from the MAC address, and its not wrong to include it.

g-coppock:
Got it, so this will be returned in the identify response as part of the WeaveDeviceDescriptor? Yeah it makes sense then that we're checking for that in order to determine whether or not to provision a thread network. This is part of what we discussed for supporting threadless devices since there's no other signal from the device whether or not the device includes a thread radio without having a mapping of vendor/product IDs. Thanks!

src/adaptations/device-layer/nRF5/ConfigurationManagerImpl.cpp
```diff
@@ -77,6 +79,13 @@ WEAVE_ERROR ConfigurationManagerImpl::_Init()
     return err;
 }
 
+WEAVE_ERROR ConfigurationManagerImpl::_GetPrimary802154MACAddress(uint8_t * buf)
+{
+    const otExtAddress *extendedAddr = otLinkGetExtendedAddress(ThreadStackMgrImpl().OTInstance());
```
jaylogue: While this approach works, it could be implemented more generically, such that GetPrimary802154MACAddress() automatically works on any platform that incorporates OpenThread.  E.g.: 

- GenericConfigurationManagerImpl<>::_GetPrimary802154MACAddress() could be rewritten to conditionally forward the call to a new interface method declared on ThreadStackManager; e.g.: ThreadStackMgr().GetPrimary802154MACAdddres().  The forwarding behavior would be conditional on WEAVE_DEVICE_CONFIG_ENABLE_THREAD being set; if not set, it would return WEAVE_DEVICE_ERROR_CONFIG_NOT_FOUND as it does not.

- GenericThreadStackManagerImpl_OpenThread<> would then provide a default implementation of the new ThreadStackManager method that works by calling otLinkGetExtendedAddress().

robszewczyk: Done, I've uploaded a new changeset

src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
```diff
@@ -806,6 +806,14 @@ exit:
     return err;
 }
 
+template<class ImplClass>
+WEAVE_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::_GetPrimary802154MACAddress(uint8_t *buf)
+{
+    const otExtAddress *extendedAddr = otLinkGetExtendedAddress(mOTInst);
+    memcpy(buf, extendedAddr, sizeof(otExtAddress));
+    return WEAVE_NO_ERROR;
```
gerickson: Style: new line before this line.

PR #315, opened 2019-07-18
add serial number in pairing test

yunhanw-google:
-- pass serial number to mock-device for pairing test

### DIFF
--- a/src/test-apps/happy/test-templates/WeavePairing.py
+++ b/src/test-apps/happy/test-templates/WeavePairing.py
@@ -2,7 +2,8 @@
 
 
 #
-#    Copyright (c) 2015-2017 Nest Labs, Inc.
+#    Copyright (c) 2019 Google, LLC.
+#    Copyright (c) 2015-2018 Nest Labs, Inc.
 #    All rights reserved.
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


jaylogue:
What is the point of this?  And why is it okay to use the device id as the serial number?

yunhanw-google:
> What is the point of this? And why is it okay to use the device id as the serial number?

It is for weave happy service registration test. Some service needs this serial number, otherwise, service returns the internal error.

We use weave node id as serial number for virtual service test, after running the test against the service, it looks ok.

src/test-apps/happy/test-templates/WeavePairing.py
```diff
@@ -324,6 +324,7 @@ def __start_device_side(self, device_info):
                 + " --vendor-id " + device_info['resource']['vendor_id'] \
                 + " --software-version " + '"'+ device_info['resource']['software_id'] + '"' \
                 + " --product-id " + device_info['resource']['product_id'] \
+                + " --serial-num " + device_info['device_weave_id'] \
```
jaylogue: I would prefer it if this code didn't always presume to use the device id as the serial number.  There should be some way for the specific test to pass in the serial number value, e.g. as part of device_info.  Only if that value is not specified should the code use the device id.

yunhanw-google: yes, we can setup the environment variable for specific test, saying "ENABALE_MOCK_DEVICE_SERIAL_NUMBER", and control if add --serial-number

However, I found that without the serial-num, the default one in mock-device become a string "mock-device", https://github.com/openweave/openweave-core/blob/master/src/test-apps/DeviceDescOptions.cpp#L81 which looks not good.

Further, can we generate the unique serial number in mock-device, which would not collide in the service, or we can generate serial number using weave node id as seed? why don' we just directly use weave node id as serial number during test? I have not yet observed the negative impact for testing.

Overall, it seems we can just pass weave node id(we store it in device_weave_id in test) for our test purpose, which is not harmful, and fixing the hard-coded "mock-device" string issue during test for emulated devices.

thanks

jaylogue: I'm fine with the code setting a unique serial number by default.  But your change forces that to _always_ be the case.  Please provide a way that specific tests can set a specific serial number value if they want to.

Environment variable doesn't make sense.  That's not how we pass the other values in.

yunhanw-google: Hi, Jay
Thanks for the suggestion.
we use same test script, test_weave_pairing_01.py to emulate all weave-supported devices and trigger service registration against service, we have passed several parameters from our continuous integration job to weave happy service tests via environment variable in the past, we use environment variable to choose which kind of device for weave pairing test, for example, https://github.com/openweave/openweave-core/blob/9af9a73c3bd2f62439e135484b7047d80cb8fcf9/src/test-apps/happy/test-templates/WeavePairing.py#L90

 I just update the pr, and add use_weave_id_as_serial_num to control if this specific device/test needs to set serial number. Since the weave node id is only unique element in our service test side, I am still using this as serial number for service testing.

https://github.com/openweave/openweave-core/pull/315/commits/9af9a73c3bd2f62439e135484b7047d80cb8fcf9#diff-e68c50b57fb852fc5c0f70bd75d79eaf

yunhanw-google:
It is ready for the review again, @jaylogue @robszewczyk 
thanks

PR #316, opened 2019-07-19
Feature/decouple happy serviceactmgr

akrithika27:


src/test-apps/happy/bin/weave-register-service.py
```diff
@@ -29,7 +29,6 @@
 import sys
 import set_test_path
 
-from happy.Utils import *
```
gerickson: If we are removing this file from having dependencies on Happy, then it needs to be moved out of src/test-apps/happy/bin/... and then needs to be moved somewhere else.

src/test-apps/happy/test-templates/ServiceAccountManager.py
```diff
@@ -32,14 +34,32 @@
 import grpc
 from grpc.framework.interfaces.face.face import ExpirationError
 from httplib import HTTPSConnection
-from happy.Driver import Driver
```
gerickson: If we are removing this file from having dependencies on Happy, then it needs to be moved out of src/test-apps/happy/bin/... and then needs to be moved somewhere else.

However, unlike the other "bin/..." file which I do not suspect is used anywhere else in the unit and functional tests, this is a test template that undergirds other service-related tests.

By gutting it of its happy dependencies, you are likely breaking its ability to fulfill its original goal and role as a test template. Consequently, you will likely need to do some refactoring to create a common base of functionality on which this test template and the tool you are trying to effect can depend.

akrithika27:
This creates a log file called serviceaccountmanager.log Attached the log to this PR. Also tested by importing ServiceAccountManager and running weave-register-service. Attached are the test logs.
[serviceaccountmanager.log](https://github.com/openweave/openweave-core/files/3412790/serviceaccountmanager.log)
[test_output.txt](https://github.com/openweave/openweave-core/files/3412791/test_output.txt)

src/test-apps/happy/bin/weave-register-service.py
robszewczyk: @gerickson  -- I agree in principle.  In the recent few weeks we've had a number of conversations around taking the service account manager and landing it in a more sane place.  The list of work items on that front included:

- refactoring the class to make it more independent from GRPC classes loaded from internet
- folding the class into some UI aligned with the weave device manager
- amending this code to work with other authentication schemes that are on the horizon for Google Nest Weave service.  

This set of changes was created opportunistically -- a third party collaborator asked to be able to run this script on MacOS.   The dependencies this script had on happy were superficial, and @akrithika27 stepped up to solve them.  

In practical terms: 
* we do need a design document the chart the direction for service account manager changes going forward. This is a substantial undertaking perhaps outside the scope of this surgical change.  I will take an action to drive the design doc, and we will review it shortly.
* The broader design will propose a better location for this script. Right now, a better place to locate these scripts would be either in `src/device-manager/python` (this begs the larger redesign decision) or in `src/test-apps/`; perhaps creating something like `src/test-apps/bin/` or `src/test-apps/scripts` .  If moving the scripts to either of the latter three locations makes the change more acceptable, I would completely agree and support that request. 
* I want to enable the third party collaborator without blocking on the larger changes.

In light of the above reasoning, I would suggest we approve the changes as they are (with perhaps moving the scripts to one of the locations listed above.

src/test-apps/happy/test-templates/ServiceAccountManager.py
robszewczyk: I'm not sure this is quite right -- this file has been misplaced to begin with, and does not fulfill the definition of a test template.  Regrettably, this area of our tests needs quite a bit of attention, down to the naming (I'm not sure that any of the classes there really fill the role of templates)

gerickson:
@robszewczyk, at minimum, let's get the files moved then. We can tackle the design-level changes later.

src/test-apps/happy/bin/weave-register-service.py
```diff
@@ -26,22 +26,38 @@
...
 
     try:
         opts, args = getopt.getopt(sys.argv[1:], "h:t:u:p:",
-            ["help", "tier=", "username=", "password=" "quiet"])
+                                   ["help", "tier=", "username=", "password=" "quiet"])
     except getopt.GetoptError as err:
-        print ServiceAccountManager.ServiceAccountManager.__doc__
-        print hred(str(err))
-        sys.exit(hred("%s: Failed server parse arguments." % (__file__)))
+        logger.debug(ServiceAccountManager.ServiceAccountManager.__doc__)
```
robszewczyk: These should be `print`ed rather than sent to `logger` -- the above code paths are executed in when we execute the script from as a top-line python main

akrithika27: Changed it to print

PR #317, opened 2019-07-22
Support brew prefix not equal to /usr/local

bukepo:
Homebrew is not always installed in /usr/local. This use standard way to
get Homebrew's prefix.

### DIFF
--- a/Makefile-Standalone
+++ b/Makefile-Standalone
@@ -111,8 +111,7 @@ ifeq ($(NO_OPENSSL),1)
 OPENSSL                         = no
 else
 ifeq ($(HOSTOS),darwin)
-OSX_OPENSSL_DIR                 = /usr/local/opt/openssl
-OPENSSL                        ?= $(shell if which brew > /dev/null && brew ls --versions openssl > /dev/null; then echo $(OSX_OPENSSL_DIR); else echo "internal"; fi) 
+OPENSSL                        ?= $(shell if which brew > /dev/null && brew ls --versions openssl > /dev/null; then brew --prefix openssl; else echo "internal"; fi)
 else
 OPENSSL                        ?= $(shell if which pkg-config > /dev/null && pkg-config --exists openssl; then echo ""; else echo "internal"; fi)
 endif


Makefile-Standalone
```diff
@@ -111,7 +111,7 @@ ifeq ($(NO_OPENSSL),1)
 OPENSSL                         = no
 else
 ifeq ($(HOSTOS),darwin)
-OSX_OPENSSL_DIR                 = /usr/local/opt/openssl
+OSX_OPENSSL_DIR                 = $(shell brew --prefix)/opt/openssl
```
gerickson: Brew may not be installed (as in my case), so I believe you'll need to be a bit more defensive in programming this as was the case demonstrated on line 115.

Makefile-Standalone
```diff
@@ -111,8 +111,8 @@ ifeq ($(NO_OPENSSL),1)
 OPENSSL                         = no
 else
 ifeq ($(HOSTOS),darwin)
-OSX_OPENSSL_DIR                 = /usr/local/opt/openssl
-OPENSSL                        ?= $(shell if which brew > /dev/null && brew ls --versions openssl > /dev/null; then echo $(OSX_OPENSSL_DIR); else echo "internal"; fi) 
+OSX_OPENSSL_DIR                 = $(shell if which brew > /dev/null; then brew --prefix; else echo /usr/local; fi)/opt/openssl
+OPENSSL                        ?= $(shell if which brew > /dev/null && brew ls --versions openssl > /dev/null; then echo $(OSX_OPENSSL_DIR); else echo "internal"; fi)
```
jaylogue: Rather than changing OSX_OPENSSL_DIR, I suggest eliminating OSX_OPENSSL_DIR altogether and replacing the "then" clause in the OPENSSL shell expression with \`brew --prefix openssl\`
E.g.:

```
if which brew > /dev/null && brew ls --versions openssl > /dev/null; then `brew --prefix openssl`; else echo "internal"; fi)
```
This way brew tells us the full install path, but only when its known to be present on the system.

bukepo: great!

PR #325, opened 2019-07-25
Documentation for factory provisioning tool, plus minor feature enhancements

jaylogue:
-- Created overview documentation for the OpenWeave factory provisioning tool describing its theory of operation and its user interface.

-- Enhanced the factory provisioning tool to allow any provisioning value to be specified in a provisioning data CSV file.  This means that the supported CSV file format is a now superset of that produced by the weave tools’s “gen-provisioning-data” command.

-- Renamed the default CSV column names to use underscores instead of spaces.  Note that the provisioning tool is still backwards compatible with CSV files using the old format column names.

### DIFF
--- a/doc/general/weave-factory-provisioning-tool.md
+++ b/doc/general/weave-factory-provisioning-tool.md
@@ -1,0 +1,171 @@
+# OpenWeave Factory Provisioning Tool
+
+The OpenWeave factory provisioning tool provides a convenient means to provisioning persistent, per-device configuration information
+onto Weave-enabled devices.  The factory provisioning tool is intended to be used as part of a manufacturing line process that stamps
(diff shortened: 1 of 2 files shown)


doc/general/weave-factory-provisioning-tool.md
```diff
@@ -0,0 +1,160 @@
+# Introduction
+
+The OpenWeave factory provisioning tool provides a convenient means to provisioning persistent, per-device configuration information
+onto Weave-enabled devices.  The factory provisioning tool is intended to be used as part of a manufacturing line process that stamps
+individual devices with unique identity and credential information prior to shipment to customers.  It can also be used by  developers to
```
gerickson: Two spaces between "by" and "developers".

PR #326, opened 2019-07-26
Name an empty union to build with clang 9

davidgao:
This empty anonymous union builds correctly with clang 8, but clang 9
would emit a warning 'declaration does not declare anything', and would
prevent weave from building because we have '-Werror' on.

To keep the interface consistent, the union is temporarily named rather
than removed. Please be sure to remove the name when adding members.

### DIFF
--- a/src/lib/profiles/data-management/Current/UpdateClient.h
+++ b/src/lib/profiles/data-management/Current/UpdateClient.h
@@ -122,7 +122,7 @@ struct UpdateClient::InEventParam
 struct UpdateClient::OutEventParam
 {
     bool DefaultHandlerCalled;
-    union
+    union _ // make clang 9 happy. remove name when adding members.
     {
     };
     void Clear() { memset(this, 0, sizeof(*this)); }


src/lib/profiles/data-management/Current/UpdateClient.h
```diff
@@ -122,7 +122,7 @@ struct UpdateClient::InEventParam
 struct UpdateClient::OutEventParam
 {
     bool DefaultHandlerCalled;
-    union
+    union _ // make clang 9 happy. remove name when adding members.
```
jaylogue: Could we instead add an empty member struct for one of the API events?

```
union
{
    struct
    {
    } UpdateComplete; // Empty structure to avoid gratuitous clang warning about unions without members.
};
```
Will this not satisfy the compiler?

davidgao: Yes, this compiles.
Just a reminder, an empty struct's size in C++ is not 0. Let's apply this one if it does not break anything.

robszewczyk:
Closing the pull request in favor of a different way of solving the same underlying issue.

PR #327, opened 2019-07-26
Updated IPv6InterfaceId to WeaveNodeId Conversion Functions.

emargolis:
Historically Weave NodeIds have always been universal (globally unique) identifiers.
Therefore, it was sufficient to set or clear the Universal/Local (U/L) bit when converting
WeaveNodoId to IPv6InterfaceId and vise versa.
In preparation for supporting also local identifiers this implementation has
been updated to invert U/L bit when converting, per RFC-3513.

### DIFF
--- a/src/lib/core/WeaveMessageLayer.h
+++ b/src/lib/core/WeaveMessageLayer.h
@@ -793,6 +793,36 @@ inline bool WeaveMessageLayer::IsBoundToLocalIPv6Address(void) const
 }
 
 /**
+ * Bit field definitions for IEEE EUI-64 Identifiers.
+ */
+enum
+{
+    kEUI64_UL_Mask          = 0x0200000000000000ULL,        /**< Bitmask for the Universal/Local (U/L) bit within an EUI-64 identifier.
+                                                                 A value of 0 indicates the id is Universally (globally) administered.
+                                                                 A value of 1 indicates the id is Locally administered. */
+
+    kEUI64_UL_Unversal      = 0,                            /**< Universal/Local bit value indicating a Universally administered EUI-64 identifier. */
+
+    kEUI64_UL_Local         = kEUI64_UL_Mask,               /**< Universal/Local bit value indicating a Locally administered EUI-64 identifier. */
+
+    kEUI64_IG_Mask          = 0x0100000000000000ULL,        /**< Bitmask for the Individual/Group (I/G) bit within an EUI-64 identifier.
+                                                                 A value of 0 indicates the id is an individual address.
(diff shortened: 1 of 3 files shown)


src/lib/core/WeaveMessageLayer.h
```diff
@@ -792,6 +792,30 @@ inline bool WeaveMessageLayer::IsBoundToLocalIPv6Address(void) const
...
+{
+    kEUI64_UL_Mask          = 0x0200000000000000ULL,        /**< Bitmask for the Universal/Local (U/L) bit within an EUI-64 identifier.
+                                                                 A value of 0 indicates the id is Universally (globally) administered.
+                                                                 A value of 1 indicates the id is Locally administered. */
+
+    kEUI64_UL_Unversal      = 0,                            /**< Universal/Local bit value indicating a Universally administered EUI-64 identifier. */
+
+    kEUI64_UL_Local         = kEUI64_UL_Mask,               /**< Universal/Local bit value indicating a Locally administered EUI-64 identifier. */
+
+    kEUI64_UL_MaxTestLocal  = 65535,                        /**< EUI-64 identifiers less or equal than this value are considered local for testing convenience. */
```
jaylogue: I think it would be better to define this in terms of a special range of Weave Node Ids, rather than EUI-64s.  So maybe make this its own enum with a name like kMaxAlwaysLocalWeaveNodeId.

emargolis: Done.

src/lib/core/WeaveMessageLayer.h
```diff
@@ -827,14 +857,14 @@ inline uint64_t WeaveFabricIdToIPv6GlobalId(uint64_t fabricId) { return (fabricI
...
 
 /**
  *  Convert a Weave node identifier to an IPv6 address interface identifier.
  *
- *  Weave node identifiers are global EUI-64s, which per RFC-3513 are converted to interface
- *  identifiers by setting the universal/local bit to 1 (bit 57 counting the LSB as 0).
+ *  Weave node identifiers are Universal/Local EUI-64s, which per RFC-3513 are converted to
+ *  interface identifiers by inverting the universal/local bit (bit 57 counting the LSB as 0).
  *
  *  As a convenience to testing, node identifiers less than 65536 are considered 'local',
```
gerickson: It might be helpful to the reader of comments-only to include the mnemonic `#kMaxAlwaysLocalWeaveNodeId` alongside the "less than 65536".

emargolis: done

PR #330, opened 2019-07-29
[Build] Remove automake and autoconf generated files

gjc13:
Intermediates generated by autoconf and automake were removed.
Also added these files to gitignore.


gerickson:
> Intermediates generated by autoconf and automake were removed.
> Also added these files to gitignore.

We intentionally check-in these bootstrap intermediate files to preclude users from having to run bootstrap. Consequently, bootstrap is largely a maintainer-only operation today.

Therefore, we don't want to `git rm` these and do not want to add them to `.gitiginore`.

gjc13:
> We intentionally check-in these bootstrap intermediate files to preclude users from having to run bootstrap. Consequently, bootstrap is largely a maintainer-only operation today.
> 
> Therefore, we don't want to `git rm` these and do not want to add them to `.gitiginore`.

I got the idea. I'm trying to add a new file and when I regenerate the build intermediates, I found they all changed due to different autotools version. Mine is 1.15 while the current toolchain version is 1.14.
Shall I downgrade my autotools or I can just update the intermediates, much thanks.

gerickson:
There are two options: 1) see https://github.com/openthread/openthread/blob/master/third_party/nlbuild-autotools/repo/README.md#build-and-install-build_and_install or 2) just check in the new intermediate files as a whole.

PR #332, opened 2019-07-30
enable lwip tunnel test: test_weave_tunnel_01.py

jenniexie:
enabling lwip tunnel test: test_weave_tunnel_01.py

### DIFF
--- a/src/test-apps/happy/tests/standalone/tunnel/test_weave_tunnel_01.py
+++ b/src/test-apps/happy/tests/standalone/tunnel/test_weave_tunnel_01.py
@@ -27,6 +27,7 @@ import sys
 import unittest
 import set_test_path
 import time
+import subprocess
 
 from happy.Utils import *
 import happy.HappyNodeList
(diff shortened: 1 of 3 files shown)


src/test-apps/happy/tests/standalone/tunnel/test_weave_tunnel_01.py
```diff
@@ -38,47 +39,33 @@
...
         if "WEAVE_SYSTEM_CONFIG_USE_LWIP" in os.environ.keys() and os.environ["WEAVE_SYSTEM_CONFIG_USE_LWIP"] == "1":
-            self.topology_file = os.path.dirname(os.path.realpath(__file__)) + \
-                "/../../../topologies/standalone/thread_wifi_on_tap_ap_service.json"
-            self.tap = "wpan0"
+            self.use_lwip = True
+            topology_shell_script = os.path.dirname(os.path.realpath(__file__)) + \
+                "/../../../topologies/standalone/thread_wifi_on_tap_ap_service.sh"
+            # tap interface, ipv4 gateway and node addr should be provided if device is tap device
+            # both BorderRouter and cloud node are tap devices here
+            self.BR_tap = "wlan0"
```
gerickson: We should not be hardcoding values like this in tests. This should come from or be derived from topology files.

jenniexie: > We should not be hardcoding values like this in tests. This should come from or be derived from topology files.

Yes, I agree with that. I will see if I can use existing APIs to get the values or add new APIs for that.

jenniexie: Hi Grant, now I looked at these parameters, these values are actually not available from happy topology, we need to pass it from test app command line as parameters to lwip stack. so it is not available from topology file.

gerickson: Then we may need to work on the Happy APIs.

jenniexie: I think we should be able to do it in happy, where we save the data just as values for lwip without actually setting the value using netns commands

@robszewczyk how do you think?

If it sounds good, I can add a draft design doc for review.

PR #335, opened 2019-08-01
Make kStatus_AccessDenied visible from objc device-manager.

lcniles:
Add translated error constant for `nl::Weave::Profiles::Common::kStatus_AccessDenied` to the objc device-manager.

This constant is needed for Nest app error routing.

### DIFF
--- a/src/device-manager/cocoa/NLProfileStatusError.mm
+++ b/src/device-manager/cocoa/NLProfileStatusError.mm
@@ -113,6 +113,8 @@ using namespace nl::Weave::Profiles;
             return kNLStatus_InternalError;
         case nl::Weave::Profiles::Common::kStatus_Continue:
             return kNLStatus_Continue;
+        case nl::Weave::Profiles::Common::kStatus_AccessDenied:
+            return kNLStatus_AccessDenied;
         default:
             return self.statusCode;
         }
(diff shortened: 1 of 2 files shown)


lcniles:
> I note that the parallel code does not exist in the parallel implementation for Java clients. Should it? Would it make the job of the integrator easier? It appears in Java, the equivalent code is packaged up in an exception that contains the error code; we do not provide the error codes in any consistent enumeration for Java clients.

I don't know the design reason for not passing the constants strait through for objc. Clang seems to handle the namespacing just fine.

gerickson:
> > I note that the parallel code does not exist in the parallel implementation for Java clients. Should it? Would it make the job of the integrator easier? It appears in Java, the equivalent code is packaged up in an exception that contains the error code; we do not provide the error codes in any consistent enumeration for Java clients.
> 
> I don't know the design reason for not passing the constants strait through for objc. Clang seems to handle the namespacing just fine. Perhaps that wasn't always true.

What would it take to unwind this practice and just pass them straight through in Objective C++?

robszewczyk:
> > > I note that the parallel code does not exist in the parallel implementation for Java clients. Should it? Would it make the job of the integrator easier? It appears in Java, the equivalent code is packaged up in an exception that contains the error code; we do not provide the error codes in any consistent enumeration for Java clients.
> > 
> > 
> > I don't know the design reason for not passing the constants strait through for objc. Clang seems to handle the namespacing just fine. Perhaps that wasn't always true.
> 
> What would it take to unwind this practice and just pass them straight through in Objective C++?

@gerickson @lcniles -- I believe this decision may have been made for packaging reasons.  Once we have an objective C header, it can be freely included across the source base in the app, without regard to the source files being objective C, objective C++ etc.  I believe the packaging of Weave headers within the app development frameworks was a bit tricky, and the original author decided that packaging only the Cocoa headers would greatly simplify things.

It does seem like @lcniles is probably the best person to fix this, if he wants to take that on. 

@gerickson -- are you also suggesting that if we take the step to align the status codes between the C++ and Objective C++, then we should also create appropriate enumerations in Java land? (right now status codes seem to be just magic numbers on the other side of the Java API)

lcniles:
> It does seem like @lcniles is probably the best person to fix this, if he wants to take that on.

sgtm. We can probably start this kind of work in Q4. I'll reach out at that time and we can put together a plan.

PR #336, opened 2019-08-05
Cleanup phoenix files

akrithika27:
Uploaded only those files that are needed by ServiceAccountManager.py. This elimiates the use of sys.path inside the script and we do not have to download the entire phoneix developer tools. The path to the phoenix imports should be set in the environment. 
Attached test script and output
[test_details.txt](https://github.com/openweave/openweave-core/files/3469348/test_details.txt)


src/test-apps/happy/test-templates/ServiceAccountManager.py
```diff
@@ -43,69 +37,25 @@
...
 
-    def __init__(self, tier, username, password, token,
-                 schema_branch=default_schema_branch):
+    def __init__(self, tier, username, password, token):
         self.tier = tier
         self.username = username
         self.password = password
 
-        use_phoenix_schema(schema_branch)
         import nestlabs.gateway.v2.gateway_api_pb2 as gateway_api_pb2
```
robszewczyk: please push those imports to the top of the file

PR #338, opened 2019-08-20
Fix high power usage in the unprovisioned state

robszewczyk:
In order to put OpenThread into a low power mode, one must disable
both the Thread module as well as the Ip6 module in OpenThread.
Modify SetThreadEnable() functions accordingly.  However, the
initialization of the OpenThread stack does not proceed as expected
without enabling Ip6 at initialization; consequently, we enable Ip6 on
init, and disable subsequently on any Thread state change that would
disable the Thread stack.

### DIFF
--- a/src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
+++ b/src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
@@ -150,15 +150,25 @@ void GenericThreadStackManagerImpl_OpenThread<ImplClass>::_OnPlatformEvent(const
 {
     if (event->Type == DeviceEventType::kThreadStateChange)
     {
-#if WEAVE_DETAIL_LOGGING
-
         Impl()->LockThreadStack();
+
+#if WEAVE_DETAIL_LOGGING
 
         LogOpenThreadStateChange(mOTInst, event->ThreadStateChange.OpenThread.Flags);
 
+#endif // WEAVE_DETAIL_LOGGING
+
+        // The API we provide to enable/disable thread controls both the Thread
+        // stack and the IP6 stack within OpenThread.  In order to properly
+        // initialize OpenThread, we need to enable IP6 in OT, and we can
+        // disable it once the OT stack is up and running and signaling events.
+        if ((otThreadGetDeviceRole(mOTInst) == OT_DEVICE_ROLE_DISABLED) && otIp6IsEnabled(mOTInst))
+        {
+            otIp6SetEnabled(mOTInst, false);
+        }
+
         Impl()->UnlockThreadStack();
 
-#endif // WEAVE_DETAIL_LOGGING
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
```diff
@@ -182,11 +191,27 @@ WEAVE_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::_SetThreadEnabl
     Impl()->LockThreadStack();
 
     bool isEnabled = (otThreadGetDeviceRole(mOTInst) != OT_DEVICE_ROLE_DISABLED);
+    bool isIp6Enabled = otIp6IsEnabled(mOTInst);
+
+    if (val && !isIp6Enabled)
+    {
+        otErr = otIp6SetEnabled(mOTInst, val);
+        VerifyOrExit(otErr == OT_ERROR_NONE, );
```
suryanshup: SuccessOrExit(otErr) instead ?

src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
```diff
@@ -150,15 +150,24 @@ void GenericThreadStackManagerImpl_OpenThread<ImplClass>::_OnPlatformEvent(const
...
+        // The API we provide to enable/disable thread controls both the Thread
+        // stack and the IP6 stack within OpenThread.  In order to properly
+        // initialize OpenThread, we need to enable IP6 in OT, and we can
+        // disable it once the OT stack is up and running and signaling events.
+        if ((otThreadGetDeviceRole(mOTInst) == OT_DEVICE_ROLE_DISABLED) && otIp6IsEnabled(mOTInst))
+        {
+            otIp6SetEnabled(mOTInst, false);
+        }
 
 #endif // WEAVE_DETAIL_LOGGING
```
suryanshup: Should move 
`#endif // WEAVE_DETAIL_LOGGING`
after line 158 otherwise the check in line 163 will get disabled if WEAVE_DETAIL_LOGGING is turned off.

src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
robszewczyk: I would absolutely agree if this was a `WEAVE_ERROR`.  For `OT_ERROR` I went with this version.

src/adaptations/device-layer/include/Weave/DeviceLayer/OpenThread/GenericThreadStackManagerImpl_OpenThread.ipp
robszewczyk: Yup, you're right. fixed.

PR #344, opened 2019-08-23
Remove the failing pairing/test_weave_ble_01.py test

robszewczyk:
It appears that Travis CI no longer supports the BLE-based tests --
the tests fail to start the Bluetooth daemon, the daemon fails with:

    00:24:42.194097 socket(PF_BLUETOOTH, SOCK_RAW, 1) = -1 EAFNOSUPPORT (Address family not supported by protocol)

We are disabling the test until such time as we have a suitable
workaround.

### DIFF
--- a/src/test-apps/Makefile.am
+++ b/src/test-apps/Makefile.am
@@ -1089,7 +1089,6 @@ if WEAVE_RUN_HAPPY_PAIRING
 if CONFIG_NETWORK_LAYER_BLE
 if CONFIG_BLE_PLATFORM_BLUEZ
 check_SCRIPTS                                 +=                \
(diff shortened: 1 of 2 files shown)


src/test-apps/Makefile.am
```diff
@@ -1089,7 +1089,6 @@ if WEAVE_RUN_HAPPY_PAIRING
 if CONFIG_NETWORK_LAYER_BLE
 if CONFIG_BLE_PLATFORM_BLUEZ
```
gerickson: Is there a reason we cannot just disable CONFIG_BLE_PLATFORM_BLUEZ?

PR #345, opened 2019-08-23
Documentation Style Guide update and improved Copybara support

Vyrastas:
- Updated the Documentation Style Guide to cover document locations, heading formats, site menus, code blocks, and Material icon support
- Added _toc.yaml files so openweave.io menu updates happen on GitHub as docs change
- Moved and copyedited the Factory Provisioning Tool guide for openweave.io import support and style consistency: removed extraneous blank spaces, headers formatted (levels and sentence case), replaced abbreviations, code block tweaks
- Minor edits to existing docs for improved openweave.io import support going forward


doc/STYLE_GUIDE.md
```diff
@@ -131,3 +176,33 @@ For example:
...
+
+### Material icons
+
+The openweave.io documentation site uses various Material icons inline with the text or in
+diagrams to represent elements of the Weave system. These icons are used to aid in your
+understanding of Weave by highlighting common elements and are not official Weave-branded
+icons.
+
+> You are not required to use these icons, but we may ask to insert them during the review
+process, in order to tie content together across the documentation site.
```
jaylogue: These don't format correctly on github, so we need to find a way to incorporate them that doesn't product confusing output on standard markdown sites.

Vyrastas: Yeah to do these there has to be some sort of indicator in the source to know which to convert... the only thing I can think that doesn't have an additional visible character is:

underscore asterisk term asterisk underscore

_*term*_

This renders the word in italics in Markdown and is not typically used to format italics.  If it was used, unless the term is on the supported icon list, nothing will happen when imported.  Does that work?

Other than that, we'd have to have some sort of visible delimiter to indicate it's a special term.  I think they're worth accounting for as we build content out around schema and profiles.

PR #347, opened 2019-08-27
Add a new function to validate a device identify request.

rogertinkoff:
Add a new Weave error WEAVE_ERROR_INVALID_DEVICE_DESCRIPTOR.

### DIFF
--- a/src/device-manager/WeaveDeviceManager.cpp
+++ b/src/device-manager/WeaveDeviceManager.cpp
@@ -357,6 +357,21 @@ exit:
     return err;
 }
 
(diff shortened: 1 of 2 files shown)


jaylogue:
> What are the appropriate unit / functional tests to create to validate this new functionality?

@gerickson : The error handling case introduced by this change can only be provoked manually or in the context of Happy tests that script the Weave Device Manager.  I have confirmed the correct behavior via manual testing.  However, currently there is no infrastructure in the Happy test suites for testing argument handling in the Device Manager.  Given the priority of this issue, I suggest we accept this change as-is.

PR #356, opened 2019-09-05
Minor fix to the configuration system

robszewczyk:
Previous code caused several flags to be included multiple times. The
new code makes it clear what is being included in the CFLAGS and
CXXFLAGS when BlueZ peripheral code is enabled in the context of
AX_CHECK_COMPILER_FLAGS.


jaylogue:
So the multiple inclusion of flags was only a problem when bluez support was enabled?

robszewczyk:
> So the multiple inclusion of flags was only a problem when bluez support was enabled?

Correct.  All other instances of calls to AX_CHECK_COMPILER_FLAGS did the correct thing and did not call the macro with cumulative flags

PR #359, opened 2019-09-06
minor fix to support weave address on service node

jenniexie:
In lwip service tunnel test 02, we do weave-ping test between thread node and service(cloud) node, which will require both thread node and service node to have weave address.

This PR is to fix 2 minor issues that:
1. weave test infra does not add weave address if it is a service node.
2. node interface type for service node should be "wan", without correct interface type, service node is not getting correct weave address subnet.

### DIFF
--- a/src/test-apps/happy/lib/WeaveNodeConfigure.py
+++ b/src/test-apps/happy/lib/WeaveNodeConfigure.py
@@ -331,8 +331,14 @@ class WeaveNodeConfigure():
             self.logger.error("[localhost] WeaveNodeConfigure: %s" % emsg)
             return
 
-        if self.weave_state.getNodeType(node) in ['ap', 'service']:
-            emsg = "ignoring initialization for nodes types 'ap' and 'service'"
+        # configure weave state including weave address for non-AP nodes,
+        # because current infra only support AP joined as a non-weave device, 
+        # all other nodes can be joined as weave devices.
+        # ToDo: Actually, AP should be able to joined as weave device or non-weave device, 
+        # Some infra work need to be sorted out to support that.
+        # Tracked issue link: https://github.com/openweave/openweave-core/issues/360
+        if self.weave_state.getNodeType(node) in ['ap']:
+            emsg = "ignoring initialization for nodes types 'ap'"
             self.logger.debug("[localhost] WeaveNodeConfigure: %s" % emsg)
             return
 
--- a/src/test-apps/happy/lib/WeaveState.py
+++ b/src/test-apps/happy/lib/WeaveState.py
@@ -238,17 +238,32 @@ class WeaveState(State):
 
         return global_prefix
 
-    def typeToWeaveSubnet(self, type):
-        if type == "wifi":
+    def typeToWeaveSubnet(self, interface_type):
+        """This function attempts to match node interface type to a matching weave subnet.
+
+        Args:
+           interface_type (str): A string containing a node interface type, example: "wifi" or "wan".
+
+        Returns:
+           integer. The return code:
+
+              1  -- weave subnet is 1 which matches a wifi interface.
+              5  -- weave subnet is 5 which matches a wan interface.
+              4  -- weave subnet is 4 which matches a mobile interface.
+              6  -- weave subnet is 6 which matches a thread interface.
+              0  -- weave subnet is 0 which matches all other interfaces.
+        """
+
+        if interface_type == "wifi":
             return 1
 
-        if type == "service":
+        if interface_type == "wan":
             return 5
 
-        if type == "mobile":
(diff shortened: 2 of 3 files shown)


src/test-apps/happy/lib/WeaveState.py
```diff
@@ -242,7 +242,7 @@ def typeToWeaveSubnet(self, type):
         if type == "wifi":
             return 1
 
-        if type == "service":
+        if type == "wan":
```
gerickson: Was this a previous typo?

src/test-apps/happy/lib/WeaveNodeConfigure.py
```diff
@@ -331,7 +331,7 @@ def __init_node(self, node, node_params={}):
             self.logger.error("[localhost] WeaveNodeConfigure: %s" % emsg)
             return
 
-        if self.weave_state.getNodeType(node) in ['ap', 'service']:
+        if self.weave_state.getNodeType(node) in ['ap']:
```
gerickson: Can you explain the nature of this change? Why was it wrong before; how is it correct now?

src/test-apps/happy/lib/WeaveState.py
jenniexie: I think original owner got confused between node type and interface type, and we never used this part of code to assign weave address to service node before.

"service" is a type value of node types, "wan" is a type value of interface types.

gerickson: I would say that then warrants a comment towards that end at the top of the "if" block for interface types. Documentation-wise, when we gain key insights or step on land mines from past authors, we should leave the code in better shape than we found it.

src/test-apps/happy/lib/WeaveNodeConfigure.py
jenniexie: sure, basically this "If" code here will decide if the weave test infra will assign weave address to a node or not. With original code, if a node is service node, it will not assign weave address to the node. Since we need weave address for service node for one of our lwip tunnel test case, so we should remove "service" from the check and service node will get a proper weave address assigned to it.

src/test-apps/happy/lib/WeaveNodeConfigure.py
```diff
@@ -331,7 +331,7 @@ def __init_node(self, node, node_params={}):
...
             emsg = "ignoring initialization for nodes types 'ap' and 'service'"
```
gerickson: Comment and message no longer matches the code.

src/test-apps/happy/lib/WeaveNodeConfigure.py
gerickson: Thanks; that should be clarified in a comment above this block.

src/test-apps/happy/lib/WeaveState.py
jenniexie: sure, will do, Thanks for feedback :)

src/test-apps/happy/lib/WeaveNodeConfigure.py
jenniexie: good eyes:)
I will fix it

src/test-apps/happy/lib/WeaveNodeConfigure.py
jenniexie: OK, will add comment for it.

src/test-apps/happy/lib/WeaveState.py
```diff
@@ -238,17 +238,32 @@ def getWeaveGlobalPrefix(self, fabric_id):
 
         return global_prefix
 
-    def typeToWeaveSubnet(self, type):
-        if type == "wifi":
+    def typeToWeaveSubnet(self, interface_type):
```
jenniexie: I changed type to interface_type
type is Python key word, we should avoid to use it in our code.

src/test-apps/happy/lib/WeaveNodeConfigure.py
```diff
@@ -331,8 +331,9 @@ def __init_node(self, node, node_params={}):
             self.logger.error("[localhost] WeaveNodeConfigure: %s" % emsg)
             return
 
-        if self.weave_state.getNodeType(node) in ['ap', 'service']:
-            emsg = "ignoring initialization for nodes types 'ap' and 'service'"
+        # configure weave state including weave address for non-ap nodes
```
gerickson: Great start, but why non-ap nodes?

src/test-apps/happy/lib/WeaveState.py
gerickson: Thanks; much improved!

PR #362, opened 2019-09-10
lwip weave tunnel test 02

jenniexie:
enabled lwip tunnel test 02


gerickson:
> enabled lwip tunnel test 02
> manual test steps are tracked in this doc:
> https://docs.google.com/document/d/1XRXu2wNzAed3tEAYU6eloFx_gHiYj6Ysy62ObrZWgIo/edit?usp=sharing

@jenniexie, thanks for sharing broader context; however, non-Google Github users cannot access such documents. I would recommend creating a Github issue, and managing the content there.

src/test-apps/happy/lib/WeaveState.py
```diff
@@ -168,6 +168,30 @@ def getNodeWeaveIPAddress(self, node_id=None, state=None):
                 return addr
         return None
 
+    def getNodeInterfaceWeaveIPAddress(self, interface_id, node_id, state=None):
```
gerickson: How does this control / influence whether that address is IPv4 or IPv6. Ostensibly that address should *only* be IPv6; however, the documentation should be clear about that fact and the code should, potentially, have an assertion or some such to trap / catch an IPv4 address that might "bubble up" through to this interface.

jenniexie: Hi Grant:
Do you mean add a check to make sure there is ipv6 address in the node's address list before we do IP.prefixMatchAddress(weave_global_prefix, addr) in line 191?

This function will return weave ip address, which should be ipv6 address by default from my understanding.

jenniexie:
> > enabled lwip tunnel test 02
> > manual test steps are tracked in this doc:
> > https://docs.google.com/document/d/1XRXu2wNzAed3tEAYU6eloFx_gHiYj6Ysy62ObrZWgIo/edit?usp=sharing
> 
> @jenniexie, thanks for sharing broader context; however, non-Google Github users cannot access such documents. I would recommend creating a Github issue, and managing the content there.

Sure, I will remove the google doc and add an issue with information there.

jenniexie:
Merging #362 into master will increase coverage by 1.61%.
This test case will improve code coverage a little bit.

PR #363, opened 2019-09-11
Ensure WeaveTunnelConnectionManager is in correct state before trying to reconnect when timer fires.

pidarped:
When the reconnect timer fires, ensure that the WeaveConnectionManager is in the correct 'NotConnected' state before attempting the reconnect.

Previously, there was only a check and exit if the state was wrong.

### DIFF
--- a/src/test-apps/TestWeaveTunnelBR.cpp
+++ b/src/test-apps/TestWeaveTunnelBR.cpp
@@ -1107,7 +1107,7 @@ static void TestTunnelNoStatusReportResetReconnectBackoff(nlTestSuite *inSuite, 
     gTestSucceeded = false;
     /* Set the test timeout to be a little longer than the Tunnel Control ExchangeContext
      * timeout */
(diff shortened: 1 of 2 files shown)


src/test-apps/TestWeaveTunnelBR.cpp
```diff
@@ -1107,7 +1107,7 @@ static void TestTunnelNoStatusReportResetReconnectBackoff(nlTestSuite *inSuite,
     gTestSucceeded = false;
     /* Set the test timeout to be a little longer than the Tunnel Control ExchangeContext
      * timeout */
-    gMaxTestDurationMillisecs = (WEAVE_CONFIG_TUNNELING_CTRL_RESPONSE_TIMEOUT_SECS + 1 ) * System::kTimerFactor_milli_per_unit;
+    gMaxTestDurationMillisecs = (WEAVE_CONFIG_TUNNELING_CTRL_RESPONSE_TIMEOUT_SECS + 5 ) * System::kTimerFactor_milli_per_unit;
```
yunhanw-google: maybe use some macro naming to represent the 1 or 5,

pidarped: This time was just a guard time for the test completion and so I left it hardcoded. Increased it slightly because I noticed the test sometimes taking slightly longer.

PR #370, opened 2019-09-13
Fixed bug in nRF52 ConnectivityManager::SetBLEAdvertisingEnabled()

jaylogue:
[ This change addresses b/140955318. ]

-- Fixed a logic bug in ConnectivityManager::SetBLEAdvertisingEnabled() where passing false would not actually stop BLE advertising.

-- Added log message when BLE advertising stopped.

### DIFF
--- a/src/adaptations/device-layer/nRF5/BLEManagerImpl.cpp
+++ b/src/adaptations/device-layer/nRF5/BLEManagerImpl.cpp
@@ -386,8 +386,8 @@ void BLEManagerImpl::DriveBLEState(void)
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/adaptations/device-layer/nRF5/BLEManagerImpl.cpp
```diff
@@ -387,7 +387,7 @@ void BLEManagerImpl::DriveBLEState(void)
     }
 
     // Otherwise, stop advertising if it is enabled.
```
gerickson: Probably good to change the comment to align with the code.

PR #372, opened 2019-09-18
add lcov report

jenniexie:
added code coverage using lcov for github openweave


gerickson:
While this isn't harmful, given that we have [codecov.io](https://codecov.io/gh/openweave/openweave-core) integration for openweave-core, is there value for this addition?

jenniexie:
> While this isn't harmful, given that we have [codecov.io](https://codecov.io/gh/openweave/openweave-core) integration for openweave-core, is there value for this addition?
Hi Grant:
    By looking at our internal lcov report in Jenkins job, it has line, function and branch coverage, vs github codecov.io only has line coverage.
Since it has more coverage data, should we have the data for openweave-core on github?

PR #379, opened 2019-10-03
add GenericTraitCatalogImp

yunhanw-google:


src/lib/profiles/data-management/Current/GenericTraitCatalogImp.h
```diff
@@ -0,0 +1,535 @@
+/*
+ *
+ *    Copyright (c) 2019 Google, Inc.
```
gerickson: See other files; officially it is Google LLC.

yunhanw-google: sure, I would update the description, this would be used in data management support on mobile client side

PR #381, opened 2019-10-06
Support for UDP source address control

jaylogue:
-- Modified UDPEndPoint and IPEndPointBasis to allow the source address for an outbound packet to be specified by the sender.  Both classes now feature a new SendMsg() method which takes an IPPacketInfo structure containing the destination information (address and port) and, optionally, a source address.  (The existing SendTo() methods on UDPEndPoint have been retained for API compatibility).  

When specified, the source address overrides the normal IP source address selection process performed by the network stack.  On sockets based systems, this feature works by passing a IP_PKTINFO/IPV6_PKTINFO "control message" to the kernel.  On LwIP, source address selection is overridden by temporarily rebinding the local address of the UDP PCB.

-- Adjusted RawEndPoint to work with the above changes to IPEndPointBasis and added a SendMsg() method to keep the interface similar to UDPEndPoint (although source address control is not needed for raw end points).

-- Moved method documentation into .cpp files to reduce clutter in class definitions.


jaylogue:
> What unit / functional tests can / should be added to ensure this behaves as intended and that no future regressions are introduced?

There are no functional tests for UDPEndPoint in the current test set.   We should rectify this, but I don't have the bandwidth to take this on in this context.  This feature will be used, however, in an upcoming change to Weave for which I do plan to add tests.

PR #384, opened 2019-10-06
Fix for issue#383: UDPEndPoint not setting REUSEADDR on later LwIP versions

jaylogue:
-- Fixed a bug where the SOF_REUSEADDR flag was not getting set on a UDP PCB when OpenWeave was compiled with with versions of LwIP later than 1.4.

-- Rewrote UDPEndPoint::GetPCB() in an attempt to simplify the logic and eliminate duplicate code paths.


jaylogue:
> What unit / functional test can be added to ensure that there aren't further regressions on this issue in the future?

We could construct a specific test case to verify the REUSEADDR flag.  However I'm working on a larger change relating to UDP sending in Weave which could probably cover this case.

Issue #368, opened 2019-09-13
Support mutli-interface IPv4 broadcast


jaylogue:
_Based on further consideration, I am simplifying the goal of this issue in order to better reflect the functionality that I believe application developers will want._

Currently, if one specifies the IPv4 broadcast address (255.255.255.255) as the destination of a UDP Weave message, the choice of interface over which the packet is set is left up to the underlying OS.  Often, this ends up being the host's default interface.  This is undesirable in certain situations, such as on a mobile phone that is connected to the private WiFi network of a device, but has a default route pointing to its cellular interface.

Proposed new behavior: Add logic to the Weave Message Layer SendMessage() code to recognize an attempt to send to the IPv4 local broadcast address (255.255.255.255) and translate this into a series of UDP packet sends directed at each interface that supports broadcast.  For each UDP packet, the destination address will be the  IPv4 local broadcast address.

_**Previous proposed behavior...**_

_... Add logic to the Weave Message Layer SendMessage() code to recognize an attempt to send to the IPv4 broadcast address and translate this into individual, per-interface sends using the appropriate subnet-specific broadcast addresses for each interface.  The Weave Message Layer will handle the work of enumerating interfaces and determining their broadcast addresses.  This has the added benefit of allowing a single SendMessage() call to broadcast across all compatible interfaces._

gerickson:
Is there a way this could be further optimized by having clients provide a "subnet" list (Thread, WiFi, etc.) with the default being "all"?

jaylogue:
@gerickson 

> Is there a way this could be further optimized by having clients provide a "subnet" list (Thread, WiFi, etc.) with the default being "all"?

The context of this issue is IPv4, so Weave IPv6 addressing isn't relevant.  A typically application of this would be a mobile trying to rendezvous with a device using IPv4.

jaylogue:
Fixed in openweave-core:master in commit 9238c86.

Issue #367, opened 2019-09-13
Better support for multiple Weave client processes on single host


jaylogue:
Modify the Weave Message Layer and associated code to better support multiple Weave-enabled processes operating as independent clients on a single host.  This should include:

- The ability to initiate outbound TCP and UDP communications from an ephemeral port.

- The ability to disable TCP/UDP listening on the standard Weave port.

- Automatic assignment of unique device ids to client instances.

gerickson:
+1

PR #393, opened 2019-10-23
Weave over UDP changes

jaylogue:
This is a series of commits related to the handling of UDP send and receive in the Weave Message and Exchange Layers.  The goals of these changes are:

- to simplify the code, in particular the code in SendMessage() and RefreshEndpoints()
- add support for new UDP send features
- correct bugs related to source address selection, multicast handling and others
- improve testing

The individual commits each have detailed commit messages, however the highlights are reproduced here:

-------------

-- Rewrote the Message Layer’s SendMessage() and RefreshEndpoints() methods to simplify the code and support new features.  As part of this, a number of global tables were eliminated, leading to an overall reduction in RAM use.

-- Added support for ephemeral UDP source ports.  Client applications can use this feature to support multiple simultaneous Weave initiators running in different processes.

-- Enhanced the SendMessage() method such that messages sent to the IPv4 broadcast address (255.255.255.255) are automatically sent to all interfaces.  This follows similar behavior for IPv6 multicast messages.

-- Added support in the Weave Message Layer for an additional listening UDP endpoint bound to the IPv4 broadcast address (255.255.255.255).  This endpoint is initialized whenever the Message Layer is bound to a specific IPv4 address, and allows the node to receive broadcast messages when bound in this way.

-- Altered the logic in RefreshEndpoints() that decides when re-initialize UDP endpoints. Specifically, already listening UDP endpoints are only re-initialized if the message layer receives specific errors when sending UDP messages indicating the underlying socket/PCB has failed.

-- Implemented a set of basic functionality tests for sending and receiving Weave messages over UDP.  Test cases cover:

- IPv4 and IPv6
- Unicast, IPv6 multicast and IPv4 broadcast
- Sending to/from IPv6 ULA and link-local addresses
- Sending over specific network interfaces
- Sending and receiving while bound to specific local addresses (IPv4 and IPv6)


jaylogue:
This PR addresses the issues described in [367](https://github.com/openweave/openweave-core/issues/367) and [368](https://github.com/openweave/openweave-core/issues/368).

src/lib/core/ExchangeContext.cpp
```diff
@@ -524,10 +543,16 @@ WEAVE_ERROR ExchangeContext::SendMessage(uint32_t profileId, uint8_t msgType, Pa
...
     if (sendFlags & kSendFlag_ReuseSourceId)
         msgInfo->Flags |= kWeaveMessageFlag_ReuseSourceId;
-    if (sendFlags & kSendFlag_MulticastFromLinkLocal)
-        msgInfo->Flags |= kWeaveMessageFlag_MulticastFromLinkLocal;
+    if (sendFlags & kSendFlag_DefaultMulticastSourceAddress)
+        msgInfo->Flags |= kWeaveMessageFlag_DefaultMulticastSourceAddress;
+
+    SetFlag(msgInfo->Flags, kWeaveMessageFlag_FromInitiator, IsInitiator());
+
+#if WEAVE_CONFIG_ENABLE_EPHEMERAL_UDP_PORT
```
gerickson: I would have expected this conditional wrapper in more places, for example lines 330 and possibly in UDPEndpoint.cpp. Any reason it isn't used in those locations?

src/device-manager/WeaveDeviceManager.cpp
```diff
@@ -422,25 +422,23 @@ WEAVE_ERROR WeaveDeviceManager::InitiateDeviceEnumeration()
...
     //
-    // Send the multicast identify request from the host's link-local addresses, rather than
-    // from its site-local or global addresses. This results in the device responding using its
-    // link-local address, which in turn causes the device manager to connect to the device using the
-    // link-local address. This is all to work around a bug in OS X/iOS that prevents those systems
-    // from communicating on any site-local IPv6 subnets when in the presence of a router that is
-    // advertising a default route to the Internet at large. (See DOLO-2479).
+    // However, if the application has enabled the 'rendezvous link-local' option, force the use of the default
+    // source address selection by passing the DefaultMulticastSourceAddress flag to SendMessage().  For typical
+    // multicast rendezvous addresses (e.g. FF02::1 all-nodes, link-local), this will resort in the source address
```
gerickson: resort -> result?

src/lib/core/ExchangeContext.cpp
jaylogue: I went back and forth on this.  I agree that SetUseEphemeralUDPPort() should be conditionalized and will make that change.  However I'm inclined to leave UseEphemeralUDPPort() as an inline that always returns false.

Regarding UDPEndPoint, ephemeral ports are a feature of the IP stack and thus have always been supported in UDPEndPoint.  The only related change to UDPEndPoint is the new GetBoundPort() method which is used (and useful) even when support for ephemeral UDP ports is turned off in Weave.

src/device-manager/WeaveDeviceManager.cpp
```diff
@@ -2751,25 +2749,23 @@ WEAVE_ERROR WeaveDeviceManager::InitiateConnection()
 
     mConState = kConnectionState_IdentifyDevice;
 
-    // Send the Identify message.
+    // Send an Identify message over UDP to the specified rendezvous address.  Typically the rendezvous address
+    // will be an multicast/broadcast address, however this can be changed by the application.
```
gerickson: an -> a

src/device-manager/WeaveDeviceManager.cpp
```diff
@@ -2751,25 +2749,23 @@ WEAVE_ERROR WeaveDeviceManager::InitiateConnection()
...
     //
-    // Send the multicast identify request from the host's link-local addresses, rather than
-    // from its site-local or global addresses. This results in the device responding using its
-    // link-local address, which in turn causes the device manager to connect to the device using the
-    // link-local address. This is all to work around a bug in OS X/iOS that prevents those systems
-    // from communicating on any site-local IPv6 subnets when in the presence of a router that is
-    // advertising a default route to the Internet at large. (See DOLO-2479).
+    // However, if the application has enabled the 'rendezvous link-local' option, force the use of the default
+    // source address selection by passing the DefaultMulticastSourceAddress flag to SendMessage().  For typical
+    // multicast rendezvous addresses (e.g. FF02::1 all-nodes, link-local), this will resort in the source address
```
gerickson: resort -> result?

src/lib/core/WeaveMessageLayer.cpp
```diff
@@ -497,16 +547,27 @@ bool WeaveMessageLayer::IsSendErrorNonCritical (WEAVE_ERROR err)
  *  @retval  errors generated from the lower Inet layer UDP endpoint during sending.
  *
  */
-WEAVE_ERROR WeaveMessageLayer::SendMessage(const IPAddress &destAddr, uint16_t destPort, InterfaceId sendIntfId,
-                                           PacketBuffer *payload, uint16_t msgSendFlags)
+WEAVE_ERROR WeaveMessageLayer::SendMessage(const IPAddress & destAddr, uint16_t destPort, InterfaceId sendIntfId,
```
gerickson: This is getting to be a monster of a method call. Any benefit, beyond complexity management such as code size, to refactoring it into a singular public method with supporting private methods?

jaylogue: My impression is that the new method is of comparable complexity to the old one, but easier to follow due to the clear phasing of the procedure.  Nonetheless, it seems reasonable to factor out some of the logic, such as the code that selects the sending end point.  

I am loath, however, to expose the sendAction enum outside of this method, as it exists solely to make the sending logic easier to understand.  Thus I think I'll keep that part of the code within SendMessage().

Issue #357, opened 2019-09-06
Python TLVWriter init bug that prevents re-use


wijnandatgoogle:
TLVWriter has in init like so:
def __init__(self, encoding=bytearray(), implicitProfile=None):
That does not lead or re-init of encoding, as you might expect. Instead, encoding should be defaulting to None, and a run-time check for None can then do:
if encoding == None:
   encoding = bytearray()

jaylogue:
Fixed in openweave-core:master in commit 37b3c15.

PR #394, opened 2019-10-23
Fix for bug in Python TLVWriter

jaylogue:
-- Fixed a bug in the Python TLVWriter implementation caused all instances of the class to use the same output buffer.


jaylogue:
This change addresses issue [357](https://github.com/openweave/openweave-core/issues/357).

PR #395, opened 2019-10-23
Fix re-joining to the disconnected Thread network

robszewczyk:
When the OpenThread interface is disconnected, the Thread stack
removes the addresses on that interface.  By the time WARM has a
chance to deal with the situation, the Weave ULA address is no longer
on the interface, and the LwIP call
`netif_remove_ip6_address_with_route` returns an error code.  In turn,
when WARM observes that the operation to remove the address fails, it
does not update the state accordingly.  When the device rejoins, WARM
(having failed to update the state), does not reassign the Weave ULA
address to the interface.  As a result, all the packets going out to
Weave addresses (and the service address in particular) are sent with
the Thread mesh-local address as the source address and thus fail to
be routed properly.  By squashing the error on remove, we enable the
WARM state machine to re-add the Weave ULA address on reattach, and
thus put all routing decisions back on their well-worn paths.

### DIFF
--- a/src/adaptations/device-layer/LwIP/WarmSupport.cpp
+++ b/src/adaptations/device-layer/LwIP/WarmSupport.cpp
@@ -132,6 +132,15 @@ PlatformResult AddRemoveHostAddress(InterfaceType inInterfaceType, const Inet::I
     else
     {
         lwipErr = netif_remove_ip6_address_with_route(netif, &ip6addr, inPrefixLength);
+        // There are two possible errors from netif_remove_ip6_address: ERR_ARG
+        // if call was made with wrong arguments, or ERR_VAL if the action could
+        // not be performed (e.g. the address was already removed).  We squash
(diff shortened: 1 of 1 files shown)


jaylogue:
@robszewczyk : Interesting find.  Can you elaborate on how the ULA gets removed from the netif in the first place?

robszewczyk:
@jaylogue I have not traced the mechanism that OpenThread triggers. I do note however that:

1. we get `ERR_VAL` because `netif_get_ip6_addr_match` returns an error.
2. adding the same address multiple times is supported in LwIP -- if an address already exists, LwIP will not duplicate the address.

robszewczyk:
@jaylogue following up on the in-person chat yesterday: it looks like the addresses are cleared in `GenericThreadStackManagerImpl_OpenThread_LwIP.ipp`  in `void GenericThreadStackManagerImpl_OpenThread_LwIP<ImplClass>::UpdateThreadInterface(bool addrChange)` around line 242

PR #402, opened 2019-10-25
NoMorePendingEventCbHelper needs to be called when flush empty updates

yunhanw-google:
--When application calls FlushUpdate, if no updated item is set,
mPendingSetState is kPendingSetEmpty, NoMorePendingEventCbHelper()
needs to be called, and notify application there is no more pending
updates.

### DIFF
--- a/src/lib/profiles/data-management/Current/SubscriptionClient.cpp
+++ b/src/lib/profiles/data-management/Current/SubscriptionClient.cpp
@@ -3249,6 +3249,11 @@ WEAVE_ERROR SubscriptionClient::FlushUpdate(bool aForce)
 exit:
     UnlockUpdateMutex();
 
+    if (mPendingSetState == kPendingSetEmpty)
+    {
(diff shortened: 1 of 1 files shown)


yunhanw-google:
> This logic isn't correct. FlushUpdate() is designed to be idempotent.
Thanks for the comments, yes, FlushUpdate should be idempotent here. I think, when application calls FlushUpdate, if no updated item is set, mPendingSetState is kPendingSetEmpty, NoMorePendingEventCbHelper() needs to be called, and notify application there is no more pending updates. I have updated the pull request, please help review. thank you!

yunhanw-google:
Should we notify app the error state via OnComplete or OnError callback when IsUpdateInFlight  or mUpdateRetryScheduled or mUpdateFlushScheduled is true?
I feel multiple call on FlushUpdate should be fine for the above cases.

yunhanw-google:
Done, thanks

PR #405, opened 2019-10-31
fix pending update state clean when discarding Update

yunhanw-google:
--When application trigger DiscardUpdate with WEAVE_NO_ERROR, pending
set state needs to be reset.

### DIFF
--- a/src/lib/profiles/data-management/Current/SubscriptionClient.cpp
+++ b/src/lib/profiles/data-management/Current/SubscriptionClient.cpp
@@ -2913,6 +2913,7 @@ void SubscriptionClient::AbortUpdates(WEAVE_ERROR aErr)
     {
         numPending = mPendingUpdateSet.GetNumItems();
         mPendingUpdateSet.Clear();
+        SetPendingSetState(kPendingSetEmpty);
 
         numInProgress = mInProgressUpdateList.GetNumItems();
         mInProgressUpdateList.Clear();


src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -2916,6 +2916,7 @@ void SubscriptionClient::AbortUpdates(WEAVE_ERROR aErr)
 
         numInProgress = mInProgressUpdateList.GetNumItems();
         mInProgressUpdateList.Clear();
+        SetPendingSetState(kPendingSetEmpty);
```
robszewczyk: Nit: add a newline above to emphasize this is a separate action from cleaning `mInProgressUpdateList`

jaylogue: I would prefer this line to go next to the mPendingUpdateSet.Clear() line above, since the state of the two variables are tightly coupled.

yunhanw-google: Done, thanks

yunhanw-google:
> Nit: minor changes to commit message:
> 
> * whitespace after the two leading dashes, and before `When`
> * replace `trigger` with `triggers`
> 
> the result should read
> 
> ```
> -- When application triggers `DiscardUpdate` with `WEAVE_NO_ERROR`, reset pending set state.
> ```

Done, thanks

PR #411, opened 2019-11-05
Support for coordinated advertising between WoBLE and ToBLE

jaylogue:
On platforms where there is no native support for simultaneous BLE advertising (e.g. Nordic’s SoftDevice), it is necessary to coordinate between the different advertising modes a Weave device may employ.  This arises in particular when a device supports both WoBLE and ToBLE, each of which requires a separate advertising regime.  The following changes make it possible for a device to switch between advertising modes, at the appropriate times, such that it can undergo Weave pairing via WoBLE while simultaneously joining a Thread network using ToBLE.

-- Added WEAVE_DEVICE_CONFIG_WOBLE_DISABLE_ADVERTISING_WHEN_PROVISIONED feature which instructs the OpenWeave Device Layer to disable WoBLE advertising when the device transitions to, or boots in, a fully provisioned state.

-- Added ThreadStackManager::OnWoBLEAdvertisingStart() and OnWoBLEAdvertisingStop() which are called by the BLEManager whenever WoBLE advertising starts/stops.

-- Added implementations of OnWoBLEAdvertisingStart() and OnWoBLEAdvertisingStop() to the nRF52840 version of ThreadStackManagerImpl to disable and re-enable ToBLE advertising when WoBLE advertising is enabled/disabled.

-- Added ConfigurationManager::IsFullyProvisioned() method which assesses whether all major device provisioning steps have been completed.


build/esp32/components/openweave/Kconfig
```diff
@@ -582,7 +582,24 @@ menu "OpenWeave Device Layer"
...
+                Limit support for Weave-over-BLE (WoBLE) to a single inbound connection.
+                
+                When set, WoBLE advertisements will stop while a WoBLE connection is active.
+
+        config WOBLE_DISABLE_ADVERTISING_WHEN_PROVISIONED
+            bool "Disable Advertising when Provisioned"
+            default n
+            depends on ENABLE_WOBLE
+            help
+                Automatically disable WoBLE advertising when the device transitions to a fully
```
gerickson: Minor: recommend being consistent about WoBLE terminology by including both expanded an parenthetical form as with line 591 above: "Weave-over-BLE (WoBLE)".

src/adaptations/device-layer/nRF5/BLEManagerImpl.cpp
```diff
@@ -278,6 +278,21 @@ void BLEManagerImpl::_OnPlatformEvent(const WeaveDeviceEvent * event)
         WeaveLogError(DeviceLayer, "BLEManagerImpl: Out of buffers during WoBLE RX");
         break;
 
+    // If WOBLE_DISABLE_ADVERTISING_WHEN_PROVISIONED is enabled, and there is a change to the
```
gerickson: This looks the same, if not nearly the same as for ESP32. Is there an argument for having a common, base implementation?

jaylogue: Eventually, yes.  It requires some thinking about how to best factor out the common logic, which I haven't had time to do yet.

Issue #410, opened 2019-11-04
SoftwareUpdateManager does not close ImageQuery exchange


jaylogue:
The code in GenericSoftwareUpdateManagerImpl<> does not properly close the exchange context used for the SWU:ImageQuery request.  The exchange context is technically not leaked, as it gets released (via Abort()) the next time a software update check happens.  But context remains allocated for all the intervening time between update checks, adding to memory pressure.

One consequence of the lack of a call to ExchangeContext::Close() is that the stand-alone ACK that should be sent after reception of the ImageQueryResponse gets delayed, or in some cases not sent at all.  This can lead to unnecessary re-transmissions of the ImageQueryResponse from the service.

jaylogue:
Fixed.

Issue #409, opened 2019-11-04
SoftwareUpdateManager: Support asynchronous ClearImageFromStorage event


jaylogue:
The interface to the SoftwareUpdateManager currently defines the ClearImageFromStorage API event as being synchronous, implying that the application must complete the clear action before the event callback function returns.  On many platforms, however, the act of clearing flash can take a long time--many seconds in some cases.  During this time, the Weave task is blocked from processing other events.  This leads to timeouts and/or excessive resource consumption as pending I/O from other tasks builds up in the Weave queue.

To resolve this, the existing ClearImageFromStorage event should be replaced with an asynchronous version (e.g. StartClearImageFromStorage), and a corresponding method should be added to SoftwareUpdateManager to signal when the clear operation is complete (e.g. ClearImageFromStorageComplete()).

jaylogue:
Fixed.

PR #415, opened 2019-11-07
SoftwareUpdateManager: Support asynchronous ClearImageFromStorage event

jaylogue:
This addresses issues https://github.com/openweave/openweave-core/issues/409 and https://github.com/openweave/openweave-core/issues/410.

-- Restructured the interface to SoftwareUpdateManager such that the work of clearing local storage in preparation for a firmware download can be done asynchronously while the Weave task continues to run and handle other interactions.  To support this, the ClearImageFromStorage API event was replaced with a new PrepareImageStorage event.  The new event is asynchronous in nature, requiring the application to call PrepareImageStorageComplete() when the work is done.

As part of this change, an additional API event was added (ResetPartialImageInfo) to handle the case where a previously downloaded image should be forgotten because it failed an integrity check.  This behavior was previously overloaded onto the ClearImageFromStorage event.

-- Modified GenericSoftwareUpdateManagerImpl<> to close the ImageQuery exchange context immediately upon receipt of a response.  When running over WRM, this ensures that the standalone ACK is flushed right away, reducing overall buffer pressure once the BDX download starts.

-- In a similar vein, GenericSoftwareUpdateManagerImpl<> was modified to release the buffer containing the ImageQueryResponse as soon as its contents have been processed.

-- Reduced the default number of BDXTransfer objects on nRF52840 platforms from 12 to 1.

-- Enhanced the documentation for the SoftwareUpdateManager interface.


src/adaptations/device-layer/include/Weave/DeviceLayer/SoftwareUpdateManager.h
```diff
@@ -52,109 +52,185 @@ class SoftwareUpdateManager
...
+        kState_Install              = 7,
 
-        kState_ApplicationManaged   = 7,
+        kState_ApplicationManaged   = 8,
 
         kState_MaxState
     };
 
+    /**
+     * API events generated by the \c SoftwareUpdateManager object.
```
suryanshup: \c  ?

jaylogue: doxygen syntax to display following word in "typewriter" font.

Issue #416, opened 2019-11-07
BleEndPointPool elements can be misaligned when compiler optimizations are active


jasongraffius:
When space optimization is turned on, [`BLEEndPoint`](https://github.com/openweave/openweave-core/blob/master/src/ble/BleLayer.cpp#L103) elements can end up not aligned correctly in the [`BleEndPointPool`](https://github.com/openweave/openweave-core/blob/master/src/ble/BleLayer.cpp#L98). Alignment of both [`PoolElement`](https://github.com/openweave/openweave-core/blob/master/src/ble/BleLayer.cpp#L109)s and [`sEndPointPool`](https://github.com/openweave/openweave-core/blob/master/src/ble/BleLayer.cpp#L111) should be forced into 8-byte alignment.

jaylogue:
Fixed.

PR #417, opened 2019-11-07
(Issue #416) Ensure 8B alignment of BLEEndpoints

jasongraffius:
Force alignment of both PoolElement and sEndPointPool to ensure that returned BLEEndPoints are 8-byte aligned

### DIFF
--- a/src/ble/BleLayer.cpp
+++ b/src/ble/BleLayer.cpp
@@ -90,9 +90,6 @@
 #define CAPABILITIES_MSG_CHECK_BYTE_1 'n'
 #define CAPABILITIES_MSG_CHECK_BYTE_2 'l'
 
-#define BLE_ENDPOINT_ALIGNMENT_TYPE uint64_t
-#define BLE_ENDPOINT_ALIGNMENT sizeof(BLE_ENDPOINT_ALIGNMENT_TYPE)
-
 // clang-format on
 
 namespace nl {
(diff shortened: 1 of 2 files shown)


src/ble/BleLayer.cpp
```diff
@@ -104,15 +107,24 @@ class BleEndPointPool
...
+            uint8_t Data[((sizeof(BLEEndPoint) + BLE_ENDPOINT_ALIGNMENT - 1) / BLE_ENDPOINT_ALIGNMENT) * BLE_ENDPOINT_ALIGNMENT];
+            BLE_ENDPOINT_ALIGNMENT_TYPE ForceAlignment;
         } PoolElement;
 
-        static uint8_t sEndPointPool[sizeof(PoolElement) * BLE_LAYER_NUM_BLE_ENDPOINTS];
+        static union
+        {
+            uint8_t Pool[sizeof(PoolElement) * BLE_LAYER_NUM_BLE_ENDPOINTS];
+            BLE_ENDPOINT_ALIGNMENT_TYPE ForceAlignment;
+        } sEndPointPool;
```
jaylogue: This is unnecessarily complex.   First off, sizeof() a type always includes the padding necessary to ensure that every element in an array of such types is aligned, provided that the array itself is aligned.  So there's no need to round the sizeof an element up to the sizeof its alignment type.

Additionally (and really for the same reason) one only needs to force the alignment of the entire array, not each element.

Thus the above unions can be simplified to just:

```
static union
{
    uint8_t Pool[sizeof(BLEEndPoint) * BLE_LAYER_NUM_BLE_ENDPOINTS];
    BLE_ENDPOINT_ALIGNMENT_TYPE ForceAlignment;
} sEndPointPool;
```

Finally, don't use #defines for types, use typedefs.  In this case, the type being defined is intimately related to the definition of BLEEndPoint.  So it is customary to declare the alignment type as a public member of the to-be-aligned class.  E.g.:

```
class NL_DLL_EXPORT BLEEndPoint : public BleLayerObject
{
public:
    typedef uint64_t AlignT;
    ...
}
```

which would reduce the pool definition to:

```
static union
{
    uint8_t Pool[sizeof(BLEEndPoint) * BLE_LAYER_NUM_BLE_ENDPOINTS];
    BLEEndPoint::AlignT ForceAlignment;
} sEndPointPool;
```

src/ble/BleLayer.cpp
```diff
@@ -104,15 +107,24 @@ class BleEndPointPool
...
+
+        // Assert sEndPointPool is aligned
+        nlASSERT(((uintptr_t) &sEndPointPool % BLE_ENDPOINT_ALIGNMENT) == 0);
+        // Assert sEndPointPool elements are aligned
+        nlASSERT(((uintptr_t) (sEndPointPool.Pool + sizeof(PoolElement)) % BLE_ENDPOINT_ALIGNMENT) == 0);
```
jaylogue: I would prefer that we not incur a runtime overhead for checking the alignment on every Get() call.  It isn't worth it.

src/ble/BleLayer.cpp
jasongraffius: The first block of code (reproduced below to be specific) was my first approach exactly, but only the first element was aligned, the second element and beyond were not aligned when the compiler optimizations were included. It works fine without the optimizations, though.

```c++
static union
{
    uint8_t Pool[sizeof(BLEEndPoint) * BLE_LAYER_NUM_BLE_ENDPOINTS];
    BLE_ENDPOINT_ALIGNMENT_TYPE ForceAlignment;
} sEndPointPool;
```

Good point on the typedef and keeping it in the class, not sure what I was thinking with that.

src/ble/BleLayer.cpp
jasongraffius: Good point, I originally used static asserts, but there's no point of runtime asserts here.

jaylogue:
Looks good!

PR #418, opened 2019-11-10
add ScheduleRun in NE

yunhanw-google:
--DataManagement Update Server needs to send notify to client when
  subscription is there. In order to decrease the stack depth, add
  ScheduleRun in NotificationEngine to achieve async notificaiton
  engine run.

### DIFF
--- a/src/lib/profiles/data-management/Current/NotificationEngine.cpp
+++ b/src/lib/profiles/data-management/Current/NotificationEngine.cpp
@@ -1,6 +1,7 @@
 /*
  *
- *    Copyright (c) 2016-2017 Nest Labs, Inc.
+ *    Copyright (c) 2016-2018 Nest Labs, Inc.
+ *    Copyright (c) 2019 Google, LLC.
  *    All rights reserved.
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


src/lib/profiles/data-management/Current/NotificationEngine.cpp
```diff
@@ -1701,6 +1701,18 @@ WEAVE_ERROR NotificationEngine::BuildSubscriptionlessNotification(PacketBuffer *
 
 #endif // WDM_ENABLE_SUBSCRIPTIONLESS_NOTIFICATION
 
+void NotificationEngine::OnNotifyScheduleWorkCallback(System::Layer * aSystemLayer, void * aAppState, System::Error)
```
jaylogue: Name is confusing.  I would call it `Run(System::Layer * aSystemLayer, void * aAppState, System::Error)`.

src/lib/profiles/data-management/Current/NotificationEngine.cpp
```diff
@@ -1701,6 +1701,18 @@ WEAVE_ERROR NotificationEngine::BuildSubscriptionlessNotification(PacketBuffer *
...
+{
+    NotificationEngine * const pEngine = reinterpret_cast<NotificationEngine *>(aAppState);
+    WeaveLogDetail(DataManagement, "%s", __func__);
```
jaylogue: No need for additional logging here.  It doesn't add anything valuable.

src/lib/profiles/data-management/Current/NotificationEngine.cpp
yunhanw-google: Done, thanks

src/lib/profiles/data-management/Current/NotificationEngine.cpp
yunhanw-google: Done, thanks

PR #424, opened 2019-11-14
Nordic BLEManager changes

jaylogue:
- Introduced a new WoBLEAdvertisingChange device event that gets generated whenever the BLEManager starts and stops WoBLE advertising.

- Fixed logic bugs in the transition between fast and slow WoBLE advertising.  The system will now properly transition between the two modes when the device’s provisioning state changes, or when expressly asked to by the application.

- Eliminated subtle race condition where WoBLE would be disabled entirely due to an unexpected error from the SoftDevice if a connection arrived just as the BLEManager was updating the advertising configuration.

- Renamed BLEManager internal flag values and added documentation to make their meaning clearer.


jaylogue:
Going to re-submit this with a version that includes changes for the ESP32.

PR #426, opened 2019-11-15
Compile inlined IsMsgCounterSyncReqInProgress iff feature flag is ena…

robszewczyk:
…bled


robszewczyk:
> What's the underlying issue here?

The underlying issue is that the member method is declared inside a conditionally compiled block; C++ compiler will complain if the code provides an implementation for an undeclared method.

robszewczyk:
> > What's the underlying issue here?
> 
> The underlying issue is that the member method is declared inside a conditionally compiled block; C++ compiler will complain if the code provides an implementation for an undeclared method.

It may have been even better to put the inlined body right at the point of declaration, but that would have impacted the layout of the declarations, so I've opted to add the conditional block here.

PR #429, opened 2019-11-15
Accessor functions for SoftDevice advertising handle

jaylogue:
- Added accessor functions to the Nordic version of the BLEManagerImpl class that allow the application to get and set the SoftDevice BLE advertising handle.

### DIFF
--- a/src/adaptations/device-layer/include/Weave/DeviceLayer/nRF5/BLEManagerImpl.h
+++ b/src/adaptations/device-layer/include/Weave/DeviceLayer/nRF5/BLEManagerImpl.h
@@ -48,6 +48,15 @@ class BLEManagerImpl final
     // the implementation methods provided by this class.
     friend BLEManager;
 
+public:
+
+    // ===== Platform-specific members available for use by the application.
+
+    uint8_t GetAdvertisingHandle(void);
+    void SetAdvertisingHandle(uint8_t handle);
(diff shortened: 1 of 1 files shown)


src/adaptations/device-layer/include/Weave/DeviceLayer/nRF5/BLEManagerImpl.h
```diff
@@ -155,6 +164,16 @@ inline BLEManagerImpl & BLEMgrImpl(void)
     return BLEManagerImpl::sInstance;
 }
 
+inline uint8_t BLEManagerImpl::GetAdvertisingHandle(void)
+{
+    return mAdvHandle;
```
gerickson: Is the handle always initialized to something sane? Is there any way for either the caller or subsequent consumers of the value to know whether the value is sane?

jaylogue: It gets initialized here: https://github.com/openweave/openweave-core/blob/5c52bb5b4fe7f3a53e17952af0be4dfe62bf21bb/src/adaptations/device-layer/nRF5/BLEManagerImpl.cpp#L92

This presumes that the Weave stack has been initialized before any call to GetAdvertisingHandle().

PR #436, opened 2019-12-05
Fixes to the device layer/nRF implementation

robszewczyk:
* Correctly dispatch at compile time on the NRF_SDH_DISPATCH_MODEL
  rather than on the enum that defines the model.  This change enables
  the softdevice to actually use the right queue posting primitives.

* When NRF_DEBUG is enabled, softdevice will dispatch log statements
  from interrupt context.  In order to support this, if appropriate,
  log timestamping needs to be available; in turn, this implies that
  TicksSinceBoot must be callable from interrupt context.  Since
  FreeRTOS does not provide generalized facility to determine whether
  a function is in an interrupt context, we make the symbol weak here,
  and leave it to more specific adaptations to provide a more robust
  implementation.

### DIFF
--- a/src/adaptations/device-layer/FreeRTOS/SystemTimeSupport.cpp
+++ b/src/adaptations/device-layer/FreeRTOS/SystemTimeSupport.cpp
@@ -40,24 +40,64 @@ constexpr uint32_t kTicksOverflowShift = (configUSE_16_BIT_TICKS) ? 16 : 32;
 
 uint64_t sBootTimeUS = 0;
 
-uint64_t TicksSinceBoot(void)
+#ifdef __CORTEX_M
+BaseType_t sNumOfOverflows;
+#endif
+} // unnamed namespace
+
+/**
+ * Returns the number of FreeRTOS ticks since the system booted.
+ *
+ * NOTE: The default implementation of this function uses FreeRTOS's
+ * vTaskSetTimeOutState() function to get the total number of ticks,
+ * irrespective of tick counter overflows.  Unfortunately, this function cannot
+ * be called in interrupt context, no equivalent ISR function exists, and
+ * FreeRTOS provides no portable way of determining whether a function is being
+ * called in an interrupt context.  Adaptations that need to use the Weave
+ * Get/SetClock methods from within an interrupt handler must override this
+ * function with a suitable alternative that works on the target platform.  The
+ * provided version is safe to call on ARM Cortex platforms with CMSIS
+ * libraries.
+ */
+
+uint64_t FreeRTOSTicksSinceBoot(void) __attribute__((weak));
+
+uint64_t FreeRTOSTicksSinceBoot(void)
 {
     TimeOut_t timeOut;
+
+#ifdef __CORTEX_M
+    if (SCB->ICSR & SCB_ICSR_VECTACTIVE_Msk) // running in an interrupt context
+    {
+        // Note that sNumOverflows may be quite stale, and under those
+        // circumstances, the function may violate monotonicity guarantees
+        timeout.xTimeOnEntering = xTaskGetTickCountFromISR();
+        timeout.xOverflowCount = sNumOfOverflows;
+    }
+    else
+    {
+#endif
+
     vTaskSetTimeOutState(&timeOut);
+
+#ifdef __CORTEX_M
+    // BaseType_t is supposed to be atomic
+    sNumOfOverflows = timeOut.xOverflowCount;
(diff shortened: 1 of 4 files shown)


src/adaptations/device-layer/nRF5/BLEManagerImpl.cpp
```diff
@@ -802,7 +802,7 @@ void BLEManagerImpl::SoftDeviceBLEEventCallback(const ble_evt_t * bleEvent, void
     }
 
     // Post the event to the Weave queue.
-#if NRF_SDH_DISPATCH_MODEL_INTERRUPT
+#if (NRF_SDH_DISPATCH_MODEL == NRF_SDH_DISPATCH_MODEL_INTERRUPT)
```
jaylogue: Doh! :)

src/adaptations/device-layer/FreeRTOS/SystemTimeSupport.cpp
```diff
@@ -40,7 +40,7 @@ constexpr uint32_t kTicksOverflowShift = (configUSE_16_BIT_TICKS) ? 16 : 32;
 
 uint64_t sBootTimeUS = 0;
 
-uint64_t TicksSinceBoot(void)
+uint64_t TicksSinceBoot(void) __attribute__((weak))
```
jaylogue: This isn't useful as-is because TicksSinceBoot() is declared in an anonymous namespace, so it can't be overridden. 

I suggest moving it into the Platform::Layer namespace and then renaming it something more specific, like FreeRTOSTicksSinceBoot().

Also I suggest adding documentation explaining the issue.  E.g.:

```
/**
 * Returns the number of FreeRTOS ticks since the system booted.
 *
 * NOTE: The default implementation of this function uses FreeRTOS's vTaskSetTimeOutState()
 * function to get the total number of ticks, irrespective of tick counter overflows.
 * Unfortunately, this function cannot be called in interrupt context, and no equivalent
 * ISR function exists. Therefore, adaptations that need to use the Weave Get/SetClock
 * methods from within an interrupt handler must override this function with a suitable
 * alternative that works on the target platform.
 */
uint64_t FreeRTOSTicksSinceBoot(void) __attribute__((weak))
```

robszewczyk: SGTM, will do.

src/adaptations/device-layer/FreeRTOS/SystemTimeSupport.cpp
```diff
@@ -40,24 +40,59 @@ constexpr uint32_t kTicksOverflowShift = (configUSE_16_BIT_TICKS) ? 16 : 32;
...
 {
     TimeOut_t timeOut;
+
+#ifdef __CORTEX_M
+    if (SCB->ICSR & SCB_ICSR_VECTACTIVE_Msk) // running in an interrupt context
+    {
+        // Note that sNumOverflows may be quite stale, and under those
+        // circumstances, the function may violate monotonicity guarantees
+        return static_cast<uint64_t>(xTaskGetTickCountFromISR()) +
+                static_cast<uint64_t>(sNumOfOverflows << kTicksOverflowShift);
```
jaylogue: Not a big deal, but you could collapse this with the return statement below by simply setting the fields within timeOut here.

robszewczyk: True.  I like this idea, it will clean it up

src/lwip/freertos/sys_arch.c
```diff
@@ -274,12 +274,11 @@ u32_t sys_now(void)
 
 sys_prot_t sys_arch_protect(void)
 {
-    taskENTER_CRITICAL();
-    return 1;
+    return taskENTER_CRITICAL_FROM_ISR();
```
jaylogue: Ah, so this is callable even outside of ISR context?

robszewczyk: Yup.  On NRF52, both variants of `ENTER_CRITICAL` do the same thing -- they both ultimately call `vPortRaiseBASEPRI()`.  The difference is that on exit, the task-variant restores the interrupts to 0, whereas the ISR variant restores them to a saved value.  There is a more significant difference on the nRF51 series --  in ISR variant, the code disables interrupts, in task variant with softdevice present, the code calls `sd_nvic_critical_region_enter` which presumably permits some sd-related ISRs to proceed unmolested.

PR #439, opened 2019-12-10
Add RSA signature support.

emargolis:
Defined types, objects, and methods for working with RSA
public and private keys and RSA signatures.

### DIFF
--- a/src/lib/support/crypto/RSA.cpp
+++ b/src/lib/support/crypto/RSA.cpp
@@ -1,0 +1,244 @@
+/*
+ *
+ *    Copyright (c) 2019 Google LLC.
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


src/lib/support/crypto/RSA.cpp
```diff
@@ -0,0 +1,247 @@
...
+ * with the given tag.
+ *
+ * @param[in] reader            The TLVReader object from which the encoded signature should
+ *                              be read.
+ * @param[in] tag               TLV tag expected to be associated with the encoded signature structure.
+ *
+ * @retval #WEAVE_NO_ERROR      If the operation succeeded.
+ * @retval other                Other Weave error codes related to signature reading.
+ */
+WEAVE_ERROR EncodedRSASignature::ReadSignature(TLVReader& reader, uint64_t tag)
```
jaylogue: Passing the tag just to check it seems unnecessary.  Is this following a pattern used elsewhere?

emargolis: Agree that it is unnecessary. Updated.

PR #442, opened 2019-12-10
Add experimental DataManagement Client support with C++ and Python

yunhanw-google:
-- add experimental data management support c++ API in mobile client side includng genericdatasink and wdmclient.
-- add experimental python API for datamanagement in client side.


src/device-manager/python/openweave/WeaveTLV.py
```diff
@@ -0,0 +1,619 @@
...
+
+            decoding["value"] = val
+            self._bytesRead += decoding["strDataLen"]
+        else:
+            raise ValueError('Attempt to decode unsupported TLV type')
+
+    def _get(self, tlv, decodings, out):
+        endOfEncoding = False
+
+        while len(tlv[self._bytesRead:]) > 0 and endOfEncoding == False:
```
wijnandatgoogle: For python3, this fails. Slicing only works on strings. This generates 'TypeError: 'int' object is not subscriptable'
I think casting tlv to str will work. len(str(tlv)[x:])

yunhanw-google: if it is integer, it should be the error code returned from data management client, i have added check, we can sync offline on it, thanks

yunhanw-google:
@jaylogue @robszewczyk all comments addressed! thanks

src/device-manager/python/openweave/WeaveStack.py
```diff
@@ -0,0 +1,225 @@
...
+class WeaveStackError(WeaveStackException):
+    def __init__(self, err, msg=None):
+        self.err = err
+        if msg != None:
+            self.msg = msg
+        else:
+            self.msg = "Weave Stack Error %ld" % (err)
+
+    def __str__(self):
+        return self.msg
```
wijnandatgoogle: I think it would be good to cast to str explicitly: return str(self.msg). I ran into cases self.msg was a bytearray

yunhanw-google: Please help reproduce it and share the log, I would fix it in follow-up PR, thanks

PR #443, opened 2019-12-10
Add Java API for DataManagement in client side

yunhanw-google:
--add JAVA API for dataManagement in client side

### DIFF
--- a/src/device-manager/java/Makefile.am
+++ b/src/device-manager/java/Makefile.am
@@ -103,6 +103,13 @@ WeaveDeviceManager_jar_JAVA_SRCS                   =        \
     nl/Weave/DeviceManager/WeaveDeviceException.java        \
     nl/Weave/DeviceManager/DeviceFeatures.java              \
     nl/Weave/DeviceManager/NevisPairingCodeDecoding.java    \
+    nl/Weave/DeviceManager/WeaveDeviceManager.java          \
+    nl/Weave/DataManagement/WdmClient.java                  \
+    nl/Weave/DataManagement/WdmClientImpl.java              \
+    nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java \
+    nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java \
+    nl/Weave/DataManagement/WdmClientFactory.java           \
+    nl/Weave/DataManagement/ResourceIdentifier.java         \
     $(NULL)
 
 WeaveDeviceManager_jar_JFLAGS                      = -source 6 -target 6
--- a/src/device-manager/java/WeaveDeviceManager-JNI.cpp
+++ b/src/device-manager/java/WeaveDeviceManager-JNI.cpp
@@ -1,7 +1,7 @@
 /*
  *
- *    Copyright (c) 2019 Google LLC
- *    Copyright (c) 2013-2017 Nest Labs, Inc.
+ *    Copyright (c) 2019-2020 Google LLC
+ *    Copyright (c) 2013-2018 Nest Labs, Inc.
  *    All rights reserved.
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
@@ -28,6 +28,7 @@
 #include <stdlib.h>
 #include <stdint.h>
 #include <pthread.h>
+#include <string>
 #include <jni.h>
 
 #include <SystemLayer/SystemLayer.h>
@@ -39,11 +40,16 @@
 #include <Weave/Profiles/common/CommonProfile.h>
 #include <Weave/Profiles/network-provisioning/NetworkProvisioning.h>
 #include <Weave/Profiles/device-description/DeviceDescription.h>
+#include <Weave/Profiles/data-management/DataManagement.h>
+#include <Weave/Profiles/data-management/Current/WdmManagedNamespace.h>
 #include <Weave/Support/logging/WeaveLogging.h>
 #include <Weave/Support/ErrorStr.h>
 #include <Weave/Support/NLDLLUtil.h>
 #include <Weave/Support/platform/PersistedStorage.h>
 #include <Weave/DeviceManager/WeaveDeviceManager.h>
+#if WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
+#include <Weave/DeviceManager/WeaveDataManagementClient.h>
+#endif // WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
 #include "AndroidBleApplicationDelegate.h"
 #include "AndroidBlePlatformDelegate.h"
 
@@ -59,6 +65,7 @@ using namespace nl::Weave;
 using namespace nl::Weave::DeviceManager;
 using namespace nl::Weave::Profiles::NetworkProvisioning;
 using namespace nl::Weave::Profiles::DeviceDescription;
+using namespace nl::Weave::Profiles::DataManagement;
 
 using nl::Weave::Profiles::DeviceDescription::IdentifyDeviceCriteria;
 using nl::Weave::System::PacketBuffer;
@@ -164,6 +171,34 @@ extern "C" {
     NL_DLL_EXPORT void Java_nl_Weave_DeviceManager_WeaveStack_handleUnsubscribeComplete(JNIEnv *env, jobject self, jint connObj, jbyteArray svcIdObj, jbyteArray charIdObj, jboolean success);
     NL_DLL_EXPORT void Java_nl_Weave_DeviceManager_WeaveStack_handleConnectionError(JNIEnv *env, jobject self, jint connObj);
     NL_DLL_EXPORT jobject Java_nl_Weave_DeviceManager_WeaveDeviceDescriptor_decode(JNIEnv *env, jclass cls, jbyteArray encodedDesc);
+#if WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_WdmClientImpl_init(JNIEnv *env, jobject self);
+    NL_DLL_EXPORT jlong Java_nl_Weave_DataManagement_WdmClientImpl_newWdmClient(JNIEnv *env, jobject self, jlong deviceMgrPtr);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_WdmClientImpl_deleteWdmClient(JNIEnv *env, jobject self, jlong wdmClientPtr);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_WdmClientImpl_setNodeId(JNIEnv *env, jobject self, jlong wdmClientPtr, jlong nodeId);
+    NL_DLL_EXPORT jlong Java_nl_Weave_DataManagement_WdmClientImpl_newDataSink(JNIEnv *env, jobject self, jlong wdmClientPtr, jobject resourceIdentifierObj, jlong profileId, jlong instanceId, jstring path);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_WdmClientImpl_beginFlushUpdate(JNIEnv *env, jobject self, jlong wdmClientPtr);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_WdmClientImpl_beginRefreshData(JNIEnv *env, jobject self, jlong wdmClientPtr);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_init(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_shutdown(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_clear(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_beginRefreshData(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setInt(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jlong value, jboolean isConditional, jboolean isSigned);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setDouble(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jdouble value, jboolean isConditional);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setBoolean(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jboolean value, jboolean isConditional);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setString(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jstring value, jboolean isConditional);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setNull(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jboolean isConditional);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setBytes(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jbyteArray value, jboolean isConditional);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setStringArray(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jobjectArray stringArray, jboolean isConditional);
+    NL_DLL_EXPORT jlong Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_getLong(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path);
+    NL_DLL_EXPORT jdouble Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_getDouble(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path);
+    NL_DLL_EXPORT jboolean Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_getBoolean(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path);
+    NL_DLL_EXPORT jstring Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_getString(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path);
+    NL_DLL_EXPORT jboolean Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_isNull(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path);
+    NL_DLL_EXPORT jbyteArray Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_getBytes(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path);
+    NL_DLL_EXPORT jobjectArray Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_getStringArray(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path);
+    NL_DLL_EXPORT jlong Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_getVersion(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr);
+#endif // WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
 };
 
 static void HandleIdentifyDeviceComplete(WeaveDeviceManager *deviceMgr, void *reqState, const WeaveDeviceDescriptor *deviceDesc);
@@ -174,7 +209,10 @@ static void HandleAddNetworkComplete(WeaveDeviceManager *deviceMgr, void *reqSta
 static void HandleGetRendezvousModeComplete(WeaveDeviceManager *deviceMgr, void *reqState, uint16_t modeFlags);
 static void HandleGetFabricConfigComplete(WeaveDeviceManager *deviceMgr, void *reqState, const uint8_t *fabricConfig, uint32_t fabricConfigLen);
 static void HandleDeviceEnumerationResponse(WeaveDeviceManager *deviceMgr, void *reqState, const DeviceDescription::WeaveDeviceDescriptor *devdesc, IPAddress deviceAddr, InterfaceId deviceIntf);
+static void HandleGenericOperationComplete(jobject obj, void *reqState);
 static void HandleSimpleOperationComplete(WeaveDeviceManager *deviceMgr, void *reqState);
+static void HandleWdmClientComplete(void *context, void *reqState);
+static void HandleGenericTraitUpdatableDataSinkComplete(void *context, void *reqState);
 static void HandleNotifyWeaveConnectionClosed(BLE_CONNECTION_OBJECT connObj);
 static bool HandleSendCharacteristic(BLE_CONNECTION_OBJECT connObj, const uint8_t *svcId, const uint8_t *charId, const uint8_t *characteristicData, uint32_t characteristicDataLen);
 static bool HandleSubscribeCharacteristic(BLE_CONNECTION_OBJECT connObj, const uint8_t *svcId, const uint8_t *charId);
@@ -183,7 +221,12 @@ static void HandlePairTokenComplete(WeaveDeviceManager *deviceMgr, void *reqStat
 static void HandleUnpairTokenComplete(WeaveDeviceManager *deviceMgr, void *reqState);
 static bool HandleCloseConnection(BLE_CONNECTION_OBJECT connObj);
 static uint16_t HandleGetMTU(BLE_CONNECTION_OBJECT connObj);
+static void HandleGenericError(jobject obj, void *reqState, WEAVE_ERROR deviceErr, DeviceStatus *devStatus);
 static void HandleError(WeaveDeviceManager *deviceMgr, void *reqState, WEAVE_ERROR err, DeviceStatus *devStatus);
+#if WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
+static void HandleWdmClientError(void *appState, void *reqState, WEAVE_ERROR deviceErr, DeviceStatus *devStatus);
+static void HandleGenericTraitUpdatableDataSinkError(void *context, void *reqState, WEAVE_ERROR deviceErr, DeviceStatus *devStatus);
+#endif // WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
 static void ThrowError(JNIEnv *env, WEAVE_ERROR errToThrow);
 static void ReportError(JNIEnv *env, WEAVE_ERROR cbErr, const char *cbName);
 static void *IOThreadMain(void *arg);
@@ -204,8 +247,15 @@ static WEAVE_ERROR N2J_NetworkInfoArray(JNIEnv *env, const NetworkInfo *inArray,
 static WEAVE_ERROR N2J_DeviceDescriptor(JNIEnv *env, const WeaveDeviceDescriptor& inDeviceDesc, jobject& outDeviceDesc);
 static WEAVE_ERROR N2J_Error(JNIEnv *env, WEAVE_ERROR inErr, jthrowable& outEx);
 static WEAVE_ERROR N2J_DeviceStatus(JNIEnv *env, DeviceStatus& devStatus, jthrowable& outEx);
+static WEAVE_ERROR J2N_ResourceIdentifier(JNIEnv *env, jobject inResourceIdentifier, ResourceIdentifier& outResourceIdentifier);
 static WEAVE_ERROR GetClassRef(JNIEnv *env, const char *clsType, jclass& outCls);
+static WEAVE_ERROR J2N_StdString(JNIEnv *env, jstring inString,  std::string& outString);
 
+static void EngineEventCallback(void * const aAppState,
+                                SubscriptionEngine::EventID aEvent,
+                                const SubscriptionEngine::InEventParam & aInParam, SubscriptionEngine::OutEventParam & aOutParam);
+static void BindingEventCallback (void * const apAppState, const nl::Weave::Binding::EventType aEvent,
+                                  const nl::Weave::Binding::InEventParam & aInParam, nl::Weave::Binding::OutEventParam & aOutParam);
 #if CURRENTLY_UNUSED
 static WEAVE_ERROR J2N_EnumVal(JNIEnv *env, jobject enumObj, int& outVal);
 static WEAVE_ERROR J2N_IntFieldVal(JNIEnv *env, jobject obj, const char *fieldName, jint& outVal);
@@ -1019,7 +1069,7 @@ void Java_nl_Weave_DeviceManager_WeaveDeviceManager_beginRegisterServicePairAcco
                                                                                     jbyteArray pairingToken, jbyteArray pairingInitData)
 {
     WEAVE_ERROR err = WEAVE_NO_ERROR;
-    WeaveDeviceManager  *deviceMgr = (WeaveDeviceManager *)deviceMgrPtr;
+    WeaveDeviceManager *deviceMgr = (WeaveDeviceManager *)deviceMgrPtr;
     const uint8_t *serviceConfigBuf;
     const uint8_t *pairingTokenBuf;
     const uint8_t *pairingInitDataBuf;
@@ -2206,12 +2256,12 @@ exit:
     }
 }
 
-void HandleSimpleOperationComplete(WeaveDeviceManager *deviceMgr, void *reqState)
+void HandleGenericOperationComplete(jobject obj, void *reqState)
 {
     WEAVE_ERROR err = WEAVE_NO_ERROR;
     JNIEnv *env;
-    jobject self = (jobject)deviceMgr->AppState;
-    jclass deviceMgrCls;
+    jobject self = obj;
+    jclass cls;
     jmethodID method;
     char methodName[128];
     bool localFramePushed = false;
@@ -2224,12 +2274,12 @@ void HandleSimpleOperationComplete(WeaveDeviceManager *deviceMgr, void *reqState
         localFramePushed = true;
     VerifyOrExit(localFramePushed, err = WEAVE_ERROR_NO_MEMORY);
 
-    deviceMgrCls = env->GetObjectClass(self);
-    VerifyOrExit(deviceMgrCls != NULL, err = WDM_JNI_ERROR_TYPE_NOT_FOUND);
+    cls = env->GetObjectClass(self);
+    VerifyOrExit(cls != NULL, err = WDM_JNI_ERROR_TYPE_NOT_FOUND);
 
     snprintf(methodName, sizeof(methodName) - 1, "on%sComplete", (const char *)reqState);
     methodName[sizeof(methodName) - 1] = 0;
-    method = env->GetMethodID(deviceMgrCls, methodName, "()V");
+    method = env->GetMethodID(cls, methodName, "()V");
     VerifyOrExit(method != NULL, err = WDM_JNI_ERROR_METHOD_NOT_FOUND);
 
     WeaveLogProgress(DeviceManager, "Calling Java %s method", methodName);
@@ -2246,6 +2296,26 @@ exit:
         env->PopLocalFrame(NULL);
 }
 
+void HandleSimpleOperationComplete(WeaveDeviceManager *deviceMgr, void *reqState)
+{
+    jobject self = (jobject)deviceMgr->AppState;
+    HandleGenericOperationComplete(self, reqState);
+}
+#if WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
+void HandleWdmClientComplete(void *context, void *reqState)
+{
+    WdmClient * wdmClient = reinterpret_cast<WdmClient *>(context);
+    jobject self = (jobject)wdmClient->mpAppState;
+    HandleGenericOperationComplete(self, reqState);
+}
+
+void HandleGenericTraitUpdatableDataSinkComplete(void *context, void *reqState)
+{
+    GenericTraitUpdatableDataSink * sink = reinterpret_cast<GenericTraitUpdatableDataSink *>(context);
+    jobject self = (jobject)sink->mpAppState;
+    HandleGenericOperationComplete(self, reqState);
+}
+#endif // WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
 void HandleNotifyWeaveConnectionClosed(BLE_CONNECTION_OBJECT connObj)
 {
     WEAVE_ERROR err = WEAVE_NO_ERROR;
@@ -2584,12 +2654,12 @@ exit:
     return mtu;
 }
 
-void HandleError(WeaveDeviceManager *deviceMgr, void *reqState, WEAVE_ERROR deviceMgrErr, DeviceStatus *devStatus)
+void HandleGenericError(jobject obj, void *reqState, WEAVE_ERROR deviceErr, DeviceStatus *devStatus)
 {
     WEAVE_ERROR err = WEAVE_NO_ERROR;
     JNIEnv *env;
-    jobject self = (jobject)deviceMgr->AppState;
-    jclass deviceMgrCls;
+    jobject self = obj;
+    jclass cls;
     jmethodID method;
     jthrowable ex;
     bool localFramePushed = false;
@@ -2602,21 +2672,21 @@ void HandleError(WeaveDeviceManager *deviceMgr, void *reqState, WEAVE_ERROR devi
         localFramePushed = true;
     VerifyOrExit(localFramePushed, err = WEAVE_ERROR_NO_MEMORY);
 
-    if (deviceMgrErr == WEAVE_ERROR_STATUS_REPORT_RECEIVED && devStatus != NULL)
+    if (deviceErr == WEAVE_ERROR_STATUS_REPORT_RECEIVED && devStatus != NULL)
     {
         err = N2J_DeviceStatus(env, *devStatus, ex);
         SuccessOrExit(err);
     }
     else
     {
-        err = N2J_Error(env, deviceMgrErr, ex);
+        err = N2J_Error(env, deviceErr, ex);
         SuccessOrExit(err);
     }
 
-    deviceMgrCls = env->GetObjectClass(self);
-    VerifyOrExit(deviceMgrCls != NULL, err = WDM_JNI_ERROR_TYPE_NOT_FOUND);
+    cls = env->GetObjectClass(self);
+    VerifyOrExit(cls != NULL, err = WDM_JNI_ERROR_TYPE_NOT_FOUND);
 
-    method = env->GetMethodID(deviceMgrCls, "onError", "(Ljava/lang/Throwable;)V");
+    method = env->GetMethodID(cls, "onError", "(Ljava/lang/Throwable;)V");
     VerifyOrExit(method != NULL, err = WDM_JNI_ERROR_METHOD_NOT_FOUND);
 
     WeaveLogProgress(DeviceManager, "Calling Java onError method");
@@ -2632,6 +2702,28 @@ exit:
     if (localFramePushed)
         env->PopLocalFrame(NULL);
 }
+
+void HandleError(WeaveDeviceManager *deviceMgr, void *reqState, WEAVE_ERROR deviceErr, DeviceStatus *devStatus)
+{
+    jobject self = (jobject)deviceMgr->AppState;
+    HandleGenericError(self, reqState, deviceErr, devStatus);
+}
+
+#if WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
+void HandleWdmClientError(void *context, void *reqState, WEAVE_ERROR deviceErr, DeviceStatus *devStatus)
+{
+    WdmClient * wdmClient = reinterpret_cast<WdmClient *>(context);
+    jobject self = (jobject)wdmClient->mpAppState;
+    HandleGenericError(self, reqState, deviceErr, devStatus);
+}
+
+void HandleGenericTraitUpdatableDataSinkError(void *context, void *reqState, WEAVE_ERROR deviceErr, DeviceStatus *devStatus)
+{
+    GenericTraitUpdatableDataSink * sink = reinterpret_cast<GenericTraitUpdatableDataSink *>(context);
+    jobject self = (jobject)sink->mpAppState;
+    HandleGenericError(self, reqState, deviceErr, devStatus);
+}
+#endif // WEAVE_CONFIG_DATA_MANAGEMENT_CLIENT_EXPERIMENTAL
 
 void ThrowError(JNIEnv *env, WEAVE_ERROR errToThrow)
 {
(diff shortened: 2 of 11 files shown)


src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -3321,6 +3452,631 @@ WEAVE_ERROR GetClassRef(JNIEnv *env, const char *clsType, jclass& outCls)
...
+            err = aInParam.BindingFailed.Reason;
+            WeaveLogDetail(DeviceManager, "kEvent_BindingFailed: reason %d", err);
+            break;
+
+        case nl::Weave::Binding::kEvent_BindingReady:
+            WeaveLogDetail(DeviceManager, "kEvent_BindingReady");
+            break;
+
+        case nl::Weave::Binding::kEvent_DefaultCheck:
+            WeaveLogDetail(DeviceManager, "kEvent_DefaultCheck");
```
g-coppock: Is this intentionally falling through to the default case, or should there be a break statement? If the former, maybe add an explicit `// fallthrough` comment?

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -3321,6 +3452,631 @@ WEAVE_ERROR GetClassRef(JNIEnv *env, const char *clsType, jclass& outCls)
...
+
+    if (err != WEAVE_NO_ERROR && err != WDM_JNI_ERROR_EXCEPTION_THROWN)
+    {
+        ThrowError(env, err);
+    }
+}
+
+void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_init(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr)
+{
+    GenericTraitUpdatableDataSink  *pDataSink = NULL;
```
g-coppock: Here and the next several, there's two spaces between the type and the *

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -3321,6 +3452,631 @@ WEAVE_ERROR GetClassRef(JNIEnv *env, const char *clsType, jclass& outCls)
...
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+    const char *pPathStr = NULL;
+    GenericTraitUpdatableDataSink  *pDataSink = (GenericTraitUpdatableDataSink *)genericTraitUpdatableDataSinkPtr;
+
+    WeaveLogProgress(DeviceManager, "setBoolean() called");
+
+    pPathStr = env->GetStringUTFChars(path, 0);
+    VerifyOrExit(pPathStr != NULL, err = WEAVE_ERROR_NO_MEMORY);
+
+    err = pDataSink->SetBoolean(pPathStr, value == JNI_TRUE , isConditional == JNI_TRUE);
```
g-coppock: Extra space before the second comma

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -3321,6 +3452,631 @@ WEAVE_ERROR GetClassRef(JNIEnv *env, const char *clsType, jclass& outCls)
...
+        env->ReleaseStringUTFChars(value, pValue);
+    }
+
+    if (err != WEAVE_NO_ERROR && err != WDM_JNI_ERROR_EXCEPTION_THROWN)
+    {
+        ThrowError(env, err);
+    }
+}
+
+void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setNULL(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jboolean isConditional)
```
g-coppock: setNULL -> setNull

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
```diff
@@ -0,0 +1,74 @@
...
+    public void setSigned(String path, long value);
+    public void setSigned(String path, BigInteger value);
+    public void setUnsigned(String path, int value);
+    public void setUnsigned(String path, long value);
+    public void setUnsigned(String path, BigInteger value);
+    public void set(String path, double value);
+    public void set(String path, boolean value);
+    public void set(String path, String value);
+    public void set(String path, byte[] value);
+    public void setNULL(String path);
```
g-coppock: setNULL -> setNull

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
```diff
@@ -0,0 +1,74 @@
...
+import android.util.Log;
+
+import java.math.BigInteger;
+import java.util.EnumSet;
+import java.util.Random;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Iterator;
+
+public interface GenericTraitUpdatableDataSink
```
g-coppock: Will there be a followup for Javadoc documentation? I can help with that if you'd like.

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
```diff
@@ -0,0 +1,74 @@
...
+    public void setUnsigned(String path, long value);
+    public void setUnsigned(String path, BigInteger value);
+    public void set(String path, double value);
+    public void set(String path, boolean value);
+    public void set(String path, String value);
+    public void set(String path, byte[] value);
+    public void setNULL(String path);
+    public int getInt(String path);
+    public long getLong(String path);
+    public BigInteger getBigInteger(String path, int bitLen);
```
g-coppock: For the BigInteger/String/byte[] accessors, is it expected that these will return null if the path is invalid, or it will throw an exception? If null, can these be annotated with @Nullable?

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+package nl.Weave.DataManagement;
+import android.os.Build;
+import android.util.Log;
+
+import java.math.BigInteger;
+/**
+ *  Represents Resource Identifier
+ */
+
```
g-coppock: Move line break to between the import statement and Javadoc start?

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
```diff
@@ -0,0 +1,85 @@
...
+public class ResourceIdentifier
+{
+    public ResourceIdentifier()
+    {
+        ResourceType = RESOURCE_TYPE_RESERVED;
+        ResourceId = SELF_NODE_ID;
+    }
+
+    private int ResourceType;
```
g-coppock: Can we initialize these in the constructor and make them final to make this class immutable? i.e.

```
private final int resourceType;
private final long resourceId;

public ResourceIdentifier(int resourceType, long resourceId) {
    this.resourceType = resourceType;
    this.resourceId = resourceId;
}
```

Note that the field names should start with lowercase as well. For the static factory methods, the method name should also start with lowercase, i.e.

```
public static ResourceIdentifier make(int resourceType, BigInteger resourceId) {
  return new ResourceIdentifier(resourcetype, value.longValue());
}
```

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public static ResourceIdentifier Make(int resourceType, byte[] ResourceIdBytes)
+    {
+        ResourceIdentifier resourceIdentifier = new ResourceIdentifier();
+        resourceIdentifier.ResourceType = resourceType;
+
+        if (ResourceIdBytes.length != 8)
+        {
+            Log.e(TAG, "unexpected resourceIdentifier, ResourceIdBytes should be 8 bytes");
+            return null;
```
g-coppock: Can we throw an `IllegalArgumentException` here instead of returning null?

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        resourceIdentifier.ResourceId = ByteArrayToLong(ResourceIdBytes);
+
+        return resourceIdentifier;
+    }
+
+    public static long ByteArrayToLong(byte[] value)
```
g-coppock: This and LongToByteArray can be private as they're only used internally in the factory methods. 

Additionally, method names should start with lowercase.

src/device-manager/java/src/nl/Weave/DataManagement/WDMClient.java
```diff
@@ -0,0 +1,37 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+ */
+package nl.Weave.DataManagement;
+
+public interface WDMClient
```
g-coppock: Same comment here for `WDMClient` -> `WdmClient`. If this is the existing style for this repo then that's fine, but typically should be camel-case for acronyms.

src/device-manager/java/src/nl/Weave/DataManagement/WDMClient.java
```diff
@@ -0,0 +1,37 @@
...
+{
+    public void connect(long deviceMgrPtr);
```
g-coppock: Can we remove this from the interface? Since this is only required the `WeaveDeviceManager` based `WDMClientImpl`?

src/device-manager/java/src/nl/Weave/DataManagement/WDMClient.java
```diff
@@ -0,0 +1,37 @@
...
+    public void close();
+    public GenericTraitUpdatableDataSinkImpl newDataSink(ResourceIdentifier resourceIdentifier, long profileId, long instanceId, String path);
```
g-coppock: This should just return the interface, not the impl

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,444 @@
...
+        mCompHandler = null;
+    }
+
+    @Override
+    public void setSigned(String path, int value, boolean isConditional)
+    {
+        boolean isSigned = true;
+        if (mTraitInstancePtr == 0)
+        {
+            Log.e(TAG, "unexpected err, mTraitInstancePtr is 0");
```
g-coppock: Here and below, should probably throw an `IllegalStateException`. Since this same code is repeated for each method, consider extracting a private method for this:

```
private void ensureNotClosed() {
  if (mTraitInstancePtr == 0) {
    throw new IllegalStateException("This data sink has already been closed.");
  }
}
```

And just calling that instead.

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,444 @@
...
+            close(mTraitInstancePtr);
+            mTraitInstancePtr = 0;
+        }
+        mCompHandler = null;
+    }
+
+    @Override
+    public void setSigned(String path, int value, boolean isConditional)
+    {
+        boolean isSigned = true;
```
g-coppock: Nothing wrong with this so feel free to ignore, but another common pattern is:

```
setInt(mTraitInstancePtr, path, value, isConditional, /* isSigned= */ true);
```

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,444 @@
...
+    private native void init(long genericTraitUpdatableDataSinkPtr);
+    private native void close(long genericTraitUpdatableDataSinkPtr);
+    private native void beginRefreshData(long genericTraitUpdatableDataSinkPtr);
+    private native void setInt(long genericTraitUpdatableDataSinkPtr, String path, long value, boolean isConditional, boolean isSigned);
+    private native void setDouble(long genericTraitUpdatableDataSinkPtr, String path, double value, boolean isConditional);
+    private native void setBoolean(long genericTraitUpdatableDataSinkPtr, String path, boolean value, boolean isConditional);
+    private native void setString(long genericTraitUpdatableDataSinkPtr, String path, String value, boolean isConditional);
+    private native void setNULL(long genericTraitUpdatableDataSinkPtr, String path, boolean isConditional);
+    private native void setBytes(long genericTraitUpdatableDataSinkPtr, String path, byte[] value, boolean isConditional);
+    private native long getInt(long genericTraitUpdatableDataSinkPtr, String path);
```
g-coppock: Can we rename the native method to `getLong` since it returns a `long`?

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,444 @@
...
+        return getString(mTraitInstancePtr, path);
+    }
+
+    @Override
+    public byte[] getBytes(String path)
+    {
+        if (mTraitInstancePtr == 0)
+        {
+            Log.e(TAG, "unexpected err, mTraitInstancePtr is 0");
+            return null;
```
g-coppock: Would be preferable for these getters to throw the same IllegalStateException if closed, rather than return null

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,444 @@
...
+
+    @Override
+    public long getLong(String path)
+    {
+        if (mTraitInstancePtr == 0)
+        {
+            Log.e(TAG, "unexpected err, mTraitInstancePtr is 0");
+            return 0;
+        }
+        return (long)getInt(mTraitInstancePtr, path);
```
g-coppock: Cast to `long` isn't needed here

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    @Override
+    public BigInteger getBigInteger(String path, int bitLen)
+    {
+        long value = (long)getInt(mTraitInstancePtr, path);
```
g-coppock: Cast to long isn't needed here

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,444 @@
...
+        String valStr = Long.toString(value);
+        BigInteger bigInt = new BigInteger(valStr);
```
g-coppock: You can just use `BigInteger.valueOf(value)` to avoid the intermediate `String`

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
g-coppock: What is `bitLen` here?

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,444 @@
...
+        if (mTraitInstancePtr == 0)
```
g-coppock: This check should be first, since the `mTraitInstancePtr` is used on line 278

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        {
+            Log.e(TAG, "unexpected err, mCompHandler is null");
+            return;
+        }
+        mCompHandler.onRefreshDataComplete();
+    }
+
+    // ----- Protected Members -----
+
+    protected CompletionHandler mCompHandler;
```
g-coppock: This can be private

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Iterator;
+
+public class WDMClientImpl implements WDMClient
+{
+    public WDMClientImpl()
+    {
+        init();
+        mCompHandler = null;
```
g-coppock: This line not needed -- it will default to null

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+    @Override
+    public void connect(long deviceMgrPtr)
+    {
+        mWDMClientPtr = newWDMClient(deviceMgrPtr);
+    }
+
+    @Override
+    public void close()
+    {
+        if (mTraitMap != null)
```
g-coppock: This can be simplified to:

```
if (mTraitMap != null) {
  for (GenericTraitUpdatableDataSink dataSink : mTraitMap.values()) {
    dataink.close();
  }
}
```

Note that the null check on the dataSink can be avoided by not inserting null into the map (instead just removing the key -- will expand on that below).

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        long traitInstancePtr = 0;
+        boolean isExist = false;
+        GenericTraitUpdatableDataSinkImpl traitInstance = null;
+
+        if (mTraitMap == null)
+        {
+            mTraitMap = new HashMap<Long, GenericTraitUpdatableDataSinkImpl>();
+        }
+
+        if (mWDMClientPtr == 0)
```
g-coppock: Can we move the pointer checks to the beginning of the method to fail fast? Otherwise this will leave the map partially initialized.

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        {
+            deleteWDMClient(mWDMClientPtr);
+            mWDMClientPtr = 0;
+        }
+
+        mCompHandler = null;
+    }
+
+    @Override
+    public GenericTraitUpdatableDataSinkImpl newDataSink(ResourceIdentifier resourceIdentifier, long profileId, long instanceId, String path)
```
g-coppock: Should return the interface type instead of implementation

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        traitInstancePtr = newDataSink(mWDMClientPtr, resourceIdentifier, profileId, instanceId, path);
+
+        if (traitInstancePtr == 0)
+        {
+            mCompHandler.onError(new Exception("Fail to create traitInstancePtr."));
+            Log.e(TAG, "traitInstancePtr is not ready");
+            return null;
+        }
+
+        isExist = mTraitMap.containsKey(traitInstancePtr);
```
g-coppock: To ensure a null value is never returned, consider changing this to:

```
GenericTraitUpdatableDataSink trait = mTraitMap.get(traitInstancePtr);
if (trait == null) {
  trait = new GenericTraitUpdatableDataSinkImpl(traitInstancePtr, this);
  mTraitMap.put(traitInstancePtr, traitInstance);
}
return trait;
```

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        Log.d(TAG, "there exists %d traits" + mTraitMap.size());
+
+        return mTraitMap.get(traitInstancePtr);
+    }
+
+    protected void removeDataSinkRef(long traitInstancePtr)
+    {
+        boolean isExist = false;
```
g-coppock: This can just be:

```
if (mTraitMap != null) {
  mTraitMap.remove(traitInstancePtr);
}
```

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        if (isExist == true)
+        {
+            mTraitMap.put(traitInstancePtr, null);
+        }
+    }
+
+    @Override
+    public void beginFlushUpdate()
+    {
+        if (mWDMClientPtr == 0)
```
g-coppock: Same comments here for all of these pointer checks. You can extract a private method that throws an `IllegalStateException` if the pointer isn't initialized and use it throughout.

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
```diff
@@ -0,0 +1,213 @@
...
+    protected void finalize() throws Throwable
+    {
+        super.finalize();
+        close();
+    }
+
+    // ----- Private Members -----
+
+    private long mWDMClientPtr;
+    private HashMap<Long, GenericTraitUpdatableDataSinkImpl> mTraitMap;
```
g-coppock: nit: can be typed as `Map` instead of `HashMap` (interface vs implementation)

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        long result = 0;
+        for (int i = 0; i < value.length; i++)
+        {
+           result += ((long) value[i] & 0xffL) << (8 * i);
+        }
+        return result;
+    }
```
g-coppock: Can you implement an equals and hashCode() method here?

```
@Override
public boolean equals(@Nullable Object other) {
  if (other == this) {
    return true;
  } else if (other instanceof ResourceIdentifier) {
    ResourceIdentifier otherId = (ResourceIdentifier) other;
    return this.resourceType == otherId.resourceType && this.resourceId == otherId.resourceId;
  } else {
    return false;
  }
}

@Override
public int hashCode() {
  return Objects.hash(resourceId, resourceType);
}
```

src/device-manager/java/Makefile.am
```diff
@@ -103,6 +103,13 @@ WeaveDeviceManager_jar_JAVA_SRCS                   =        \
     nl/Weave/DeviceManager/WeaveDeviceException.java        \
     nl/Weave/DeviceManager/DeviceFeatures.java              \
     nl/Weave/DeviceManager/NevisPairingCodeDecoding.java    \
+	nl/Weave/DeviceManager/WeaveDeviceManager.java 				    \
+	nl/Weave/DataManagement/WDMClient.java 				    \
+	nl/Weave/DataManagement/WDMClientImpl.java 				    \
+	nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java 				    \
+	nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java \
+	nl/Weave/DataManagement/WDMClientFactory.java 				    \
+	nl/Weave/DataManagement/ResourceIdentifier.java 				    \
```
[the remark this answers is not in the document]
yunhanw-google: Done, thanks

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -164,6 +166,29 @@ extern "C" {
...
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setInt(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jlong value, jboolean isConditional, jboolean isSigned);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setDouble(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jdouble value, jboolean isConditional);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setBoolean(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jboolean value, jboolean isConditional);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setString(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jstring value, jboolean isConditional);
+    NL_DLL_EXPORT void Java_nl_Weave_DataManagement_GenericTraitUpdatableDataSinkImpl_setNULL(JNIEnv *env, jobject self, jlong genericTraitUpdatableDataSinkPtr, jstring path, jboolean isConditional);
```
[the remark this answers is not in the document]
yunhanw-google: Done, thanks

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -174,7 +199,10 @@ static void HandleAddNetworkComplete(WeaveDeviceManager *deviceMgr, void *reqSta
 static void HandleGetRendezvousModeComplete(WeaveDeviceManager *deviceMgr, void *reqState, uint16_t modeFlags);
 static void HandleGetFabricConfigComplete(WeaveDeviceManager *deviceMgr, void *reqState, const uint8_t *fabricConfig, uint32_t fabricConfigLen);
 static void HandleDeviceEnumerationResponse(WeaveDeviceManager *deviceMgr, void *reqState, const DeviceDescription::WeaveDeviceDescriptor *devdesc, IPAddress deviceAddr, InterfaceId deviceIntf);
+static void HandleGenericOperationComplete(jobject obj, void *reqState);
 static void HandleSimpleOperationComplete(WeaveDeviceManager *deviceMgr, void *reqState);
+static void HandleWDMClientComplete(void *context, void *reqState);
```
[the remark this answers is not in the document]
yunhanw-google: Yep, thanks

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -2765,6 +2842,43 @@ WEAVE_ERROR N2J_NewStringUTF(JNIEnv *env, const char *inStr, jstring& outString)
 
     return err;
 }
+
+WEAVE_ERROR N2J_NewStringUTFViaByteArray(JNIEnv *env, const uint8_t *inArray, uint32_t inArrayLen, jstring& outString)
```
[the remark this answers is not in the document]
yunhanw-google: Done, thanks

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -2765,6 +2842,43 @@ WEAVE_ERROR N2J_NewStringUTF(JNIEnv *env, const char *inStr, jstring& outString)
...
+
+    exit:
+    // error code propagated from here, so clear any possible
+    // exceptions that arose here
+    env->ExceptionClear();
+
+    if (utf8Encoding != NULL)
+    env->DeleteLocalRef(utf8Encoding);
+    if (charArray != NULL)
+    env->DeleteLocalRef(charArray);
```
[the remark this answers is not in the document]
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
yunhanw-google: Removed, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: If multiple data sink are being deallocated, and WDMClient is being deallocated by java garbage collection in parallel, the mTraitMap may be updated dynamically, the close may execute on the one that has been deallocated, the crash may happen.

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
yunhanw-google: @Nullable is from javax, which is not generic? I cannot use javax when compiling weave

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Will do it, thanks

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -2765,6 +2842,43 @@ WEAVE_ERROR N2J_NewStringUTF(JNIEnv *env, const char *inStr, jstring& outString)
...
+    outString = (jstring) env->NewObject(java_lang_String, ctor, charArray, utf8Encoding);
+    VerifyOrExit(outString != NULL, err = WEAVE_ERROR_NO_MEMORY);
+
+    exit:
+    // error code propagated from here, so clear any possible
+    // exceptions that arose here
+    env->ExceptionClear();
+
+    if (utf8Encoding != NULL)
+    env->DeleteLocalRef(utf8Encoding);
```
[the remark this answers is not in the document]
yunhanw-google: We have used bytesData to get string, then generate  valueStr and pass it to upper java layer
err = pDataSink->GetString(pPathStr, &bytesData);
    SuccessOrExit(err);

    err = N2J_NewStringUTF8ViaByteArray(env, bytesData.mpDataBuf, bytesData.mDataLen, valueStr);
    SuccessOrExit(err);

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -2765,6 +2842,43 @@ WEAVE_ERROR N2J_NewStringUTF(JNIEnv *env, const char *inStr, jstring& outString)
 
     return err;
 }
+
+WEAVE_ERROR N2J_NewStringUTFViaByteArray(JNIEnv *env, const uint8_t *inArray, uint32_t inArrayLen, jstring& outString)
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+    jbyteArray charArray = NULL;
+    jstring utf8Encoding = NULL;
+    jclass java_lang_String = NULL;
```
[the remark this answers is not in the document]
yunhanw-google: roughly I keep consistency between N2J_NewStringUTF and N2J_NewStringUTF8ViaByteArray

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -2765,6 +2842,43 @@ WEAVE_ERROR N2J_NewStringUTF(JNIEnv *env, const char *inStr, jstring& outString)
...
+    utf8Encoding = env->NewStringUTF("UTF-8");
+    VerifyOrExit(utf8Encoding != NULL, err = WEAVE_ERROR_NO_MEMORY);
+
+    java_lang_String = env->FindClass("java/lang/String");
+    VerifyOrExit(java_lang_String != NULL, err = WDM_JNI_ERROR_TYPE_NOT_FOUND);
+
+    ctor = env->GetMethodID(java_lang_String, "<init>", "([BLjava/lang/String;)V");
+    VerifyOrExit(ctor != NULL, err = WDM_JNI_ERROR_METHOD_NOT_FOUND);
+
+    outString = (jstring) env->NewObject(java_lang_String, ctor, charArray, utf8Encoding);
```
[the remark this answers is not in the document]
yunhanw-google: yes, we need DeleteLocalRef for outString, I am fixing it via 
    err = N2J_NewStringUTF8ViaByteArray(env, bytesData.mpDataBuf, bytesData.mDataLen, valueStr);
    SuccessOrExit(err);

exit:
    if (pPathStr != NULL)
    {
        env->ReleaseStringUTFChars(path, pPathStr);
    }

    if (err != WEAVE_NO_ERROR && err != WDM_JNI_ERROR_EXCEPTION_THROWN)
    {
        ThrowError(env, err);
    }

    env->DeleteLocalRef(valueStr);

src/device-manager/java/WeaveDeviceManager-JNI.cpp
yunhanw-google: yes, it is intentionally failing through, i am adding the comment.
        case nl::Weave::Binding::kEvent_DefaultCheck:
            WeaveLogDetail(DeviceManager, "kEvent_DefaultCheck");
            // fall through
thanks

src/device-manager/java/WeaveDeviceManager-JNI.cpp
yunhanw-google: thanks for this catch. fixing it.

src/device-manager/java/WeaveDeviceManager-JNI.cpp
yunhanw-google: thanks, fixed it.

src/device-manager/java/WeaveDeviceManager-JNI.cpp
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
yunhanw-google: Yep, please comment it. I have added the simple description on each file.

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: agree. updated the naming on it. thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClient.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClient.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
yunhanw-google: just update it to lowercase.
This has been used as helper function in testMain so I make it as public. you may also need it?
ResourceIdentifier resourceIdentifier = ResourceIdentifier.make(ResourceIdentifier.RESOURCE_TYPE_RESERVED, ResourceIdentifier.longToByteArray(-2))

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
yunhanw-google: I might have done it in latest code, please help recheck, thanks

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: Done, thanks.

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
yunhanw-google: @nullable is from javax, which is not generic? I cannot use javax when compiling weave.

in jni layer, for example, when getLong failed, it will ThrowError
    if (err != WEAVE_NO_ERROR && err != WDM_JNI_ERROR_EXCEPTION_THROWN)
    {
        ThrowError(env, err);
    }

void ThrowError(JNIEnv *env, WEAVE_ERROR errToThrow)
{
    WEAVE_ERROR err = WEAVE_NO_ERROR;
    jthrowable ex;

    err = N2J_Error(env, errToThrow, ex);
    if (err == WEAVE_NO_ERROR)
        env->Throw(ex);
}

src/device-manager/java/WeaveDeviceManager-JNI.cpp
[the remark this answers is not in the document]
yunhanw-google: maximum transmission unit (MTU), several connected words forms the naming abbreviation(MTU). TCP, DNS are another two examples.
WDM is also abbreviation of several words, weave data management.
For these abbreviation, I feel either all lowercase or uppercase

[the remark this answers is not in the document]
yunhanw-google: Just rename it to Wdm. I think we should follow cameral pattern. Thanks for the suggestion.

src/device-manager/java/src/nl/Weave/DataManagement/WDMClient.java
yunhanw-google: Done, rename it to WdmClient

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: same as the above comment, If multiple data sink are being deallocated, and WDMClient is being deallocated by java garbage collection in parallel, the mTraitMap may be updated dynamically, the close may execute on the one that has been deallocated, the crash may happen.

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
g-coppock: I don't think that could be an issue here. If this WDMClientImpl is holding a reference to the GenericTraitUpdatableDataSink, then they will not be garbage collected. So by definition of holding a reference to them here you should not run into any concurrency issues with GC.

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
g-coppock: Same here, don't think you can have a concurrency issue with GC here.

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
```diff
@@ -0,0 +1,78 @@
...
+    public void setSigned(String path, long value, boolean isConditional);
+    public void setSigned(String path, BigInteger value, boolean isConditional);
+    public void setUnsigned(String path, int value, boolean isConditional);
+    public void setUnsigned(String path, long value, boolean isConditional);
+    public void setUnsigned(String path, BigInteger value, boolean isConditional);
+    public void set(String path, double value, boolean isConditional);
+    public void set(String path, boolean value, boolean isConditional);
+    public void set(String path, String value, boolean isConditional);
+    public void set(String path, byte[] value, boolean isConditional);
+    public void setNull(String path, boolean isConditional);
```
g-coppock: Was just thinking about this a bit more. Should we remove the setNull() function entirely, and use wrapper classes on the set functions? e.g.

setSigned(String path, Integer value, boolean isConditional);
setSigned(String path, Long value, boolean isConditional);
set(String path, Double value, boolean isConditional);
set(String path, String value, boolean isConditional);

That way, you can use the same function, just providing null as the argument. It does incur some auto-boxing overhead but I don't think that's really an issue here.

jaylogue: I think that's fine idea, but I would prefer to keep setNull() as well.

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
```diff
@@ -0,0 +1,78 @@
...
+import java.math.BigInteger;
+import java.util.EnumSet;
+import java.util.Random;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Iterator;
+
+public interface GenericTraitUpdatableDataSink
+{
+    public void setSigned(String path, int value, boolean isConditional);
```
g-coppock: For setSigned() methods:

```
/**
 * Assigns the provided value to the given path as a signed integer value.
 * 
 * @param path the proto path to the property to modify
 * @param value the new value to assign to the property
 * @param isConditional whether or not to allow overwriting any conflicting changes. If true, then if a later
 *     version of the trait has modified this property, this update will be dropped; otherwise, this value will 
 *     overwrite the newer change
 */
```
 
For setUnsigned() methods:

```
/**
 * Assigns the provided value to the given path as an unsigned integer value.
 * 
 * @param path the proto path to the property to modify
 * @param value the new value to assign to the property
 * @param isConditional whether or not to allow overwriting any conflicting changes. If true, then if a later
 *     version of the trait has modified this property, this update will be dropped; otherwise, this value will 
 *     overwrite the newer change
 */
```
 
For general set() methods:

```
/**
 * Assigns the provided value to the given path.
 * 
 * @param path the proto path to the property to modify
 * @param value the new value to assign to the property
 * @param isConditional whether or not to allow overwriting any conflicting changes. If true, then if a later
 *     version of the trait has modified this property, this update will be dropped; otherwise, this value will 
 *     overwrite the newer change
 */
```
 
For getters() could do something like this, although throwing a generic "Exception" isn't great. Not sure what the C++ code throws for this, but it could be nice to define our own exceptions like `InvalidPathException` or `InvalidTypeException` and throw specific errors depending on what case is causing the failure.

```
/**
 * Returns the value assigned to the property at the given path within this trait.
 *
 * @throws Exception if no property exists at this path, or if the type of the property does not match
 */
```
 
For getVersion()

```
/** Returns the version of the trait represented by this data sink. */
```

For beginRefreshData():

```
/** 
 * Begins a sync of the trait data. The result of this operation can be observed through the {@link CompletionHandler}
 * that has been assigned via {@link #setCompletionHandler}.
 */
```

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: @jaylogue what do you think of? using this way, we can remove several similar functions for integers, even for integer array.

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
g-coppock: @jaylogue any ideas if we can add the jsr305 lib as a dependency easily? This would be very helpful for our use from Kotlin code in the Nest/Home apps.

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
g-coppock: Looks good, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,358 @@
...
+    public void setCompletionHandler(CompletionHandler compHandler)
+    {
+        mCompHandler = compHandler;
+    }
+
+    private void onError(Throwable err)
+    {
+        if (mCompHandler == null)
+        {
+            throw new IllegalStateException("unexpected err, mCompHandler is null");
```
g-coppock: Since this will be surfaced to the client, consider an error such as:

"No CompletionHandler is set. Did you forget to call setCompletionHandler()?"

Same for onRefreshdataComplete()

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -0,0 +1,358 @@
...
+
+    private void ensureNotClosed() {
+      if (mTraitInstancePtr == 0) {
+        throw new IllegalStateException("This data sink has already been closed.");
+      }
+    }
+
+    // ----- Protected Members -----
+
+    private CompletionHandler mCompHandler;
```
g-coppock: Move down with private members?

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
```diff
@@ -0,0 +1,105 @@
...
+        this.resourceType = resourceType;
+        this.resourceId = resourceId;
+    }
+
+    private final int resourceType;
+    private final long resourceId;
+
+    public static ResourceIdentifier make(int resourceType, BigInteger resourceId)
+    {
+        ResourceIdentifier resourceIdentifier = new ResourceIdentifier(resourceType, resourceId.longValue());
```
g-coppock: Could inline the return, here and the next method. i.e.

```
return new ResourceIdentifier(resourceType, resourceId.longValue());
```

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientImpl.java
```diff
@@ -0,0 +1,205 @@
...
+    public void setCompletionHandler(CompletionHandler compHandler)
+    {
+        mCompHandler = compHandler;
+    }
+
+    private void onError(Throwable err)
+    {
+        if (mCompHandler == null)
+        {
+            throw new IllegalStateException("unexpected err, mCompHandler is null");
```
g-coppock: Could extract a private method here to avoid duplication:

```
private CompletionHandler requireCompletionHandler() {
  return Objects.requireNonNull(
      mCompletionHandler, "No CompletionHandler set. Did you forget to call setCompletionHandler()?");
}
```

Then just use it like:

```
requireCompletionHandler().onError(err);
requireCompletionHandler().onFlushUpdateComplete();
```

etc.

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientImpl.java
```diff
@@ -0,0 +1,205 @@
...
+
+    private void ensureNotClosed() {
+      if (mWdmClientPtr == 0) {
+        throw new IllegalStateException("This WdmClient has already been closed.");
+      }
+    }
+
+    // ----- Protected Members -----
+
+    private CompletionHandler mCompHandler;
```
g-coppock: Move to private section?

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
g-coppock: I might have been misunderstood here. I just meant instead of having a local boolean, inlining the value with a comment. i.e. instead of:

```
boolean isSigned = true;
setInt(mTraitInstancePtr, path, value, isConditional, isSigned);
```

You can just write:

```
setInt(mTraitInstancePtr, path, value, isConditional, /* isSigned= */ true);
```

yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: ok, thanks, done

src/device-manager/java/src/nl/Weave/DataManagement/WDMClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
yunhanw-google: I am keeping setNull here, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
yunhanw-google: thanks, will add the rest of them.

src/device-manager/java/src/nl/Weave/DataManagement/ResourceIdentifier.java
yunhanw-google: done, thanks!

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientImpl.java
yunhanw-google: Yep, done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: yep, done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: just update the error message, done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientImpl.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSink.java
yunhanw-google: regarding the specific exception, InvalidPathException , InvalidTypeException , it looks fine, I would prefer address it in later PR on it

yunhanw-google:
all comments are addressed, thanks

PR #445, opened 2019-12-10
Added Operational Device Credentials for Testing Purposes.

emargolis:
Added Test CA credentials.

These credentials will be used to generate test operational certificate by the Mock CA service.


src/test-apps/Certs.cpp
```diff
@@ -475,6 +476,658 @@ uint8_t TestDevice10_PrivateKey[] =
 uint16_t TestDevice10_PrivateKeyLength = sizeof(TestDevice10_PrivateKey);
 
 
+// Operational credentials: node Id, self-signed certificate, service-assigned certificate and private key for Test Device 1
```
jaylogue: Why do we need per-formed operational credentials?  Isn't that counter to the whole idea?

emargolis: I was thinking to use these pre-formed operational credentials in some unit level testing. But I think you are making valid point and we can always generate operational credentials as we need them. I'll update this PR and leave only Test CA credentials.

PR #446, opened 2019-12-11
Add automatic re-bootstrapping to the convenience Makefiles.

robszewczyk:


robszewczyk:
Note to reviewers: I've opted to identify the successful `bootstrap` run via `Makefile.in` rather than `configure` because `configure` tended to conflict with the phony target in the convenience Makefiles.

PR #454, opened 2019-12-12
Relocate the test-driver script to clarify its ownership

robszewczyk:
weave-parallel-test-engine was originally based on the autotools'
test-driver.  Weave's version of the script added better signal
handling as well as a few tweaks to better support parallel test
execution.  Regrettably, these tweaks are quite specific to Weave, and
as a result awkward to upstream.  This commit relocates the script to
the location most closely related in the extended tree -- we stipulate
the appropriate place to land this script is in the
nlbuild-autotools/repo/third_party/autoconf, right next to the
test-driver from which our script originated.


robszewczyk:
Note to the reviewers:  This solution is not ideal, as the `weave-parallel-test-engine`, a modification of autoconf's `test-driver` likely should not be upstreamed into nlbuild-autotools.  If @gerickson is willing to accept such a script into nlbuild-autotools (subject to approriate generalization) I would be excited to pursue such a path.  Other solutions I have considered:

* maintaining a patch in Weave to be applied against `nlbuild-autotools/repo/third_party/autoconf/test-driver`.  Pro: clear separation of ownership.  Cons: fragile solution, patch itself cannot carry copyright, location for the patch is not obvious
* splitting out a separate repository for `weave-parallel-test-engine`, followed by submodule/clone flow. Pros: defensible organization, clear separation of ownership.  Cons: overhead. 
* Different location for the script in `third_party/` .  I could not find a better place for the master location that would clearly delineate the ownership, claims, and provenance of the script.

src/test-apps/Makefile.am
```diff
@@ -74,7 +73,7 @@ SUBDIRS                                               = \
 # Define a package-specific test driver (for 'make check').
 #
 
-LOG_DRIVER=$(abs_srcdir)/weave-parallel-test-engine
+LOG_DRIVER=$(abs_top_nlbuild_autotools_dir)/third_party/autoconf/weave-parallel-test-engine
```
gerickson: It may be OK for openweave-core to place this here in its own copy of nlbuild-autotools; however, I don't think this belongs in nlbuild-autotools unless the changes can be generically upstreamed.

gerickson:
> Different location for the script in third_party/ . I could not find a better place for the master location that would clearly delineate the ownership, claims, and provenance of the script.

Any reason to not put it in third_party/autoconf/repo/<wherever is appropriate relative to where test-driver lives in the upstream project?

src/test-apps/Makefile.am
robszewczyk: I wholeheartedly agree.  Now, there is a set of changes that I could see being generically useful:

* idea of grace runs for flaky tests (though it pains me to admit it, it seems like a pragmatic feature)
* better signal handling (e.g. if the test under consideration was killed via a signal)
* an environment variable that conveys a unique number to the program under test in a parallel execution scenario

robszewczyk:
> > Different location for the script in third_party/ . I could not find a better place for the master location that would clearly delineate the ownership, claims, and provenance of the script.
> 
> Any reason to not put it in third_party/autoconf/repo/<wherever is appropriate relative to where test-driver lives in the upstream project?

SGTM

robszewczyk:
> > > Different location for the script in third_party/ . I could not find a better place for the master location that would clearly delineate the ownership, claims, and provenance of the script.
> > 
> > 
> > Any reason to not put it in third_party/autoconf/repo/<wherever is appropriate relative to where test-driver lives in the upstream project?
> 
> SGTM

Script relocated, commit message updated.  Thanks for the suggestion, @gerickson !

PR #455, opened 2019-12-13
Implemented Certificate Provisioning Protocol Engine.

emargolis:
-- Defined abstract interface object (WeaveNodeOpAuthDelegate)
     to which platform specific actions are delegated during
     Weave node operational authentication.
  -- Defined abstract interface object (WeaveNodeManufAttestDelegate)
     to which platform specific actions are delegated during
     Weave node manufacturer attestation.
  -- Defined and implemented certificate provisioning engine, used to
     get new Weave operational device certificate from the CA service.

### DIFF
--- a/src/lib/profiles/security/WeaveCertProvisioning.cpp
+++ b/src/lib/profiles/security/WeaveCertProvisioning.cpp
@@ -1,0 +1,773 @@
+/*
+ *
+ *    Copyright (c) 2019-2020 Google LLC.
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ *    @file
+ *      This file implements the Certificate Provisioning Protocol, used to
+ *      get new Weave operational device certificate from the CA service.
+ *
+ */
+
+#include <Weave/Core/WeaveCore.h>
+#include <Weave/Core/WeaveTLV.h>
+#include <Weave/Support/CodeUtils.h>
+#include <Weave/Core/WeaveEncoding.h>
+#include <Weave/Profiles/WeaveProfiles.h>
+#include "WeaveCertProvisioning.h"
+#include <Weave/Profiles/common/CommonProfile.h>
+#include <Weave/Support/crypto/HashAlgos.h>
+
+namespace nl {
+namespace Weave {
+namespace Profiles {
+namespace Security {
+namespace CertProvisioning {
+
+using namespace nl::Weave::Crypto;
+using namespace nl::Weave::Encoding;
+using namespace nl::Weave::TLV;
+using namespace nl::Weave::Platform::Security;
+
+// ==================== Public Members ====================
+
+/**
+ * Initialize a WeaveCertProvEngine object in preparation for sending get certificate request message.
+ *
+ *  @param[in]  binding              A Binding object that will be used to establish communication with the
+ *                                   peer node.
+ *  @param[in]  opAuthDelegate       A pointer to a operational authentication delegate object that will be used to
+ *                                   construct and sign using node's operatational credentials.
+ *  @param[in]  mfrAttestDelegate    A pointer to a manufacturer attestation delegate object that will be used to
+ *                                   construct and sign request using node's manufacturer provisioned credentials.
+ *  @param[in]  eventCallback        A pointer to a function that will be called by the WeaveCertProvEngine object
+ *                                   to deliver API events to the application.
+ *  @param[in]  appState             A pointer to an application-defined object which will be passed back
+ *                                   to the application whenever an API event occurs.
+ */
+WEAVE_ERROR WeaveCertProvEngine::Init(Binding * binding, WeaveNodeOpAuthDelegate * opAuthDelegate, WeaveNodeMfrAttestDelegate * mfrAttestDelegate, EventCallback eventCallback, void * appState)
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+
+    VerifyOrExit(opAuthDelegate != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
+
+    VerifyOrExit(eventCallback != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
+
+    AppState = appState;
+    mReqType = kReqType_NotSpecified;
+    mDoMfrAttest = false;
+    mBinding = binding;
+    mOpAuthDelegate = opAuthDelegate;
+    mMfrAttestDelegate = mfrAttestDelegate;
+    mEventCallback = eventCallback;
+    mEC = NULL;
+
+    // Retain a reference to the binding object.
+    if (binding != NULL)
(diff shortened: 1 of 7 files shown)


jaylogue:
Please replace the "manuf" abbreviation with "mfr" / "mfg".

src/lib/profiles/security/WeaveCertProvisioning.cpp
```diff
@@ -0,0 +1,767 @@
...
+            // Deliver a CommunicationError API event to the application.
+            client->mEventCallback(client->AppState, kEvent_CommunicationError, inParam, outParam);
+        }
+    }
+
+    // Process GetCertificateResponse message from the CA service.
+    else if (profileId == kWeaveProfile_Security && msgType == kMsgType_GetCertificateResponse)
+    {
+        // If the request is now complete, clear the request state and enter Idle state.
+        client->HandleRequestDone();
```
jaylogue: I think this is broken.  HandleRequestDone() always aborts the EC.  In this case, however, aborting the will result in the ACK never being sent to the service.   A simple solution would be to expressly close (and NULL) the EC here before calling HandleRequestDone().

src/lib/profiles/security/WeaveCertProvisioning.cpp
```diff
@@ -0,0 +1,767 @@
...
+
+exit:
+    // Free the payload buffer if it hasn't been sent.
+    if (msgBuf != NULL)
+        PacketBuffer::Free(msgBuf);
+
+    if (err != WEAVE_NO_ERROR)
+    {
+        // Enter Idle state.
+        mState = kState_Idle;
```
jaylogue: I expected a call to HandleRequestDone() here.  Without one, the EC will be leaked.

src/lib/profiles/security/WeaveCertProvisioning.cpp
```diff
@@ -0,0 +1,767 @@
...
+    }
+
+    // Process GetCertificateResponse message from the CA service.
+    else if (profileId == kWeaveProfile_Security && msgType == kMsgType_GetCertificateResponse)
+    {
+        // If the request is now complete, clear the request state and enter Idle state.
+        client->HandleRequestDone();
+
+        // Process message and deliver result to the application using API event callback function.
+        client->ProcessGetCertificateResponse(msgBuf);
```
jaylogue: Maybe we should add a comment explaining why we ignore the error return error.  Something like:

_Note that, in this context, any error returned by ProcessGetCertificateResponse() has already been seen by the application.  Thus we can safely ignore it here._

src/lib/profiles/security/WeaveCertProvisioning.cpp
emargolis: done

src/lib/profiles/security/WeaveCertProvisioning.cpp
emargolis: done

src/lib/profiles/security/WeaveCertProvisioning.cpp
emargolis: done

src/lib/profiles/security/WeaveSecurity.h
```diff
@@ -243,6 +252,45 @@ enum
     kTag_GroupKeySignature_KeyId                = 2,    //  [ unsigned int ] Weave KeyId to be used to generate and verify the signature
     kTag_GroupKeySignature_Signature            = 3,    //  [ byte string ] Signature bytes themselves.
 
+    // ---- Context-specific Tags for GetCertificateRequest Message Structure ----
+    kTag_GetCertReqMsg_ReqType                  = 1,    // [ unsigned int ] Identifies the certificate request type.
+    kTag_GetCertReqMsg_Authorize_PairingToken   = 2,    // [ byte string, optional ] Pairing token from the service, in BER integer encoding.
```
jaylogue: This is not BER.  From the point of view of the device and the protocol, this is an opaque byte string.

src/lib/profiles/security/WeaveSecurity.h
```diff
@@ -243,6 +252,45 @@ enum
...
+                                                        //                           This is used to authorize GetCertificate to the service.
+    kTag_GetCertReqMsg_Authorize_PairingInitData= 3,    // [ byte string, optional ] Pairing initialization data from the service, in BER integer encoding.
```
jaylogue: This is not BER.  From the point of view of the device and the protocol, this is an opaque byte string.

src/lib/profiles/security/WeaveSecurity.h
jaylogue: The indention used in the comments does not match the reset of the value, and at least to my eyes, is harder to read.

src/lib/profiles/security/WeaveSecurity.h
```diff
@@ -243,6 +252,45 @@ enum
...
+    kTag_GetCertReqMsg_MfrAttest_WeaveCert      = 6,    // [ structure ] Weave device certificate provisioned at the factory at manufacturing time.
+                                                        //               This has the same internal structure as an kTag_WeaveCertificate.
+    kTag_GetCertReqMsg_MfrAttest_WeaveRelCerts  = 7,    // [ array, optional ] An optional array of intermediate Weave certificates, which are needed to
+                                                        //                     validate included Weave device certificate. May be omitted if validators
+                                                        //                     are expected to have the necessary certificates for validation.
+    kTag_GetCertReqMsg_MfrAttest_X509Cert       = 8,    // [ byte string ] X509 device certificate provisioned at the factory at manufacturing time.
+                                                        //                 This is an ASN1 encoded ECDSA or RSA certificate TLV-encoded as data blob.
+    kTag_GetCertReqMsg_MfrAttest_X509RelCerts   = 9,    // [ array, optional ] An optional array of intermediate X509 certificates, which are needed to
```
jaylogue: The comment should describe the type of the array's elements.

src/lib/profiles/security/WeaveSecurity.h
```diff
@@ -243,6 +252,45 @@ enum
...
+    kTag_GetCertReqMsg_OpDeviceCert             = 4,    // [ structure ] Weave operational device certificate.
+                                                        //               This has the same internal structure as an kTag_WeaveCertificate.
+    kTag_GetCertReqMsg_OpRelatedCerts           = 5,    // [ array, optional ] An optional array of intermediate Weave certificates, which are needed to
+                                                        //                     validate Weave operational device certificate. May be omitted if validators
+                                                        //                     are expected to have the necessary certificates for validation.
+    kTag_GetCertReqMsg_MfrAttest_WeaveCert      = 6,    // [ structure ] Weave device certificate provisioned at the factory at manufacturing time.
+                                                        //               This has the same internal structure as an kTag_WeaveCertificate.
+    kTag_GetCertReqMsg_MfrAttest_WeaveRelCerts  = 7,    // [ array, optional ] An optional array of intermediate Weave certificates, which are needed to
```
jaylogue: The comment should describe the type of the array's elements.

emargolis: updated

src/lib/profiles/security/WeaveSecurity.h
emargolis: updated

src/lib/profiles/security/WeaveSecurity.h
emargolis: updated

src/lib/profiles/security/WeaveSecurity.h
emargolis: updated

src/lib/profiles/security/WeaveSecurity.h
emargolis: updated

jaylogue:
Is it safe to delete the branch?

robszewczyk:
> Is it safe to delete the branch?

Absolutely.  Branch deleted.

PR #458, opened 2019-12-18
Update BUILDING with required Python packages

Vyrastas:
These packages were added to openweave.io/guides/build, so adding here too to keep in sync.

Also fixed some code-font formatting issues.


robszewczyk:
Thank you @Vyrastas !

PR #460, opened 2019-12-19
CommandSender: support for custom commands

holbrookt:
This is a re-opening of a PR started by johnsj@google.com. The original PR description is:

This PR adds a new CommandSender class that encapsulates the protocol aspects of sending out a command request, simplifying dealing with command requests for the application layer.

The class has a number of modalities that an application might find
useful including:

Initializing the object once for use with a peer with subsequent
command transmissions just altering the payload

Variant on 1, where the application desires to send out commands
in quick succession, implicitly indicating on each subsequent transmission
that it no longer cares about the previous command transmission.

Initializing the object once to send out a particular command, then
altering the peer + payload to send it to by passing in a different binding
each time to SendCommand. This is particular useful for OneWay
commands.


src/lib/profiles/data-management/Current/Command.cpp
```diff
@@ -37,13 +37,14 @@
...
 
-#if WEAVE_CONFIG_ENABLE_RELIABLE_MESSAGING && WDM_PUBLISHER_ENABLE_CUSTOM_COMMANDS
+#if WEAVE_CONFIG_ENABLE_RELIABLE_MESSAGING && WDM_PUBLISHER_ENABLE_CUSTOM_COMMAND_HANDLER
 
 namespace nl {
 namespace Weave {
 namespace Profiles {
 namespace WeaveMakeManagedNamespaceIdentifier(DataManagement, kWeaveManagedNamespaceDesignation_Current) {
 
+using namespace nl::Weave;
```
holbrookt: From previous PR (@robszewczyk):

Given the editing changes elsewhere, please remove the using decaration

src/lib/core/WeaveDMConfig.h
```diff
@@ -467,23 +467,36 @@
...
+ *  @def WEAVE_CONFIG_ENABLE_WDM_CUSTOM_COMMAND_SENDER
+ *
+ *  @brief
+ *    Enable (1) or disable (0) command sender support
+ *    in Weave Data Management Next profile. This feature is
+ *    optional and could be disabled.
+ *
+ */
+#ifndef WEAVE_CONFIG_ENABLE_WDM_CUSTOM_COMMAND_SENDER
+#define WEAVE_CONFIG_ENABLE_WDM_CUSTOM_COMMAND_SENDER 0
```
holbrookt: From previous PR (@jaylogue):

Does this need to be off by default? I presume it won't cause any extra code to be linked if no one references it.

src/lib/profiles/data-management/Current/CommandSender.cpp
```diff
@@ -0,0 +1,537 @@
...
+        mEventCallback = aEventCallback;
+        mAppState = aAppState;
+
+        inParam.Clear();
+        outParam.Clear();
+
+        // Test the application to ensure it's calling the default handler appropriately.
+        aEventCallback(aAppState, kEvent_DefaultCheck, inParam, outParam);
+        VerifyOrExit(outParam.defaultHandlerCalled, err = WEAVE_ERROR_DEFAULT_EVENT_HANDLER_NOT_CALLED);
+    }
```
holbrookt: From previous PR (@jaylogue):

Wrap the above in #if DEBUG.

holbrookt: Do you mean lines 170 and 171?

src/lib/profiles/data-management/Current/CommandSender.cpp
```diff
@@ -0,0 +1,537 @@
...
+WEAVE_ERROR CommandSender::Init(Binding *aBinding, const EventCallback aEventCallback, void * const aAppState)
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+
+    VerifyOrExit(aBinding != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
+
+    mBinding = aBinding;
+    mBinding->AddRef();
+
+    if (aEventCallback)
```
holbrookt: from previous PR (@jaylogue):

It should be an error for event callback to be NULL.

holbrookt: @mrjerryjohns 

What was intended here? Is the DefaultEventHandler supposed to be used if EventCallback is NULL? Or is that a separate function?

src/lib/profiles/data-management/Current/Command.cpp
holbrookt: Rob, there are many other cpp files with "using namespace" - why are you requesting that it be removed here?

src/lib/core/WeaveDMConfig.h
holbrookt: @mrjerryjohns Is this weave config even necessary?

src/lib/profiles/data-management/Current/CommandSender.h
```diff
@@ -0,0 +1,261 @@
...
+
+        TraitDataSink *Sink;
+
+        ResourceIdentifier ResourceId;
+        uint32_t ProfileId;
+        SchemaVersionRange VersionRange;
+        uint64_t InstanceId;
+        uint32_t CommandType;
+
+        uint8_t Flags;
```
holbrookt: @mrjerryjohns it might be good to provide some explanation of these fields. Which ones are required to set?

src/lib/core/WeaveDMConfig.h
mrjerryjohns: Agreed, I think it's safe to enable it by default given the natural dead-stripping that'll happen if there are no references to it.

src/lib/profiles/data-management/Current/CommandSender.h
mrjerryjohns: Good idea.

ResourceId, ProfileId, CommandType should be populated. InstanceId, Flags and VersionRange default to 0 if unspecified.

Perhaps having two constructors, one with each argument explicitly specified, and another using the TraitCatalog approach to discover some of those parameters (i.e rename the PopulateTraitPath to be another overload constructor) might be a good idea.

@robszewczyk thoughts?

src/lib/profiles/data-management/Current/CommandSender.h
```diff
@@ -0,0 +1,261 @@
...
+ *        This class also helps applications infer if the data within an associated TraitDataSink has caught up to the side-effects
+ *        of the commmand (based on the version provided in the command response). The application is responsible for managing the
+ *        storage of that object.
+ *
+ * ## Weave Binding
+ *
+ * An object of this class can be intialized with a Weave Binding, which will serve as the default
+ * Binding to use to send Commands. The user may also provide a Binding to each call of
+ * SendCommand(), which will override the default Binding. It is not necessary to provide a default
+ * Binding, however any binding provided to CommandSender must already be initialized.
```
holbrookt: @mrjerryjohns @robszewczyk Is it acceptable to require that the user has already initialized the binding?

holbrookt: @mrjerryjohns @robszewczyk Is it acceptable to require that the user has already initialized the binding?

src/lib/profiles/data-management/Current/CommandSender.h
jaylogue: Given that this structure is part of the public API, all fields should be documented, including describing when and how they should be set.

As far as I can see, this structure is only used as a means to avoid having a long list of arguments to SendCommand().  As such, I wouldn't expect it to have a constructor.

src/lib/profiles/data-management/Current/CommandSender.h
jaylogue: By 'initialized" do you mean prepared?

The standard pattern we use is to have the protocol object (CommandSender) detect when the binding isn't prepared and raise a PrepareRequested event to the binding user.

holbrookt: Yes I did mean prepared. I can implement that pattern here as well if necessary

holbrookt: After talking with Jay, we've decided that this can and will be added in a later PR. For now the user will be responsible for preparing the Binding

PR #462, opened 2020-01-10
Refactor TraitCatalog to explicitly instantiate SingleResourceCatalogs

robszewczyk:
We break out the instantiation of the SingleResourceCatalog<>
implementation into a separate TraitCatalog.ipp file, and explicitly
instantiate the SingleResourceCatalog<> for TraitDataSink and
TraitDataSource.  This refactoring results in faster compilation
times, and cleans up binary dependencies.


jaylogue:
@robszewczyk : What do you think about moving SingleResourceTraitCatalog into its own .h, .ipp, .cpp files, and then including SingleResourceTraitCatalog.h in TraitCatalog.h for compatibility?  That way TraitCatalog.h consists of purely generic definitions.

robszewczyk:
> @robszewczyk : What do you think about moving SingleResourceTraitCatalog into its own .h, .ipp, .cpp files, and then including SingleResourceTraitCatalog.h in TraitCatalog.h for compatibility? That way TraitCatalog.h consists of purely generic definitions.

Sounds good.  I will submit the revised PR.

src/lib/profiles/data-management/Current/TraitCatalog.h
```diff
@@ -194,296 +194,6 @@ class SingleResourceTraitCatalog : public TraitCatalogBase<T>
 typedef SingleResourceTraitCatalog<TraitDataSink> SingleResourceSinkTraitCatalog;
 typedef SingleResourceTraitCatalog<TraitDataSource> SingleResourceSourceTraitCatalog;
```
jaylogue: Might want to place extern template instantiations here to make sure the compiler generate doesn't any code for these types when TraitCatalog.h is included.  E.g.:

```
extern template class SingleResourceTraitCatalog<TraitDataSink>;
extern template class SingleResourceTraitCatalog<TraitDataSource>;
```

robszewczyk:
> > @robszewczyk : What do you think about moving SingleResourceTraitCatalog into its own .h, .ipp, .cpp files, and then including SingleResourceTraitCatalog.h in TraitCatalog.h for compatibility? That way TraitCatalog.h consists of purely generic definitions.
> 
> Sounds good. I will submit the revised PR.

Done.

src/lib/profiles/data-management/Current/TraitCatalog.h
robszewczyk: It would a bit of a belt-and-suspenders approach here? The way the template is split up right now we should be achieving the same behavior.  The above approach would ensure that if `SingleResourceTraitCatalog.h`was subsequently altered to include definitions of actual functions/linking symbols, those would not be compiled here, but we already stipulate such behavior via `.h`/`.ipp` split.

robszewczyk: I think I will merge this as is -- I find it hard to think about the interactions of `typedef`, template instantiation, and `extern`.  We can clean up in a subsequent PR

PR #463, opened 2020-01-10
Make the software update timer oblivious to service connectivity events

robszewczyk:
The existing behavior reset the software timers based on connectivity
to the service: when connectivity to the service was lost, the exiting
timer was cancelled, and when connectivity to the service was
restored, the idle timer was restored.  The intended effect was to
avoid software update checks when they were doomed to fail (service
unreachable) while preventing "thundering herd" behavior on service
outages.  However, the actual most common behavior was to reset the
long-running timer on any flapping in Thread network connectivity.

The solution is to just use the software update timer policy,
oblivious to the service connectivity.  The new code fails eagerly in
`PrepareBinding` when there is no service connectivity, which
generates no useless network activity.  The platform events are only
used to track whether we have a chance to succeed in contacting the
service.  Existing behavior already addresses the group behavior in
the event of mass service recovery -- the timer policy implements a
long, randomized background check, followed by randomized, exponential
backoff retry that ultimately reverts to the background rate.  In the
event of a synchronized start of a population of devices -- e.g. power
outage for line powered devices -- existing behavior also ensures
protection againts "thundering herds".

### DIFF
--- a/src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl.ipp
+++ b/src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl.ipp
@@ -62,7 +62,6 @@ void GenericSoftwareUpdateManagerImpl<ImplClass>::DoInit()
 {
     mShouldRetry = false;
     mScheduledCheckEnabled = false;
-    mHaveServiceConnectivity = false;
     mIgnorePartialImage = false;
 
     mEventHandlerCallback = NULL;
@@ -121,6 +120,12 @@ void GenericSoftwareUpdateManagerImpl<ImplClass>::PrepareBinding(intptr_t arg)
     GenericSoftwareUpdateManagerImpl<ImplClass> * self = &SoftwareUpdateMgrImpl();
 
     self->Cleanup();
+
+    if (!ConnectivityMgr().HaveServiceConnectivity())
+    {
+        WeaveLogProgress(DeviceLayer, "Software Update Check: No service connectivity");
+        ExitNow(err = WEAVE_ERROR_NOT_CONNECTED);
+    }
 
     self->mBinding = ExchangeMgr.NewBinding(HandleServiceBindingEvent, NULL);
     VerifyOrExit(self->mBinding != NULL, err = WEAVE_ERROR_NO_MEMORY);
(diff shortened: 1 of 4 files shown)


src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl.ipp
```diff
@@ -122,6 +122,8 @@ void GenericSoftwareUpdateManagerImpl<ImplClass>::PrepareBinding(intptr_t arg)
 
     self->Cleanup();
 
+    VerifyOrExit(self->mHaveServiceConnectivity, err = WEAVE_ERROR_INCORRECT_STATE);
```
jaylogue: The previous code generated a log message specifically noting the lack of service connectivity.  I think that would be useful to preserve, to help developers make sense of an "Incorrect state" error in this case.

I could also imagine using the WEAVE_ERROR_NOT_CONNECTED error in this case.

src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl.ipp
```diff
@@ -441,20 +443,17 @@ void GenericSoftwareUpdateManagerImpl<ImplClass>::_OnPlatformEvent(const WeaveDe
 {
     GenericSoftwareUpdateManagerImpl<ImplClass> * self = &SoftwareUpdateMgrImpl();
 
-     if (event->Type == DeviceEventType::kServiceConnectivityChange)
-     {
+    if (event->Type == DeviceEventType::kServiceConnectivityChange)
```
jaylogue: The ConnectivityManager provides a HaveServiceConnectivity() method.  So I think we can eliminate all the event handling logic here, along with mHaveServiceConnectivity.

src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl.ipp
robszewczyk: I thought about using `WEAVE_ERROR_NOT_CONNECTED`, I ultimately steered away from it because its use was around objects with actual connections.  But it does fit better here, so I'm changing the error and adding back the log message.   Thanks for the suggestion!

src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl.ipp
robszewczyk: Done.  Thanks for the suggestion.   I perhaps cut a bit more than initially intended -- I eliminated all references to `GenericSoftwareUpdateManagerImpl<>::_OnPlatformEvent`.  This eliminates it from the pattern of having the PlatformManager forward calls to specific managers on platform events, but it is clear how to recreate such a pattern when the need arises again.

src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl.ipp
robszewczyk: Done

src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericSoftwareUpdateManagerImpl.ipp
jaylogue: Agreed.  I think its cleaner not to forward OnPlatformEvent() if it's not really needed.

PR #466, opened 2020-01-14
Fixup Python3 support in Device Manager /WoBLE code on Linux

robszewczyk:
Bluez code contained a number of issues that were not caught earlier:

* `dbus.String` is essentially a string in Python3 but must be encoded
  in Python 2 if we want to use it with any modules that expect a
  string

* binascii.hexlify() returns a binary string. As such, it should
  always be decoded

* Better handling bytearrays -- avoid unnecessary conversions to
  string.


wijnandatgoogle:
All looks good to me.

PR #467, opened 2020-01-15
Device Layer: Added Operational Device Credentials Change Event

emargolis:
Added Operational Device Credentials Change Event to the Weave Device Layer.


jaylogue:
Where is this new event supposed to be generated?

PR #469, opened 2020-01-24
Weave time stray const

samhansen:
The `{ NULL } -> { }` is a bit of a RFC.  It resolves -Wmissing-field-initializers warnings, and I think it results in default-0 values in most C and C++ compilers.  RFC?

### DIFF
--- a/src/lib/profiles/time/WeaveTime.h
+++ b/src/lib/profiles/time/WeaveTime.h
@@ -1245,7 +1245,7 @@ protected:
     WEAVE_ERROR SendSyncRequest(void);
 
     void SetClientState(const ClientState state);
-    const char * const GetClientStateName(void) const;
+    const char * GetClientStateName(void) const;
 
     static void HandleTimeChangeNotification(ExchangeContext *aEC, const IPPacketInfo *aPktInfo,
         const WeaveMessageInfo *aMsgInfo, uint32_t aProfileId, uint8_t aMsgType, PacketBuffer *aPayload);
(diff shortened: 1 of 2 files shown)


gerickson:
> The `{ NULL } -> { }` is a bit of a RFC. It resolves -Wmissing-field-initializers warnings, and I think it results in default-0 values in most C and C++ compilers. RFC?

I'd rather see an explicit value there (a mnemonic, if need be). If NULL isn't appropriate, then let's figure out one that is and, if necessary, create a mnemonic for it.

Regardless, if the NULL change isn't germane to the const change, then I'd remove this from the PR and create another one for that as PRs should be well-scoped and largely self-contained.

robszewczyk:
> > The `{ NULL } -> { }` is a bit of a RFC. It resolves -Wmissing-field-initializers warnings, and I think it results in default-0 values in most C and C++ compilers. RFC?
> 
> I'd rather see an explicit value there (a mnemonic, if need be). If NULL isn't appropriate, then let's figure out one that is and, if necessary, create a mnemonic for it.
> 
> Regardless, if the NULL change isn't germane to the const change, then I'd remove this from the PR and create another one for that as PRs should be well-scoped and largely self-contained.

Agreed with both.  Perhaps defining a `const OptionDef OptionDefEndOfList` or equivalent in nlargparser.hpp would be the way to go?

jaylogue:
Does this compile with c++03?

samhansen:
@jaylogue assuming you are asking about the empty struct initialization, yes via `-std=c++03` with both g++ and clang.  The particular compiler versions I checked are:

```
g++ (Debian 8.3.0-6) 8.3.0
clang version 8.0.1-3+build1 (tags/RELEASE_801/final)
```

samhansen:
Thanks, I'm moving the struct initialization to its own PR (including terminal sentinel).  PTAL just for the const qualifier commit.

PR #470, opened 2020-01-24
Assume OptionDefEndOfList terminal sentinel for array defs

samhansen:
This adds an EndOfList sentinel for the OptionDef arrays.

### DIFF
--- a/src/lib/support/nlargparser.cpp
+++ b/src/lib/support/nlargparser.cpp
@@ -1082,7 +1082,7 @@ HelpOptions::HelpOptions(const char *appName, const char *appUsage, const char *
     {
         { "help",      kNoArgument, 'h' },
         { "version",   kNoArgument, 'v' },
-        { NULL }
+        { }
     };
     OptionDefs = optionDefs;
 
--- a/src/test-apps/CASEOptions.cpp
+++ b/src/test-apps/CASEOptions.cpp
@@ -152,7 +152,7 @@ CASEOptions::CASEOptions()
         { "case-use-known-key",     kNoArgument,            kToolCommonOpt_CASEUseKnownECDHKey  },
 #endif // WEAVE_CONFIG_SECURITY_TEST_MODE
 #endif // WEAVE_CONFIG_ENABLE_CASE_INITIATOR || WEAVE_CONFIG_ENABLE_CASE_RESPONDER
-        { NULL }
+        { }
     };
     OptionDefs = optionDefs;
 
--- a/src/test-apps/DeviceDescOptions.cpp
+++ b/src/test-apps/DeviceDescOptions.cpp
@@ -44,7 +44,7 @@ DeviceDescOptions::DeviceDescOptions()
         { "product-id",       kArgumentRequired, kToolCommonOpt_DeviceProductId        },
         { "product-rev",      kArgumentRequired, kToolCommonOpt_DeviceProductRevision  },
         { "software-version", kArgumentRequired, kToolCommonOpt_DeviceSoftwareVersion  },
-        { NULL }
+        { }
     };
     OptionDefs = optionDefs;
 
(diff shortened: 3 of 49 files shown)


jaylogue:
Why do we want this?

src/lib/support/nlargparser.hpp
```diff
@@ -70,6 +70,11 @@ struct OptionDef
                                              for the option. */
 };
 
+/**
+ * Defines a terminal sentinel for a sequence of OptionDef structs.
+ */
+const OptionDef OptionDefEndOfList = { nullptr, kNoArgument, 0 };
```
gerickson: Minor nit, but I might suggest calling this "kOptionDefEndOfList" to indicate the constant convention.

samhansen:
@jaylogue it resolves a build error resulting from -Werror=missing-field-initializers assuming -Wmissing-field-initializers is in effect.

src/lib/support/nlargparser.hpp
samhansen: Done, thanks :)

jaylogue:
@samhansen Why are we trying to build with those warnings on?  In general, we don't write code that avoids all possible compiler warnings.

Additionally, having a special "end-of" object just for OptionDef is fussy and not a pattern I'm likely to adhere to.

jaylogue:
@gerickson I know you opined otherwise, but I find having a type-specific end-of-list marker to be fussy and just another special thing I have to remember / explain to other people.  { } is C++ cannon, and probably the most straightforward replacement for the (unfairly) maligned { NULL } pattern.

gerickson:
> @gerickson I know you opined otherwise, but I find having a type-specific end-of-list marker to be fussy and just another special thing I have to remember / explain to other people. { } is C++ cannon, and probably the most straightforward replacement for the (unfairly) maligned { NULL } pattern.

My preference is we pick a pattern and be consistent about it.

So, if {} is cannon and it works for all EOL use cases in our code base for the compilers and environments we support, then we should be consistent about it and use that. If not and if NULL, in this case, was necessary but didn't work because of casting or whatnot, then a type-specific EOL marker is what we should use and be consistent with. I would absolutely want to avoid:

```
#ifdef SOME_COMPILER_A
    { (some type cast)NULL }
#else
    { NULL }
#endif
```

or:

```
    { (some type cast)NULL }
```

in one place and:

```
   { (some other type cast)NULL }
```

in another place. But at the end of the day, if NULL worked/works, then we should leave it alone.

Bottom line, my suggestion was a preference towards "If we have to change this, here's how I'd change it" (particularly if the default EOL values are / were non-zero).

jaylogue:
EOL markers always depend on the nature of the list, so it can't be truly universal.  { NULL } is perfectly legal for structures where the first field is a pointer (or integer) and NULL is the "end" marker. However it triggers some pedantic warnings about possibly missing field initializers.  { } is the definitive way to say "I intend for this structure to be initialized to all zeros".

gerickson:
> EOL markers always depend on the nature of the list, so it can't be truly universal. { NULL } is perfectly legal for structures where the first field is a pointer (or integer) and NULL is the "end" marker. However it triggers some pedantic warnings about possibly missing field initializers. { } is the definitive way to say "I intend for this structure to be initialized to all zeros".

Absolutely understood and agreed.

I am simply driving at that if {} or { NULL } fit all of our use cases, then those are the right pattern to follow and this PR is unneeded. But, if we start getting into "this list needs { NULL, 0, kFoo }" and "that list needs { 0, kBar, kFoo, NULL }" and "that other list needs { 1, NULL, kFoobar }" whether for compilers or not, then we start to fall away from a consistent pattern and I'd argue for making things consistent again with named, typed, EOL markers to handle that divergence.

jaylogue:
@gerickson Understood.

@samhansen Lets abandon this and go with { } as in the original PR.

samhansen:
> @samhansen Lets abandon this and go with { } as in the original PR.

Done.  ptal

PR #471, opened 2020-01-24
Python3 compatibility for DeviceManager

robszewczyk:
A few missing bits that made it past the previous changes:
* _BytesArrayToVoidPtr operates on `bytes` class rather than on `str`
class.  In Python 2, `bytes` is merely an alias for `str` so no
changes required, in Python3, `bytes` is the binary string returned by
various utilities such as `base64.b64decode()`
* Wheel packaging changed to reflect that the package is expected to
work on Python3.

### DIFF
--- a/src/device-manager/python/openweave/WeaveDeviceMgr.py
+++ b/src/device-manager/python/openweave/WeaveDeviceMgr.py
@@ -100,7 +100,7 @@ def _VoidPtrToByteArray(ptr, len):
 
 def _ByteArrayToVoidPtr(array):
     if array != None:
-        if not (isinstance(array, str) or isinstance(array, bytearray)):
+        if not (isinstance(array, bytes) or isinstance(array, bytearray)):
             raise TypeError("Array must be an str or a bytearray")
         return cast( (c_byte * len(array)) .from_buffer_copy(array), c_void_p)
     else:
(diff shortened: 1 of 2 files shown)


src/device-manager/python/openweave/WeaveDeviceMgr.py
```diff
@@ -1017,6 +1020,12 @@ def RegisterServicePairAccount(self, serviceId, accountId, serviceConfig, pairin
         if accountId is not None and '\x00' in accountId:
             raise ValueError("Unexpected NUL character in accountId")
 
+        if pairingToken is not None and isinstance(pairingToken, str):
```
robszewczyk: Reviewers: please note that this is likely an artifact of a bug in the API: `weave-device-manager` calls `RegisterServicePairAccount` with an expectation to pass in a `str` as a type to `pairingToken` and `pairingInitData`.  This is inconsistent with the convention elsewhere in the code, I'd be happy to change the API convention instead.

jaylogue: As discussed off-line, I think its suitable for the pairingToken and pairingInitData arguments to accept both string and bytes, as this give the API user the most flexibility.

Issue #473, opened 2020-01-29
Device Layer: Clang build issues


jasongraffius:
When building the Device Layer with clang, a few idiosyncrasies of clang cropped up that should probably be rectified.

Specifically, clang seems to have an issue with friend declarations of the form `friend Type ::fully::qualified::function::name(...);`, at least in the specific situations encountered in some of the Device Layer headers. One such example is in [PlatformManager.h](https://github.com/openweave/openweave-core/blob/ded21fc283ead61312bd276cf9558e7fe104c93b/src/adaptations/device-layer/include/Weave/DeviceLayer/PlatformManager.h#L93-L96):

```c++
    friend ::nl::Weave::System::Error ::nl::Weave::System::Platform::Layer::PostEvent(::nl::Weave::System::Layer & aLayer, void * aContext, ::nl::Weave::System::Object & aTarget, ::nl::Weave::System::EventType aType, uintptr_t aArgument);
    friend ::nl::Weave::System::Error ::nl::Weave::System::Platform::Layer::DispatchEvents(::nl::Weave::System::Layer & aLayer, void * aContext);
    friend ::nl::Weave::System::Error ::nl::Weave::System::Platform::Layer::DispatchEvent(::nl::Weave::System::Layer & aLayer, void * aContext, ::nl::Weave::System::Event aEvent);
    friend ::nl::Weave::System::Error ::nl::Weave::System::Platform::Layer::StartTimer(::nl::Weave::System::Layer & aLayer, void * aContext, uint32_t aMilliseconds);
```

The proposed fix for this is to delimit the fully qualified name of the function by surrounding it with parentheses:

```c++
    friend ::nl::Weave::System::Error (::nl::Weave::System::Platform::Layer::PostEvent(::nl::Weave::System::Layer & aLayer, void * aContext, ::nl::Weave::System::Object & aTarget, ::nl::Weave::System::EventType aType, uintptr_t aArgument));
    friend ::nl::Weave::System::Error (::nl::Weave::System::Platform::Layer::DispatchEvents(::nl::Weave::System::Layer & aLayer, void * aContext));
    friend ::nl::Weave::System::Error (::nl::Weave::System::Platform::Layer::DispatchEvent(::nl::Weave::System::Layer & aLayer, void * aContext, ::nl::Weave::System::Event aEvent));
[...]

PR #474, opened 2020-01-29
Fixes for clang build in Device Layer

jasongraffius:
Apply fixes proposed in #473

### DIFF
--- a/src/adaptations/device-layer/include/Weave/DeviceLayer/PlatformManager.h
+++ b/src/adaptations/device-layer/include/Weave/DeviceLayer/PlatformManager.h
@@ -90,10 +90,11 @@ private:
     template<class> friend class Internal::GenericThreadStackManagerImpl_OpenThread;
     template<class> friend class Internal::GenericThreadStackManagerImpl_OpenThread_LwIP;
     template<class> friend class Internal::GenericConfigurationManagerImpl;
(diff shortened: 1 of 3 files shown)


src/adaptations/device-layer/include/Weave/DeviceLayer/PlatformManager.h
```diff
@@ -28,6 +28,19 @@
 
 namespace nl {
 namespace Weave {
+namespace System {
```
robszewczyk: I believe this be done via appropriate include of the authoritative header file

jasongraffius: I agree, however, including the authoritative header file (SystemLayer.h) only defines these functions when `WEAVE_SYSTEM_CONFIG_USE_LWIP` is asserted, so when that is not asserted, the build will fail.

jaylogue: Upon reflection, I think this is the right location for these forward declarations, since the PlatformManager class chooses to friend these functions regardless of whether they're actually defined/used.

robszewczyk: Agreed.  I was going to argue that "what if those functions change?  Now we have to change this header as well".  but that's a wrong argument -- we would already have to change this header because the friend declaration repeats the declaration.  With the proposed change, changing the declarations in `SystemLayer.h` would not result in a compilation failure but rather in a link failure subsequently.

PR #475, opened 2020-01-29
move StoreDataElement to private in TraitUpdatableDataSource

yunhanw-google:
-- StoreDataElement in TraitUpdatableDataSource is only invoked by
SubscritpionEngine when processing update request, and
SubscritpionEngine is friend of TraitUpdatableDataSource, therefore
 move StoreDataElement to private.

### DIFF
--- a/src/lib/profiles/data-management/Current/TraitData.h
+++ b/src/lib/profiles/data-management/Current/TraitData.h
@@ -1058,18 +1058,6 @@ public:
         } mDictionaryItemDelete;
     };
 
-    /**
-     * Given a reader that points to a data element conformant to a schema bound to this object, this method processes that data and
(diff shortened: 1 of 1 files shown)


robszewczyk:
In general, it seems like a good policy to minimize the API surface.  In this instance, there is an additional complication that the equivalent function -- `StoreDataElement` exists in the `TraitDataSink` as a public method.  In both cases, the method is called under identical circumstances -- it is meant to be called by the engine only.  I would suggest we preserve such parallelism: either decline this PR and keep things as they are, or make both `StoreDataElement` methods private and add explicitly add the allowed callers via friend classes.  If we opt to pursue the second route, the friend list would need to be expanded to include unit test code.  

Leaving this without approval or change request, as this is meant to open a discussion.

PR #477, opened 2020-01-31
add kStatus_InvalidValueInUpdate

yunhanw-google:
-- add kStatus_InvalidValueInUpdate to represent data element value error in wdm
profie.
-- when processing update request, we need to skip this DE when DE generates
WEAVE_ERROR_WRONG_TLV_TYPE.


jaylogue:
This error case represents a client's failure to properly implement the trait schema.  As such, I don't think a special status code is needed.  The publisher can simply send Common:BadRequest.

yunhanw-google:
> This error case represents a client's failure to properly implement the trait schema. As such, I don't think a special status code is needed. The publisher can simply send Common:BadRequest.

I see, thanks!

Issue #399, opened 2019-10-24
Add support for wireless regulatory provisioning


jaylogue:
**_PROPOSED: Extend the Weave Network Provisioning profile with features to support runtime provisioning of wireless regulatory configuration on devices._**  

To support wireless regulatory provisioning, it is proposed to extend the Network Provisioning profile with three new message types, comprising two new protocol patterns.

### SetWirelessRegulatoryConfig (msg type 15)

The SetWirelessRegulatoryConfig message updates a device with new wireless regulatory configuration information.  Information conveyed with the message describes the operating location of the device with respect to regulations governing wireless radio transmissions.  The payload of a SetWirelessRegulatoryConfig message consists of the following TLV structure:

```
SetWirelessRegulatoryConfigArguments [anon] => STRUCTURE
{
    /** ISO 3166-1 country code, or “00” for world-wide.
     */
    regulatoryDomain [1, optional] : STRING [len 2],

    /** Operating location relevant to wireless regulatory rules.
     */
    operatingLocation [2, optional] : UNSIGNED INTEGER [range 0..255]
    {
        reserved = 0,  /* may not be sent. */  
        unknown = 1,
        indoors = 2,
        outdoors = 3
    }
}
```

The regulatoryDomain argument must identify the country in which the device is or will be operated, OR another country that is subject to the same or compatible regulatory rules as the operating country.  Additionally, the special value “00” may be supplied to indicate the device should operate in a mode which is compatible with all regulatory rules world-wide.

The operatingLocation argument identifies a logical location in which the device is or will be operated that may be relevant to the regulatory rules of the operating country.  This value is an integer enumeration that may be extended over time.  The `unknown` operating location signifies that the device's specific operating location is not known, or that it may change over time.  (Note that the `reserved` value _may not_ be sent in any message).

The values specified for regulatoryDomain and operatingLocation govern the transmission behavior of *all* radios built-in to the device.  This includes WiFi and 802.15.4 radios, if so provisioned.  The effects of setting wireless regulatory configuration should be persistent on the device at least until the device is factory reset.

Either or both the regulatoryDomain or operationLocation arguments may be absent in a SetWirelessRegulatoryConfig message. If an argument is not present, then its corresponding current value on the target device is not changed.

[...]

PR #478, opened 2020-02-03
Network Provisioning: Support for wireless regulatory provisioning

jaylogue:
This change implements the feature described in https://github.com/openweave/openweave-core/issues/399.

- Added three new message types to the Network Provisioning profile:
GetWirelessRegulatoryConfig, GetWirelessRegulatoryConfigComplete and SetWirelessRegulatoryConfig.

- Extended the NetworkProvisioningServer and NetworkProvisioningDelegate classes to support the new provisioning protocol.

- Created WirelessRegConfig class and associated types to represent wireless regulatory config in memory and provide functions for encoding/decoding to Weave TLV.

- Added APIs to C++ WeaveDeviceManager class for getting and setting wireless regulatory configuration information. Extended the Java and Python (but not Objective-C++) wrappers for the Device Manager to support the APIs.

- Added support for wireless regulatory provisioning to the mock device and extended the WeaveDeviceManager happy test to exercise it.

- Extended the WeaveDeviceManager JNI tests to exercise the new Java APIs.

### DIFF
--- a/src/device-manager/WeaveDeviceManager.cpp
+++ b/src/device-manager/WeaveDeviceManager.cpp
@@ -1842,6 +1842,77 @@ exit:
     return err;
 }
 
+WEAVE_ERROR WeaveDeviceManager::GetWirelessRegulatoryConfig(void* appReqState, GetWirelessRegulatoryConfigCompleteFunct onComplete,
+        ErrorFunct onError)
+{
+    WEAVE_ERROR     err     = WEAVE_NO_ERROR;
+    PacketBuffer*   msgBuf  = NULL;
+
+    if (mOpState != kOpState_Idle)
+        return WEAVE_ERROR_INCORRECT_STATE;
+
+    VerifyOrExit(onComplete != NULL && onError != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
+
+    msgBuf = PacketBuffer::New();
+    VerifyOrExit(msgBuf != NULL, err = WEAVE_ERROR_NO_MEMORY);
+
+    mAppReqState = appReqState;
+    mOnComplete.GetWirelessRegulatoryConfig = onComplete;
+    mOnError = onError;
+    mOpState = kOpState_GetWirelessRegulatoryConfig;
+
+    err = SendRequest(kWeaveProfile_NetworkProvisioning, NetworkProvisioning::kMsgType_GetWirelessRegulatoryConfig, msgBuf,
+            HandleNetworkProvisioningResponse);
+    msgBuf = NULL;
+
+exit:
+    if (msgBuf != NULL)
+        PacketBuffer::Free(msgBuf);
+    if (err != WEAVE_NO_ERROR)
+        ClearOpState();
+    return err;
+}
+
+WEAVE_ERROR WeaveDeviceManager::SetWirelessRegulatoryConfig(const WirelessRegConfig *regConfig, void* appReqState,
+        CompleteFunct onComplete, ErrorFunct onError)
+{
+    WEAVE_ERROR     err     = WEAVE_NO_ERROR;
+    PacketBuffer*   msgBuf  = NULL;
+    TLVWriter       writer;
+
+    if (mOpState != kOpState_Idle)
+        return WEAVE_ERROR_INCORRECT_STATE;
+
+    VerifyOrExit(onComplete != NULL && onError != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
+
+    msgBuf = PacketBuffer::New();
+    VerifyOrExit(msgBuf != NULL, err = WEAVE_ERROR_NO_MEMORY);
+
+    writer.Init(msgBuf);
+
+    err = regConfig->Encode(writer);
+    SuccessOrExit(err);
+
+    err = writer.Finalize();
+    SuccessOrExit(err);
+
+    mAppReqState = appReqState;
+    mOnComplete.General = onComplete;
+    mOnError = onError;
+    mOpState = kOpState_SetWirelessRegulatoryConfig;
+
+    err = SendRequest(kWeaveProfile_NetworkProvisioning, NetworkProvisioning::kMsgType_SetWirelessRegulatoryConfig, msgBuf,
+            HandleNetworkProvisioningResponse);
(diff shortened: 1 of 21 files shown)


src/device-manager/WeaveDeviceManager.cpp
```diff
@@ -1839,6 +1839,77 @@ WEAVE_ERROR WeaveDeviceManager::SetRendezvousMode(uint16_t modeFlags, void* appR
     return err;
 }
 
+WEAVE_ERROR WeaveDeviceManager::GetWirelessRegulatoryConfig(void* appReqState, GetWirelessRegulatoryConfigCompleteFunct onComplete,
+        ErrorFunct onError)
+{
+    WEAVE_ERROR     err     = WEAVE_NO_ERROR;
+    PacketBuffer*   msgBuf  = NULL;
+
+    if (mOpState != kOpState_Idle)
```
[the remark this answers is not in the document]
jaylogue: No.  That would disrupt the current state of the device manager object. (Notice the code in exit that calls ClearOpState() if err != WEAVE_NO_ERROR).

This pattern of returning directly is essential when we need to inform the user that the method cannot be called in the current state but we don't the call to disrupt that state.

src/device-manager/WeaveDeviceManager.cpp
```diff
@@ -1839,6 +1839,77 @@ WEAVE_ERROR WeaveDeviceManager::SetRendezvousMode(uint16_t modeFlags, void* appR
...
+}
+
+WEAVE_ERROR WeaveDeviceManager::SetWirelessRegulatoryConfig(const WirelessRegConfig *regConfig, void* appReqState,
+        CompleteFunct onComplete, ErrorFunct onError)
+{
+    WEAVE_ERROR     err     = WEAVE_NO_ERROR;
+    PacketBuffer*   msgBuf  = NULL;
+    TLVWriter       writer;
+
+    if (mOpState != kOpState_Idle)
```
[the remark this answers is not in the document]
jaylogue: See above.

src/device-manager/WeaveDeviceManager.cpp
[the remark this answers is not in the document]
yunhanw-google: agree, thanks for the clarification.

src/device-manager/java/src/nl/Weave/DeviceManager/WirelessRegulatoryConfig.java
```diff
@@ -0,0 +1,53 @@
...
+package nl.Weave.DeviceManager;
+
+/**
+ * Container for wireless regulatory configuration information.
+ */
+public class WirelessRegulatoryConfig
+{
+    /** Active wireless regulatory domain.
+     */
+    public String RegDomain;
```
g-coppock: Can we refactor this to make it immutable? Something like this:

```
public final class WirelessRegulatoryConfig {
  private final String regulatoryDomain;
  private final WirelessOperatingLocation operatingLocation;
  private final List<String> supportedRegulatoryDomains;

  public WirelessRegulatoryConfig(
      String regulatoryDomain, 
      WirelessOperatingLocation operatingLocation, 
      List<String> supportedRegulatoryDomains) {
        this.regulatoryDomain = regulatoryDomain;
        this.operatingLocation = operatingLocation;
        this.supportedRegulatoryDomains = new ArrayList<>(supportedRegulatoryDomains);
  }

  public WirelessOperatingLocation getOperatingLocation() { 
    return operatingLocation;
  }

  public String getRegulatoryDomain() {
    return regulatoryDomain;
  }

  public List<String> getSupportedRegulatoryDomains() {
    return new ArrayList<>(supportedRegulatoryDomains);
  }
}
```

This also avoids the need for a factory method (`Make`) although you could have both. This guarantees that all fields are initialized, though, and cannot be modified.

jaylogue: Making the fields private requires considerable re-working of the JNI code.  I could add the getters and the constructor, and then make the fields package scope final.  I would still need Make(), but that could be package scope as well.  Thoughts?

g-coppock: Chatted offline, didn't realize this would require a larger refactor of the JNI code. Avoiding the getters and just having public final fields should be fine.

jaylogue: My latest commit makes the class immutable but preserves the existing public fields.

src/device-manager/java/src/nl/Weave/DeviceManager/WirelessRegulatoryConfig.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public final WirelessOperatingLocation OpLocation;
+    
+    /** Supported regulatory domains
+     */
+    public final String[] SupportedRegDomains;
+    
+    public WirelessRegulatoryConfig(String regDomain, WirelessOperatingLocation opLocation)
+    {
+		this.RegDomain = regDomain;
```
g-coppock: Fix indent to 4-spaces here and below?

src/device-manager/java/WeaveDeviceManager-JNI.cpp
```diff
@@ -3294,28 +3421,11 @@ WEAVE_ERROR N2J_NetworkInfo(JNIEnv *env, const NetworkInfo& inNetworkInfo, jobje
...
-        VerifyOrExit(!env->ExceptionCheck(), err = WDM_JNI_ERROR_EXCEPTION_THROWN);
-
-        env->DeleteLocalRef(elem);
-        elem = NULL;
-    }
-
-exit:
-    env->DeleteLocalRef(elem);
-    return err;
+    return N2J_GenericArray(env, inArrayLen, sNetworkInfoCls, outArray,
```
robszewczyk: Neat!  A clarification: Is the primary reason for `N2J_GenericArray` for being a template so that it may take both closures and functions?  Could it not work just with closures without being a template?

jaylogue: Discussed off-line

PR #481, opened 2020-02-05
[nrf5] Enable enhancement features of OpenThread

bukepo:
- Enable jam detection
- Enable child supervisoin

### DIFF
--- a/build/nrf5/nrf5-openthread.mk
+++ b/build/nrf5/nrf5-openthread.mk
@@ -161,7 +161,9 @@ OPENTHREAD_CONFIGURE_OPTIONS = \
     --srcdir="$(OPENTHREAD_ROOT)" \
     --enable-ftd \
     --enable-mtd \
+    --enable-child-supervision \
     --enable-diag \
+    --enable-jam-detector \
     --enable-linker-map \
     --with-examples=nrf52840 \
     --disable-tools \
--- a/build/nrf5/nrf5-openthread.mk
+++ b/build/nrf5/nrf5-openthread.mk
@@ -163,7 +163,7 @@ OPENTHREAD_CONFIGURE_OPTIONS = \
     --enable-mtd \
     --enable-child-supervision \
     --enable-diag \
-    --enable-jam-detector \
+    --enable-jam-detection \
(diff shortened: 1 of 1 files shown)


build/nrf5/nrf5-openthread.mk
```diff
@@ -161,7 +161,9 @@ OPENTHREAD_CONFIGURE_OPTIONS = \
     --srcdir="$(OPENTHREAD_ROOT)" \
     --enable-ftd \
     --enable-mtd \
+    --enable-child-supervision \
     --enable-diag \
+    --enable-jam-detector \
```
jwhui:
```suggestion
    --enable-jam-detection \
```

jaylogue:
Why do we want these on for all projects that use OpenWeave?  Are these features necessary/appropriate for all use cases?  What if a project doesn't want these features enabled?

bukepo:
@jaylogue I think these features are appropriate for the lock and border router. Since the current makefile assumes an old OpenThread commit, it doesn’t support enable/disable all features in OpenThreadConfig.h. So it seems we have to enable it here for now before we move to new OpenThread.

jaylogue:
@bukepo This makefile is potentially used by any product that uses OpenWeave on the nRF52840.  So the default OpenThread options we select should be the right ones for all such products, not just locks.

I'm fine with this change if we think it should apply universally.  However, if we don't, then I think individual products should enable them as needed in their product makefiles by overriding OPENTHREAD_CONFIGURE_OPTIONS.  

For example, you can enable jam detection and child supervision in the [openweave-nrf52840-lock-example](https://github.com/openweave/openweave-nrf52840-lock-example) by adding the following lines to the top-level [Makefile](https://github.com/openweave/openweave-nrf52840-lock-example/blob/master/Makefile):

```
OPENTHREAD_CONFIGURE_OPTIONS += \
    --enable-child-supervision \
    --enable-jam-detection
```

(Insert these lines after the include statements and before the call to GenerateBuildRules).

This would also be a quick way for our device partners to test this feature without waiting for an OpenWeave change.

bukepo:
Moved into lock example.

PR #483, opened 2020-02-05
Bug/misc fixes

robszewczyk:
Please excuse the co-mingling of multiple fixes in a single pull request, each of the changes otherwise has absolutely minimal content which in total does not clearly merrit a full build. 
Changes in this pull request: 

* in `TraitSchemaEngine::ParseTagString` static analysis returned a warning that the comparison was always true because of limited range of the operands.  The change correctly checks the parsed result prior to narrowing
* `GenericTraitCatalogImpl.cpp` contains an explicit instantiation of the template classes.  These classes rely on richer C++ library than what's available on many of our targets.  The change guards the template instantiation with a compile time instantiation; the developer is free to either enable the flag or pull in the templates and instantiate them directly, but the build system continues to compile without `<map>` support
* Fixes for dead stores flagged by static analysis

### DIFF
--- a/src/lib/profiles/data-management/Current/TraitData.cpp
+++ b/src/lib/profiles/data-management/Current/TraitData.cpp
@@ -91,16 +91,15 @@ WEAVE_ERROR UpdateDictionaryDirtyPathCut::CutPath(PropertyPathHandle aPathhandle
 WEAVE_ERROR TraitSchemaEngine::ParseTagString(const char * apTagString, char ** apEndptr, uint8_t & aParseRes) const
 {
     WEAVE_ERROR err = WEAVE_NO_ERROR;
-
+    unsigned long int tag;
     VerifyOrExit(apTagString != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
     VerifyOrExit(*apTagString == '/', err = WEAVE_ERROR_INVALID_ARGUMENT);
 
     apTagString++;
-
-    aParseRes = strtoul(apTagString, apEndptr, 0);
+    tag = strtoul(apTagString, apEndptr, 0);
     VerifyOrExit(!(*apEndptr == apTagString || (**apEndptr != '\0' && **apEndptr != '/')), err = WEAVE_ERROR_INVALID_ARGUMENT);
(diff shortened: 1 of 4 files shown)


src/lib/profiles/data-management/Current/TraitData.cpp
```diff
@@ -91,16 +91,16 @@ WEAVE_ERROR UpdateDictionaryDirtyPathCut::CutPath(PropertyPathHandle aPathhandle
...
-
+    unsigned long int tag;
     VerifyOrExit(apTagString != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
     VerifyOrExit(*apTagString == '/', err = WEAVE_ERROR_INVALID_ARGUMENT);
 
     apTagString++;
 
-    aParseRes = strtoul(apTagString, apEndptr, 0);
+    tag = strtoul(apTagString, apEndptr, 0);
     VerifyOrExit(!(*apEndptr == apTagString || (**apEndptr != '\0' && **apEndptr != '/')), err = WEAVE_ERROR_INVALID_ARGUMENT);
```
jaylogue: This code fails to check for ERANGE errors, e.g. when the value given exceeds that which will fit within unsigned long.  See ParseInt() in nlargparser for required logic: https://github.com/openweave/openweave-core/blob/7627fe106fa56abcecfc126de1c21b0f24fe39be/src/lib/support/nlargparser.cpp#L698-L708

robszewczyk: In principle, you're absolutely correct.  In practice, the runtime difference in the code is which  error we return -- the `ERANGE` case is effectively handled by the check in line 102.  I believe either error is appropriate to return in the `ERANGE` case; the main difference is that if at some point the `kContextTagMaxNum` became larger than `ULONG_MAX` the code would no longer work (although in that case it should give us the compile time warning that precipitated this change).

I typed in the above paragraphs largely because I was wary of checking in device code that relied on `errno`.  Only at the end did I process that this code is executing on platforms that need to process strings, etc. 

I will change this to follow specs of `strtoul`.

robszewczyk: I tried the fixes with `errno` in place, ended up running into issues with embedded targets still somehow pulling in this method.  Given the reasoning above, I'm going to go back to the original code.

PR #487, opened 2020-02-07
several changes on update server

yunhanw-google:
-- Remove unsupported kEventChangeBegin and kEventChangeEnd
-- Add kEventUpdateProcessingComplete, trigger it when updating trait version.
-- Update status report using kStatus_BadRequest when update request is malformed.
-- Add kStatus_InvalidValueInUpdate to represent data element value error in wdm
profie.
-- When processing update request, we need to skip this DE when DE generates
WEAVE_ERROR_WRONG_TLV_TYPE and WEAVE_ERROR_TLV_TAG_NOT_FOUND.
-- Rename BuildStatusDataHandleElement to UpdateStatusDataHandleElement
-- Add InitializeStatusDataHandleList to initialize StatusDataHandleList
with defaut incorrect state, whenenever the data element in list has
been processed successfully, update StatusDataHandleList with sucess
status. Once it hit fatal error, we end up the conditional and
unconditional data list loop, and update trait version for successful
DEs.
-- note: when starting to process update request,
update.CheckSchemaValidity() is triggered so that malformed update request
should be checked, which needs feature flags, WEAVE_CONFIG_DATA_MANAGEMENT_ENABLE_SCHEMA_CHECK
and WEAVE_DETAIL_LOGGING, and they are defaulted enabled.

### DIFF
--- a/src/lib/profiles/data-management/Current/SubscriptionEngine.cpp
+++ b/src/lib/profiles/data-management/Current/SubscriptionEngine.cpp
@@ -39,8 +39,7 @@
 #include <SystemLayer/SystemStats.h>
 
 #ifndef WEAVE_WDM_ALIGNED_TYPE
-#define WEAVE_WDM_ALIGNED_TYPE(address, type) \
-  reinterpret_cast<type *> WEAVE_SYSTEM_ALIGN_SIZE((size_t)(address), 4)
+#define WEAVE_WDM_ALIGNED_TYPE(address, type) reinterpret_cast<type *> WEAVE_SYSTEM_ALIGN_SIZE((size_t)(address), 4)
 #endif
 
 namespace nl {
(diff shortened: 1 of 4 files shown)


src/lib/profiles/data-management/Current/TraitData.h
```diff
@@ -987,19 +987,14 @@ class TraitUpdatableDataSource : public TraitDataSource, protected TraitSchemaEn
...
         /* Start of a data element */
         kEventDataElementBegin,
 
         /* End of a data element */
         kEventDataElementEnd,
 
-        /* End of a change record */
-        kEventChangeEnd,
+        /* a change record happend */
+        kEventChangeHappend,
```
jaylogue: This name is not specific enough.  As far as I can tell, this event is raised at the end of update processing if the data source was updated.  Thus the name should be something like UpdateProcessingComplete.

yunhanw-google: Done, thanks

src/lib/profiles/data-management/Current/SubscriptionEngine.cpp
```diff
@@ -1561,6 +1560,36 @@ WEAVE_ERROR SubscriptionEngine::AllocateRightSizedBuffer(PacketBuffer *& buf, co
...
+                                                               uint32_t & aNumDataElements)
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+    Weave::TLV::TLVReader dataReader;
+    dataReader.Init(aReader);
+
+    for (aNumDataElements = 0; WEAVE_NO_ERROR == (err = dataReader.Next()); aNumDataElements++)
+    {
+
+        apStatusDataHandleList[aNumDataElements].mProfileId       = Weave::Profiles::kWeaveProfile_Common;
```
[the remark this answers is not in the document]
yunhanw-google: Done, thanks!

PR #492, opened 2020-02-10
New to openweave

marcosb:
Hi,

Googler here looking to contribute; I was perusing the code and noticed there's a lot of manual locks/unlocks, instead of using RAII.

Is there a reason for not using SBRM here? It seems that if anything were to go wrong currently, the lock would be held indefinitely.

Thanks for your time, and if I missed something, let me know!

### DIFF
--- a/src/inet/AsyncDNSResolverSockets.cpp
+++ b/src/inet/AsyncDNSResolverSockets.cpp
@@ -36,6 +36,29 @@
 
 namespace nl {
 namespace Inet {
+    
+class LockHolder {
+ public:
+    LockHolder(pthread_mutex_t ayncDNSMutex)
+        :mAsyncDNSMutex(ayncDNSMutex)
+    {
+        int pthreadErr;
+
+        pthreadErr = pthread_mutex_lock(&mAsyncDNSMutex);
+        VerifyOrDie(pthreadErr == 0);
+    }
+
+    ~LockHolder()
+    {
+        int pthreadErr;
+
+        pthreadErr = pthread_mutex_unlock(&mAsyncDNSMutex);
+        VerifyOrDie(pthreadErr == 0);
+    }
+    
+ private:
+    pthread_mutex_t         mAsyncDNSMutex;
+}
 
 /**
  *  The explicit initializer for the AsynchronousDNSResolverSockets class.
@@ -87,14 +110,14 @@ INET_ERROR AsyncDNSResolverSockets::Shutdown(void)
     INET_ERROR err = INET_NO_ERROR;
     int pthreadErr;
 
-    AsyncMutexLock();
+    {
+        LockHolder asyncMutexLock(mAsyncDNSMutex);
 
-    mInet->State = InetLayer::kState_ShutdownInProgress;
+        mInet->State = InetLayer::kState_ShutdownInProgress;
 
-    pthreadErr = pthread_cond_broadcast(&mAsyncDNSCondVar);
-    VerifyOrDie(pthreadErr == 0);
-
-    AsyncMutexUnlock();
+        pthreadErr = pthread_cond_broadcast(&mAsyncDNSCondVar);
+        VerifyOrDie(pthreadErr == 0);
+    }
 
     // Have the Weave thread join the thread pool for asynchronous DNS resolution.
     for (int i = 0; i < INET_CONFIG_DNS_ASYNC_MAX_THREAD_COUNT; i++)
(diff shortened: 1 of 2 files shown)


jaylogue:
Hi Marcos,

Thanks for your contribution.  The pattern you propose is certainly not unacceptable for use in OpenWeave, especially given that the size and complexity of generated code is effectively the same.  That said, I think we have steered away from this pattern because of its limited value in our context; specifically due to our preference for a single exit path for methods and the fact that we don't use exceptions.  The other issue I see with RAII patterns is the poor readability for people not intimately familiar with the pattern--e.g. embedded C developers.  The lack of definitive "lock is taken here" / "lock is released here" statements can obscure the true behavior of the code.

marcosb:
Hey @jaylogue, thanks for the response! I probably picked a terrible example of where to use RAII, apologies for that - I picked a simple file just to get the discussion started.

Object::Release is probably a better example of multiple exit points & a place where RAII would be really useful; my thought was that if weave uses this pattern throughout, instances like this would not look so frightening & be so bug-prone:

https://github.com/openweave/openweave-core/blob/master/src/inet/DNSResolver.cpp

I can get a prototype CL for that file if you think it's worthwhile, just wanted to get a quick discussion going first.

marcosb:
OK made the sample changes to DNSResolver; not totally sure if it compiles but I'll check once it's done.

At least it gives an idea though of how enforcing the RAII pattern (even in places where it doesn't help much) could help guide people to using this to write safer code.

E.g. if we go this route, we can actually make the Release method private, so that only AutoRelease can do the releasing, thus guiding engineers to use the safer & more readable auto-release method

marcosb:
@jaylogue So this is interesting; you'll notice that in my changes to auto-release, one of the exit points wasn't releasing:
https://github.com/openweave/openweave-core/blob/master/src/inet/DNSResolver.cpp#L200

So there's 2 possibilities here:

1. This was a bug & it should've been releasing
2. This was WAI (I think it's reasonable to add a method to NOT auto-release in cases like this), but then there's no unit test coverage here

src/inet/DNSResolver.cpp
```diff
@@ -298,12 +292,11 @@ INET_ERROR DNSResolver::Cancel()
  */
 void DNSResolver::HandleResolveComplete()
 {
+    Weave::System::Object::AutoRelease objectRelease(*this);
```
jaylogue: I see this as an anti-pattern.  In this case we're dealing with a reference count, vs. a lock, where the lifetime of reference spans multiple states of the object.  In this case, a single call to a method called Release() is clearly the most obvious pattern.

src/inet/DNSResolver.cpp
```diff
@@ -91,9 +91,9 @@ INET_ERROR DNSResolver::Resolve(const char *hostName, uint16_t hostNameLen, uint
         DNSResolver::OnResolveCompleteFunct onComplete, void *appState)
 {
     INET_ERROR res = INET_NO_ERROR;
+    Weave::System::Object::AutoRelease objectRelease(*this);
```
jaylogue: Similar to my comment below, I don't see this as a desirable pattern.  Rather, I would argue that the returns should be replaced with ExitNow() jumps to an exit: label containing a single call to Release().

src/inet/AsyncDNSResolverSockets.cpp
```diff
@@ -172,8 +195,7 @@ INET_ERROR AsyncDNSResolverSockets::EnqueueRequest(DNSResolver &resolver)
 {
     INET_ERROR err = INET_NO_ERROR;
     int pthreadErr;
-
-    AsyncMutexLock();
+    LockHolder asyncMutexLock(*this);
```
jaylogue: Similar to above, I think this pattern makes it easy to miss the acquisition of the lock.  I suggest always using an independent block to clearly demark the scope of the lock, with the lock being the only variable declared at the start of the block.  E.g.:

```
INET_ERROR AsyncDNSResolverSockets::EnqueueRequest(DNSResolver &resolver)
{
    INET_ERROR err = INET_NO_ERROR;
    int pthreadErr;

    {
        auto lock = LockHolder::AquireLock(*this);

        ...
    }

    return err;
}
```

src/inet/AsyncDNSResolverSockets.cpp
```diff
@@ -87,14 +110,14 @@ INET_ERROR AsyncDNSResolverSockets::Shutdown(void)
     INET_ERROR err = INET_NO_ERROR;
     int pthreadErr;
 
-    AsyncMutexLock();
-
-    mInet->State = InetLayer::kState_ShutdownInProgress;
+    {
+        LockHolder asyncMutexLock(*this);
```
jaylogue: This illustrates one of the downsides of this pattern--the point at which the lock is acquired doesn't read as an imperative statement, and hence is easily missed by the unaware reader.

I think a construct like this would be more obvious:

```
{
    auto lock = LockHolder::AquireLock(*this);

    ...
}
```

PR #498, opened 2020-02-25
Support for suspending and restoring Weave security sessions

jaylogue:
- Added methods to WeaveFabricState to suspend and restore an active Weave security session in a serialized form.  This feature can be used by devices that do not retain RAM when sleeping to avoid the need to re-establish sessions on wake.

- Added option to weave-ping tool to exercise the session suspend/restore APIs.

### DIFF
--- a/src/lib/core/WeaveFabricState.cpp
+++ b/src/lib/core/WeaveFabricState.cpp
@@ -62,6 +62,7 @@ using namespace nl::Weave::Encoding;
 using namespace nl::Weave::Crypto;
 using namespace nl::Weave::Profiles;
 using namespace nl::Weave::Profiles::FabricProvisioning;
+using namespace nl::Weave::Profiles::Security;
 using namespace nl::Weave::Profiles::Security::AppKeys;
 
 #if WEAVE_CONFIG_SECURITY_TEST_MODE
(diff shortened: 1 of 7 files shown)


src/lib/core/WeaveFabricState.cpp
```diff
@@ -602,6 +590,294 @@ WEAVE_ERROR WeaveFabricState::GetSharedSessionEndNodeIds(const WeaveSessionKey *
...
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+    WeaveSessionKey * sessionKey;
+
+    // Lookup the specified session.
+    err = GetSessionKey(keyId, peerNodeId, sessionKey);
+    SuccessOrExit(err);
+
+    // Assert various requirements about the session.
+    VerifyOrExit(sessionKey->IsKeySet(), err = WEAVE_ERROR_KEY_NOT_FOUND);
+    VerifyOrExit(!sessionKey->IsSuspended(), err = WEAVE_ERROR_KEY_NOT_FOUND);
```
pidarped: err = WEAVE_ERROR_SESSION_KEY_SUSPENDED

src/lib/core/WeaveFabricState.cpp
```diff
@@ -602,6 +590,294 @@ WEAVE_ERROR WeaveFabricState::GetSharedSessionEndNodeIds(const WeaveSessionKey *
...
+        }
+    }
+}
+
+/**
+ * Suspend and serialize the state of an active Weave Security Session.
+ *
+ */
+WEAVE_ERROR WeaveFabricState::SuspendSession(uint16_t keyId, uint64_t peerNodeId, uint8_t * buf, uint16_t bufSize, uint16_t & serializedSessionLen)
+{
```
pidarped: Do you think this method should try to suspend a session if it is not a shared one?

jaylogue: I don't see any reason not to.

src/lib/core/WeaveFabricState.cpp
jaylogue: Fixed in latest commit.

src/lib/core/WeaveFabricState.cpp
pidarped: I guess the rationale was to keep the method agnostic of the session type?

jaylogue: Mostly, yes.  By design, it only supports non-connection oriented CASE sessions.  However I didn't see any reason to block its use, e.g., for a device-to-device CASE session.

pidarped: Thanks. Overall LGTM.

PR #502, opened 2020-02-25
Add Weave Pairing Codes document

jaylogue:


Vyrastas:
Do you want to add all these PDFs to openweave.io too?  We've already got the others hosted here https://openweave.io/reference/specifications

jaylogue:
@Vyrastas Yes, I think that makes sense.

PR #503, opened 2020-02-25
Enable obsolscent interface when compiling objective c code in mac os

yunhanw-google:
-- When compiling weave objective c device manager in mac os, Unknown type name Inet buffer compilation issue happens, needs to enable obsolscent interface.


yunhanw-google:
> Please explain what compilation errors this fixes and why objective-c code is using obsolete interfaces.

Yes, you are right, we don't need this change. The test application use the obsolete code. I will close this one, thanks

PR #506, opened 2020-02-28
Provide implementation of if_nameindex

robszewczyk:
Android introduced the above functions in API version 24. This patch
implements these functions in terms of getifaddrs.  Implementation is
also provided for the matching if_freenameindex.  This backport may
show slight differences from the if_nameindex -- the implementation
exploits the observation that getifaddrs returns IP addresses grouped
by interface name; should this property be violated, the resulting
if_nameindex will contain duplicate entries.


src/inet/InetInterface.cpp
```diff
@@ -226,6 +226,125 @@ void CloseIOCTLSocket(void)
 
 #if WEAVE_SYSTEM_CONFIG_USE_SOCKETS
 
+#if __ANDROID_API__ >= 21 && __ANDROID_API__ < 24
```
g-coppock: Could this be:

`#if __ANDROID__ && __ANDROID_API__ < 24` instead? That way this can also work when building for < 21, as well as for non-Android targets.

src/inet/InetInterface.cpp
```diff
@@ -226,6 +226,125 @@ void CloseIOCTLSocket(void)
...
+            tmpval[intfIter].if_index = index;
+            tmpval[intfIter].if_name  = strdup(addrIter->ifa_name);
+            intfIter++;
+        }
+        lastIntfName = addrIter->ifa_name;
+    }
+
+    // coalesce on interface index
+    maxIntfNum = 0;
+    for (size_t i = 0; tmpval[i].if_index != 0; i++)
```
g-coppock: I may be mistaken but I think you could just use the value of `intfIter` here rather than looping through again? That value is only incremented when index != 0 so it should be equivalent.

src/inet/InetInterface.cpp
```diff
@@ -251,7 +370,11 @@ InterfaceIterator::~InterfaceIterator(void)
 {
     if (mIntfArray != NULL)
     {
+#if __ANDROID_API__ >= 21 && __ANDROID_API__ < 24
```
g-coppock: Same here for the condition

src/inet/InetInterface.cpp
```diff
@@ -304,7 +427,11 @@ bool InterfaceIterator::Next(void)
 
     if (mIntfArray == NULL)
     {
+#if __ANDROID_API__ >= 21 && __ANDROID_API__ < 24
```
g-coppock: Same here for the condition

src/inet/InetInterface.cpp
jaylogue: I think we should wait until we understand the utility of this work-around before we apply it to non-Android targets.

g-coppock: Sorry, to clarify I just meant that having the __ANDROID__ flag would ensure it's *only* for Android, but applies to Android for all API levels < 24.

src/inet/InetInterface.cpp
robszewczyk: Not quite.  The list returned from `getifaddrs` contains an entry for each IP address in the system.  I've tried to make very few assumptions about the ordering of that list, and I've been trying to make no assumptions on whether the `if_index` is dense. 

1. the first pass through the list from `getifaddrs` coalesces the on consecutive entries with the same interface name.  Arguably this could be an optional step.  This (along with just the projection from `struct ifaddrs` to `struct if_nameindex` ) takes place in lines 265..293. Note that at this point in time, `tmpval` is dense, it may contain duplicate entries for the same interface, and there is no relationship between the number of entries and the max index of the interface. 
2. the second pass (lines 295..311) reshuffles the entries: we use the `if_index` to index into the array to be returned.   This creates a duplicate free array of interfaces, but this array is potentially sparse -- if there are gaps between interface indeces, the array will contain 0s in those entries, and thus not conform to the expected output of `if_nameindex`. 
3. the third pass (lines 316..321)  coalesces the empty entries.   At the end of the pass, all the entries with `0` in `if_index` are at the end of the array.

g-coppock: Got it, thanks for the explanation!

src/inet/InetInterface.cpp
robszewczyk: Agreed.  I believe we only use `__ANDROID__API__ > = 21`, until we have a use case here, I'd hold off.

g-coppock: Currently the PairingKit library in google3 compiles with 19, so that means it will fail to compile, as it will be using the API 24 function. There should be no need to specify a minimum here, just that it's android and < 24.

robszewczyk: OK, will change accordingly

robszewczyk: changed

PR #507, opened 2020-03-02
Fix Objective C warnings

robszewczyk:
Fixed the following warnings:

* conflicting return type in implementation of 'setNodeId:': 'void *' vs 'void'
* extra tokens at end of #endif directive
* unannotated fall-through between switch labels
* designated initializer should only invoke a designated initializer on 'super'
* convenience initializer should not invoke an initializer on 'super'
* hiding of overloaded virtual functions


yunhanw-google:
Thanks for addressing these warnings!

src/device-manager/cocoa/NLNetworkInfo.h
```diff
@@ -60,7 +60,7 @@ extern const int NLThreadChannel_NotSpecified;
 @property (nonatomic, strong) NSData * ThreadPSKc; // The Thread pre-shared key commissioner, or NULL if not specified.
 
 
-- (instancetype)init NS_DESIGNATED_INITIALIZER;
+- (instancetype)init;
```
robszewczyk: Reviewers note: the `NS_DESIGNATED_INITIALIZER` is specified in a protected header

PR #508, opened 2020-03-02
Improved multi-process support in WeaveDeviceManager

jaylogue:
- Assign a random Weave node id to the local Weave stack when the WeaveDeviceManager is used on Android, iOS and from within Python.  This ensures that devices will treat separate WeaveDeviceManager instances running in different processes on the same host as distinct Weave nodes.

- Enabled use of an ephemeral UDP port for outgoing communication from the WeaveDeviceManager on the Android, iOS and Python.

- Added a build feature that enables listening on the Weave UDP port only when the Device Manager is attempting to identify or rendezvous with a device.  This reduces competition for the Weave port when the WeaveDeviceManager is used by multiple processes on the same host. This feature is enabled by default on Android, iOS and within Python.

- Minor adjustments to logging in WeaveDeviceManager, WeaveMessageLayer and Python wrapper code.

### DIFF
--- a/src/device-manager/WeaveDeviceManager.cpp
+++ b/src/device-manager/WeaveDeviceManager.cpp
@@ -139,6 +139,9 @@ WEAVE_ERROR WeaveDeviceManager::Init(WeaveExchangeManager *exchangeMgr, WeaveSec
     mUseAccessToken = true;
     mConnectedToRemoteDevice = false;
     mIsUnsecuredConnectionListenerSet = false;
+#if WEAVE_CONFIG_DEVICE_MGR_DEMAND_ENABLE_UDP
+    mUDPEnabled = false;
+#endif // WEAVE_CONFIG_DEVICE_MGR_DEMAND_ENABLE_UDP
     mPingSize = 0;
     mTokenPairingCertificate = NULL;
     mTokenPairingCertificateLen = 0;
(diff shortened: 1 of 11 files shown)


src/device-manager/WeaveDeviceManager.cpp
```diff
@@ -2726,6 +2741,69 @@ void WeaveDeviceManager::ClearOpState()
...
+
+exit:
+    return err;
+}
+
+WEAVE_ERROR WeaveDeviceManager::DisableUDP(void)
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+
+    // If UDP listening was enabled on demand, disable is once there are no
```
g-coppock: `disable is` -> `disable it`

`instances need` -> `instances that need

src/device-manager/WeaveDeviceManager.cpp
```diff
@@ -2726,6 +2741,69 @@ void WeaveDeviceManager::ClearOpState()
...
+
+        if (!mUDPEnabled)
+        {
+            sUDPDemandEnableCount++;
+            mUDPEnabled = true;
+        }
+    }
+    else
+    {
+        if (sUDPDemandEnableCount > 0 && !mUDPEnabled)
```
pidarped: Can sUDPDemandEnableCount == 0 in this else block?

src/device-manager/WeaveDeviceManager.cpp
```diff
@@ -2726,6 +2741,69 @@ void WeaveDeviceManager::ClearOpState()
...
 
+#if WEAVE_CONFIG_DEVICE_MGR_DEMAND_ENABLE_UDP
+
+uint32_t WeaveDeviceManager::sUDPDemandEnableCount = 0;
+
+WEAVE_ERROR WeaveDeviceManager::EnableUDP(void)
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+
+    // If UDP listening is NOT already enable in the Weave Message Layer,
```
pidarped: already enable -> already enabled

src/device-manager/WeaveDeviceManager.cpp
jaylogue: Yes.  If UDP has been enabled in the Message Layer globally, i.e. before any Device Manager object attempts to use UDP, then sUDPDemandEnableCount will be 0.  This signals to the code that "enable counting" is not being performed because UDP is on globally.

PR #510, opened 2020-03-03
add TESTS=0|1 option in standalone compilation

yunhanw-google:
-- some traits, for example, localsetting and localecapability trait, are used not only in mobile client, but also in mock-device, for that case, when using weave data management client in mobile client or linux pairing tool, we need to disalbe these internal trait via TESTS=0 in standalone compilation.

### DIFF
--- a/Makefile-Standalone
+++ b/Makefile-Standalone
@@ -38,6 +38,7 @@ USE_LWIP                       ?= 0
 NO_OPENSSL                     ?= 0
 BLUEZ                          ?= 0
 USE_FUZZING                    ?= 0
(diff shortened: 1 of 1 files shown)


jaylogue:
> for that case, when using weave data management client in mobile client or linux pairing tool, we need to disalbe these internal trait

This makefile is for building OpenWeave for stand-alone (i.e. linux / mac desktop), not for building for mobile clients.  So why is this needed?

jaylogue:
@wijnandatgoogle Should you be using a different makefile other than StandAlone?

PR #512, opened 2020-03-04
Persistent service directory

didishe90:
-Add WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY for enabling and disabling this feature
-Save the byte stream received from directory server to persistent storage if ‘ServiceDirectory Redirect’ is not issued.
-Clear persisted storage whenever cache is cleared.

### DIFF
--- a/src/lib/core/WeaveConfig.h
+++ b/src/lib/core/WeaveConfig.h
@@ -1905,6 +1905,23 @@
 #endif // WEAVE_CONFIG_ENABLE_SERVICE_DIRECTORY
 
 /**
+ *  @def WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+ *
+ *  @brief
+ *    If set to (1), persistent service directory is enabled.
+ *    Default value is (0) or disabled.
+ *
+ *  @note
+ *    Enabling this feature flag persists the service directory
+ *    cache retrieved from the directory server and use it to
+ *    populate the in-memory cache at initialization time.
+ *
+ */
+#ifndef WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+#define WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY               0
+#endif // WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+
+/**
  *  @def WEAVE_CONFIG_SERVICE_DIR_CONNECT_TIMEOUT_MSECS
  *
  *  @brief
--- a/src/lib/profiles/service-directory/ServiceDirectory.cpp
+++ b/src/lib/profiles/service-directory/ServiceDirectory.cpp
@@ -358,6 +358,30 @@ WEAVE_ERROR WeaveServiceManager::init(WeaveExchangeManager *aExchangeMgr,
 
     clearCacheState();
 
+#if WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+
+    if (Platform::IsPersistentServiceDirPresent(kPersistedServiceDirVersion))
+    {
+        WeaveLogProgress(ServiceDirectory, "Persistent service directory exists");
+
+        err = loadPersistentServiceDirIntoCache();
+
+        if (err != WEAVE_NO_ERROR)
+        {
+            Platform::ClearPersistentServiceDir();
+            WeaveLogProgress(ServiceDirectory, "Restore error, persistent service directory cleared");
+
+        }
+        else
+        {
+            mCacheState = kServiceMgrState_Resolved;
+            WeaveLogProgress(ServiceDirectory, "Persistent service directory successfully restored");
+        }
+
+    }
+
+#endif // WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+
     finalizeConnectRequests();
 
 exit:
@@ -427,6 +451,10 @@ WEAVE_ERROR WeaveServiceManager::connect(uint64_t aServiceEp,
          * the only way forward is to get the root directory from the
          * service config and install it.
          */
+
+#if WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+        Platform::ClearPersistentServiceDir();
+#endif // WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
 
         err = mAccessor(mCache.base, mCache.length);
         SuccessOrExit(err);
@@ -903,6 +931,11 @@ void WeaveServiceManager::unresolve(void)
 
     if (mCacheState > kServiceMgrState_Resolving)
     {
+
+#if WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+        Platform::ClearPersistentServiceDir();
+#endif //WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+
         cleanupExchangeContext();
 
         mCacheState = kServiceMgrState_Resolving;
@@ -948,6 +981,10 @@ void WeaveServiceManager::reset(void)
 
     clearWorkingState();
     clearCacheState();
+
+#if WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+    Platform::ClearPersistentServiceDir();
+#endif //WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
 
     finalizeConnectRequests();
 }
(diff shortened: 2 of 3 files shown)


src/lib/core/WeaveConfig.h
```diff
@@ -1904,6 +1904,23 @@
...
 
+/**
+ *  @def WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+ *
+ *  @brief
+ *    If set to (1), persistent service directory is enabled.
+ *    Default value is (0) or disabled.
+ *
+ *  @note
+ *    Enabling this profile persists service directory after
```
pidarped: Recommend rephrasing as "Enabling this feature flag persists the service directory cache retrieved from the directory server and use it to populate the in-memory cache at initialization time.

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -341,6 +341,10 @@ WEAVE_ERROR WeaveServiceManager::init(WeaveExchangeManager *aExchangeMgr,
 
     WeaveLogProgress(ServiceDirectory, "init()");
 
+#if WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+    PacketBuffer* msgBuf = NULL;
+#endif
+
```
pidarped: Move this before the logging line.

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -358,10 +362,41 @@ WEAVE_ERROR WeaveServiceManager::init(WeaveExchangeManager *aExchangeMgr,
 
     clearCacheState();
 
+#if WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+
+    if (platform::PersistentServiceDirExist(kMsgType_ServiceEndpointResponse))
```
pidarped: Recommend renaming method as IsPersistentServiceDirPresent(..). Also, potentially define a local variable version and assigning to it the messageType before passing to the function.

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -358,10 +362,41 @@ WEAVE_ERROR WeaveServiceManager::init(WeaveExchangeManager *aExchangeMgr,
...
+        SuccessOrExit(err);
+
+        msgBuf->SetDataLength(len);
+
+        err = unpackPacketBuffer(msgBuf, false);
+        SuccessOrExit(err);
+
+        mCacheState = kServiceMgrState_Resolved;
+
+        WeaveLogProgress(ServiceDirectory, "resolved");
```
pidarped: Add some additional context in this log clarifying that it was fetched from persisted cache.

pidarped: The above logic could be encapsulated inside a function and called from init(). Name of function could be something like LoadPersistentServiceDirIntoCache(). That would hide all the PacketBuffer related logic inside that function.

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -949,6 +984,10 @@ void WeaveServiceManager::reset(void)
     clearWorkingState();
     clearCacheState();
 
+#if WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+    platform::ClearPersistentServiceDir();
+#endif //WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+
```
pidarped: What about the unresolve() call?

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -1576,6 +1548,132 @@ WEAVE_ERROR WeaveServiceManager::lookupAndConnect(WeaveConnection *aConnection,
...
+    SuccessOrExit(err);
+
+exit:
+
+    if (msgBuf != NULL)
+        PacketBuffer::Free(msgBuf);
+
+    return err;
+}
+#endif
```
pidarped: Add comment for ifdef at above line. The pretty check is failing.

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -1133,79 +1165,20 @@ void WeaveServiceManager::onResponseReceived(uint32_t aProfileId, uint8_t aMsgTy
...
-             * here we have directory information beyond the root directory but
-             * we're not done yet.
-             */
-
-            mDirectory.length = dirLen;
-            writePtr = mDirectory.base = mCache.base;
-
-            err = cacheDirectory(i, mDirectory.length, writePtr);
+            WeaveLogProgress(ServiceDirectory, "not redirecting, store service directory");
+            err = platform::StorePersistentServiceDir(aMsg->Start(),
```
jaylogue: The behavior of this function is unclear.  If a copy of SD exists on disk when StorePersistentServiceDir() is called, but the version number passed to the function does not match the version stored on disk, is StorePersistentServiceDir() expected to clear the old version before writing the new version?

I would much prefer that this code called ClearPersistentServiceDir() before calling StorePersistentServiceDir().  In fact, I would prefer that it called ClearPersistentServiceDir() _before_ sending the service directory request.

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -1576,6 +1549,132 @@ WEAVE_ERROR WeaveServiceManager::lookupAndConnect(WeaveConnection *aConnection,
...
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+
+    WeaveLogProgress(ServiceDirectory, "Unpacks the packetBuffer to cache");
+
+    /*
+     * this block unpacks the service directory message.
+     */
+
+    {
```
jaylogue: This block is unnecessary.

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -1576,6 +1549,132 @@ WEAVE_ERROR WeaveServiceManager::lookupAndConnect(WeaveConnection *aConnection,
...
+            SuccessOrExit(err);
+        }
+
+        else
+        {
+            mSuffixTable.length = 0;
+            mSuffixTable.base = NULL;
+        }
+
+        if (timePresent & useTimePresent)
```
jaylogue: Use boolean AND, please!  (& -> &&)

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -1576,6 +1549,132 @@ WEAVE_ERROR WeaveServiceManager::lookupAndConnect(WeaveConnection *aConnection,
...
+
+/**
+ *  @brief
+ *    This method unpacks the packetbuffer to cache.
+ *
+ *  @param [in] aMsg         The content of this message.
+ *  @param [in] useTimePresent If need to handle time info
+ *  @param [in] redir        Return whether need to redirect
+ */
+WEAVE_ERROR WeaveServiceManager::unpackPacketBuffer(PacketBuffer *aMsg, bool useTimePresent, bool *redir)
```
jaylogue: useTimePresent -> useTimeInfo

src/lib/profiles/service-directory/ServiceDirectory.cpp
```diff
@@ -358,6 +362,21 @@ WEAVE_ERROR WeaveServiceManager::init(WeaveExchangeManager *aExchangeMgr,
...
     clearCacheState();
 
+#if WEAVE_CONFIG_PERSIST_SERVICE_DIRECTORY
+
+    if (platform::IsPersistentServiceDirPresent(version))
+    {
+        WeaveLogProgress(ServiceDirectory, "Persistent Service Directory Exist");
+
+        err = loadPersistentServiceDirIntoCache();
+        SuccessOrExit(err);
```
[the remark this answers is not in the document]
didishe90: I updated. Please take a look.

src/lib/profiles/service-directory/ServiceDirectory.cpp
didishe90: I called ClearPersistentServiceDir() in "connect", if state is "Initial". Please take a look.

PR #519, opened 2020-03-13
Add an option to allocate logging buffers on heap

kghost:
There is limit data segment on some FreeRTOS platform, because the
logging buffer is very large, we may want to move logging buffers
from static data segment to dynamic allocated heap.

The patch introduced an new define option:

WEAVE_DEVICE_CONFIG_EVENT_LOGGING_BUFFER_STATIC

When it is defined, the logging buffer will be allocated statically
on data segment; when it is not defined, the logging buffers will be
allocated dynamically on heap.

Signed-off-by: Zang MingJie <mingjiez@google.com>

### DIFF
--- a/src/adaptations/device-layer/EventLogging.cpp
+++ b/src/adaptations/device-layer/EventLogging.cpp
@@ -35,21 +35,22 @@ namespace Weave {
 namespace DeviceLayer {
 namespace Internal {
 
+/* BSS segment is limited on some device, provides an option to allocate large memory buffer on heap */
+#ifdef WEAVE_DEVICE_CONFIG_EVENT_LOGGING_BUFFER_STATIC
 uint64_t gCritEventBuffer[RoundUp(WEAVE_DEVICE_CONFIG_EVENT_LOGGING_CRIT_BUFFER_SIZE, sizeof(uint64_t)) / sizeof(uint64_t)];
-static nl::Weave::PersistedCounter sCritEventIdCounter;
-
 uint64_t gProdEventBuffer[RoundUp(WEAVE_DEVICE_CONFIG_EVENT_LOGGING_PROD_BUFFER_SIZE, sizeof(uint64_t)) / sizeof(uint64_t)];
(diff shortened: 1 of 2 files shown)


src/adaptations/device-layer/EventLogging.cpp
```diff
@@ -35,21 +35,22 @@ namespace Weave {
...
 #if WEAVE_DEVICE_CONFIG_EVENT_LOGGING_INFO_BUFFER_SIZE
 uint64_t gInfoEventBuffer[RoundUp(WEAVE_DEVICE_CONFIG_EVENT_LOGGING_INFO_BUFFER_SIZE, sizeof(uint64_t)) / sizeof(uint64_t)];
-static nl::Weave::PersistedCounter sInfoEventIdCounter;
 #endif
-
 #if WEAVE_DEVICE_CONFIG_EVENT_LOGGING_DEBUG_BUFFER_SIZE
 uint64_t gDebugEventBuffer[RoundUp(WEAVE_DEVICE_CONFIG_EVENT_LOGGING_DEBUG_BUFFER_SIZE, sizeof(uint64_t)) / sizeof(uint64_t)];
-static nl::Weave::PersistedCounter sDebugEventIdCounter;
 #endif
+#endif
```
gerickson: Recommend preprocessor label comment here.

src/adaptations/device-layer/EventLogging.cpp
```diff
@@ -35,21 +35,22 @@ namespace Weave {
...
+
+static nl::Weave::PersistedCounter sCritEventIdCounter;
+static nl::Weave::PersistedCounter sProdEventIdCounter;
+static nl::Weave::PersistedCounter sInfoEventIdCounter;
+static nl::Weave::PersistedCounter sDebugEventIdCounter;
```
jaylogue: These need to be inside #if / #endif blocks so that they are only defined if the corresponding importance level is enabled.

PR #522, opened 2020-03-17
Fix a bug in network telemetry manager

suryanshup:
-- Enable the polling timer only when the device is
   network provisioined.

### DIFF
--- a/src/adaptations/device-layer/NetworkTelemetryManager.cpp
+++ b/src/adaptations/device-layer/NetworkTelemetryManager.cpp
@@ -46,6 +46,7 @@ NetworkTelemetryManager NetworkTelemetryManager::sInstance;
 
 WeaveTelemetryBase::WeaveTelemetryBase()
 {
+    mEnabled = false;
 }
 
 void WeaveTelemetryBase::Init(uint32_t aIntervalMsec)
@@ -91,7 +92,7 @@ WEAVE_ERROR NetworkTelemetryManager::Init(void)
 {
     WEAVE_ERROR err = WEAVE_NO_ERROR;
 
-    WeaveLogProgress(DeviceLayer, "Initiating Network Telemetry & Topology\n");
+    WeaveLogProgress(DeviceLayer, "Initiating Network Telemetry & Topology");
 
 #if WEAVE_DEVICE_CONFIG_ENABLE_WIFI_TELEMETRY
     mWiFiTelemetry.Init(WEAVE_DEVICE_CONFIG_DEFAULT_TELEMETRY_INTERVAL_MS);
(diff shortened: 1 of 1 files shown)


src/adaptations/device-layer/include/Weave/DeviceLayer/WeaveDeviceConfig.h
```diff
@@ -103,7 +103,7 @@
  * The maximum size (in bytes) of a debug logging message.
  */
 #ifndef WEAVE_DEVICE_CONFIG_LOG_MESSAGE_MAX_SIZE
-#define WEAVE_DEVICE_CONFIG_LOG_MESSAGE_MAX_SIZE 256
+#define WEAVE_DEVICE_CONFIG_LOG_MESSAGE_MAX_SIZE 512
```
jaylogue: Why is this needed for all devices?

src/adaptations/device-layer/NetworkTelemetryManager.cpp
```diff
@@ -109,6 +112,39 @@ WEAVE_ERROR NetworkTelemetryManager::Init(void)
...
 }
 
+void NetworkTelemetryManager::OnPlatformEvent(const WeaveDeviceEvent* event)
+{
+#if WEAVE_DEVICE_CONFIG_ENABLE_WIFI_TELEMETRY
+    if (event->Type == DeviceEventType::kWiFiConnectivityChange)
+    {
+        if (ConnectivityMgr().IsWiFiStationProvisioned() && ConnectivityMgr().IsWiFiStationEnabled())
+        {
+            mWiFiTelemetry.Enable();
```
jaylogue: This has the effect of forcing telemetry on whenever WiFi connectivity changes.  Is that the correct thing to do in every case?  It effectively means the pubic Disable() function is useless, because it wont stay disabled for long.

suryanshup: Is there a use case for disabling telemetry once it has been enabled ?

src/adaptations/device-layer/include/Weave/DeviceLayer/WeaveDeviceConfig.h
suryanshup: I'll move this change to 
src/adaptations/device-layer/include/Weave/DeviceLayer/nRF5/WeaveDevicePlatformConfig.h
instead.

suryanshup: Moved this to the weave config file in the lock example app.
PR is here: https://github.com/openweave/openweave-nrf52840-lock-example/pull/57

src/adaptations/device-layer/NetworkTelemetryManager.cpp
suryanshup: Fixed

src/adaptations/device-layer/NetworkTelemetryManager.cpp
```diff
@@ -109,14 +110,18 @@ WEAVE_ERROR NetworkTelemetryManager::Init(void)
...
 #if WEAVE_DEVICE_CONFIG_ENABLE_WIFI_TELEMETRY
 void WiFiTelemetry::GetTelemetryStatsAndLogEvent(void)
 {
     WEAVE_ERROR err;
 
-    err = ConnectivityMgr().GetAndLogWifiStatsCounters();
-    SuccessOrExit(err);
+    if (ConnectivityMgr().IsWiFiStationProvisioned() &&
+        ConnectivityMgr().IsWiFiStationEnabled() &&
+        ConfigurationMgr().IsPairedToAccount())
```
jaylogue: I'm going to accept this, but I'm still really uncertain about the IsPairedToAccount() logic.  Remind me the rationale for not logging stats when the device isn't paired to an account; in particular, what goes wrong if we do?

turon: @suryanshup I agree that capturing the answer to Jay's question this is important.

If the proper fix is being deferred, would it make sense to create a github issue with the explanation and reference it with a `//TODO(#issue)` comment?

suryanshup: Based on an offline conversation I had with Jay, we agreed that we should remove the service provisioning check. I will fix that in a separate PR

PR #524, opened 2020-03-18
Make Software version max length in DeviceDescription configurable(default to 32 bytes)

pidarped:
This increase is to accommodate some version strings which
were getting truncated as well as an additional release track
information for some device categories for an initial factory check.


src/lib/profiles/device-description/DeviceDescription.h
```diff
@@ -125,7 +125,7 @@ class NL_DLL_EXPORT WeaveDeviceDescriptor
         kMaxSerialNumberLength                  = 32,  /**< Maximum serial number length. */
         kMaxPairingCodeLength                   = 16,  /**< Maximum pairing code length. */
         kMaxRendezvousWiFiESSID                 = 32,  /**< Maximum WiFi ESSID for Rendezvous length. */
-        kMaxSoftwareVersionLength               = 32   /**< Maximum software version length. */
+        kMaxSoftwareVersionLength               = 128   /**< Maximum software version length. */
```
jaylogue: This will blow up the size on small devices unnecessarily.  Can't we make this a configurable option with a default of 32?

pidarped: Agreed. Will post an update.

pidarped:
> Pradip, in order to make this useful, I would request that you make additional change to the default config s.t. Weave Device Manager can support the longer version out of the box. I would also request that we create an appropriate tracking items s.t. the implementations for iOS and Android that integrate Weave Device Manager include the longer version string.

Thanks. Yes, making it configurable across devices suddenly has implications on the response parsing in mobile apps. Overriding it to be sufficiently large enough for the longer version in the ios and android WeaveProjectConfig.h files should be sufficient. Will post an update.

pidarped:
> software_version needs to be updated in?
> https://github.com/openweave/openweave-schema-vendor-common/blob/01fa8b4c07cbafc5905f9fc2c1ca20fe4b7be7ca/weave/trait/description/device_identity_trait.proto#L158
> 
> ```
> /**
>  *  The Software Version property is a UTF-8 String
>  *  representing primary firmware or software currently
>  *  running on the device or product.
>  *
>  */
> string software_version                        = 7 [(wdl.prop) = {string_constraints: {max_length: 32}}];
> ```
> 
> https://github.com/openweave/openweave-schema-vendor-common/blob/01fa8b4c07cbafc5905f9fc2c1ca20fe4b7be7ca/weave/trait/peerdevices/peer_devices_trait.proto#L86
> /// Software version of the device
> string software_version = 5 [(wdl.prop) = {string_constraints: {max_length: 32}}];

Thanks for bringing these up. We could change the max_length for these as well, although am wondering if there is a pressing need for that. So, probably fine to keep the traits at the default config. The local interaction with WeaveDeviceManager needed to be changed for accommodating the factory checking requirement.

yunhanw-google:
> > software_version needs to be updated in?
> > https://github.com/openweave/openweave-schema-vendor-common/blob/01fa8b4c07cbafc5905f9fc2c1ca20fe4b7be7ca/weave/trait/description/device_identity_trait.proto#L158
> > ```
> > /**
> >  *  The Software Version property is a UTF-8 String
> >  *  representing primary firmware or software currently
> >  *  running on the device or product.
> >  *
> >  */
> > string software_version                        = 7 [(wdl.prop) = {string_constraints: {max_length: 32}}];
> > ```
> > 
> > 
> > https://github.com/openweave/openweave-schema-vendor-common/blob/01fa8b4c07cbafc5905f9fc2c1ca20fe4b7be7ca/weave/trait/peerdevices/peer_devices_trait.proto#L86
> > /// Software version of the device
> > string software_version = 5 [(wdl.prop) = {string_constraints: {max_length: 32}}];
> 
> Thanks for bringing these up. We could change the max_length for these as well, although am wondering if there is a pressing need for that. So, probably fine to keep the traits at the default config. The local interaction with WeaveDeviceManager needed to be changed for accommodating the factory checking requirement.

For DeviceIdentityTrait, it is device-owned trait, service subscribe it, then the string would overflowed.

PR #527, opened 2020-03-21
Adjust WdmClient returned error

yunhanw-google:
-- when GetTraitSchemaEngine fails, it means user does not load the
correct profile id, update err with WEAVE_ERROR_INVALID_PROFILE_ID.
-- when WdmClient receive kEvent_OnSubscriptionTerminated, pass back
mReason back to application.

### DIFF
--- a/src/device-manager/WeaveDataManagementClient.cpp
+++ b/src/device-manager/WeaveDataManagementClient.cpp
@@ -985,7 +985,7 @@ void WdmClient::ClientEventCallback(void * const aAppState, SubscriptionClient::
         WeaveLogDetail(DataManagement, "Client->kEvent_OnSubscriptionTerminated. Reason: %u, peer = 0x%" PRIX64 "\n",
                        aInParam.mSubscriptionTerminated.mReason, aInParam.mSubscriptionTerminated.mClient->GetPeerNodeId());
         pWdmClient->mpSubscriptionClient->AbortSubscription();
-        err = WEAVE_ERROR_INCORRECT_STATE;
+        err = aInParam.mSubscriptionTerminated.mReason;
         break;
     case SubscriptionClient::kEvent_OnUpdateComplete:
         if ((aInParam.mUpdateComplete.mReason == WEAVE_NO_ERROR) &&
@@ -1067,7 +1067,7 @@ WEAVE_ERROR WdmClient::NewDataSink(const ResourceIdentifier & aResourceId, uint3
     PropertyPathHandle handle = kNullPropertyPathHandle;
 
     const TraitSchemaEngine * pEngine = TraitSchemaDirectory::GetTraitSchemaEngine(aProfileId);
-    VerifyOrExit(pEngine != NULL, err = WEAVE_ERROR_INCORRECT_STATE);
+    VerifyOrExit(pEngine != NULL, err = WEAVE_ERROR_INVALID_PROFILE_ID);
 
     VerifyOrExit(mpSubscriptionClient != NULL, err = WEAVE_ERROR_INCORRECT_STATE);
 


src/device-manager/WeaveDataManagementClient.cpp
```diff
@@ -985,7 +985,7 @@ void WdmClient::ClientEventCallback(void * const aAppState, SubscriptionClient::
         WeaveLogDetail(DataManagement, "Client->kEvent_OnSubscriptionTerminated. Reason: %u, peer = 0x%" PRIX64 "\n",
                        aInParam.mSubscriptionTerminated.mReason, aInParam.mSubscriptionTerminated.mClient->GetPeerNodeId());
         pWdmClient->mpSubscriptionClient->AbortSubscription();
-        err = WEAVE_ERROR_INCORRECT_STATE;
+        err = WEAVE_ERROR_INCORRECT_SUBSCRIPTION;
```
jaylogue: This error name doesn't make sense.  How can a subscription be "incorrect"?

What exactly is happening in this case statement?  It looks like the subscription client object is telling its user that the subscription has terminated.  This event includes a reason (error) code, and optionally a status code and profile.  This information has to be passed back to the application.

yunhanw-google: Yep, you are right, I should return reason code. Just update the PR, and return reason code to application.

Thanks

src/device-manager/WeaveDataManagementClient.cpp
```diff
@@ -985,7 +985,7 @@ void WdmClient::ClientEventCallback(void * const aAppState, SubscriptionClient::
         WeaveLogDetail(DataManagement, "Client->kEvent_OnSubscriptionTerminated. Reason: %u, peer = 0x%" PRIX64 "\n",
                        aInParam.mSubscriptionTerminated.mReason, aInParam.mSubscriptionTerminated.mClient->GetPeerNodeId());
         pWdmClient->mpSubscriptionClient->AbortSubscription();
-        err = WEAVE_ERROR_INCORRECT_STATE;
+        err = aInParam.mSubscriptionTerminated.mReason;
```
jaylogue, line 988: This isn't going to be very useful for applications in the case where the subscription terminated due to an error from the publisher.  In that case, the information returned by the publisher is in the status report, and mSubscriptionTerminated.mReason will be WEAVE_ERROR_STATUS_REPORT_RECEIVED.  There needs to be a way to return the status report information back to the application.

yunhanw-google: > This isn't going to be very useful for applications in the case where the subscription terminated due to an error from the publisher. In that case, the information returned by the publisher is in the status report, and mSubscriptionTerminated.mReason will be WEAVE_ERROR_STATUS_REPORT_RECEIVED. There needs to be a way to return the status report information back to the application.

Yep, agree, mSubscriptionTerminated has already included profile id and status code, I think we can return 4 variable(error code, isStatusCodeValid, statusProfileId, statusCode) to application, and we can add new refresh callback, OnRefreshError

        struct
        {
            WEAVE_ERROR mReason;

            bool mIsStatusCodeValid;
            uint32_t mStatusProfileId;
            uint16_t mStatusCode;
            ReferencedTLVData * mAdditionalInfoPtr;
            bool mWillRetry;

            SubscriptionClient * mClient;
        } mSubscriptionTerminated;

g-coppock: I would prefer if we can follow the conventions used by WeaveDeviceManager. On the Java side, we normally get an onError callback with a Throwable as the parameter type. The concrete type of that Throwable will be either WeaveDeviceException (for profile/status errors) or WeaveDeviceManagerException (for status code errors).

jaylogue: @yunhanw-google Kevin's approach is right.  In C++ pass the reason, status profile and status code back via the onError callback.  Then in the Java wrapper code, translate these into WeaveDeviceException or WeaveDeviceManagerException.

No need to pass isStatusCodeValid to C++ callback.  If isStatusCodeValid == true, return profile and status code and set err = WEAVE_ERROR_STATUS_REPORT_RECEIVED.  If false, just return err = mReason and set profile and code to 0.

In Java wrapper, use err == WEAVE_ERROR_STATUS_REPORT_RECEIVED to decide which type of exception to return.

yunhanw-google:
I am creating onRefreshDataComplete to handle reason, status profile and status code. I would close this PR, and move #540, since this change would be related to large improvement in PR 540.

PR #529, opened 2020-03-23
Address issues in ResourceIdentifier parsing and tests.

robszewczyk:
ResourceIdentifier parsing code overly optimistically assumed a fixed
length of the actual ResourceID.  The parsing code only checked that
the buffer length was larger than the resource prefix, and after that,
it assumed the actual resource ID was a fixed string of 16 hexadecimal
digits. This pull request addresses that shortcoming.

The test code was incorrect as well, passing the size of the pointer
into the `FromString` methods as opposed to size of the buffer.  This
led to the tests succeeding on platforms with 64 bit pointers (only in
conjunction with the bug above) and failing on platforms with 32 bit
pointers.

### DIFF
--- a/src/lib/profiles/data-management/Current/ResourceIdentifier.cpp
+++ b/src/lib/profiles/data-management/Current/ResourceIdentifier.cpp
@@ -57,8 +57,8 @@ WEAVE_ERROR ResourceIdentifier::ToTLV(nl::Weave::TLV::TLVWriter & aWriter, const
     }
     else if (ResourceType == Schema::Weave::Common::RESOURCE_TYPE_DEVICE)
     {
-            err = aWriter.Put(aTag, ResourceId);
-            SuccessOrExit(err);
+        err = aWriter.Put(aTag, ResourceId);
+        SuccessOrExit(err);
     }
     else
     {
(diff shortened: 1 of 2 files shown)


src/lib/profiles/data-management/Current/ResourceIdentifier.cpp
```diff
@@ -237,6 +237,9 @@ WEAVE_ERROR ResourceIdentifier::FromString(const char * inBuffer, size_t bufferL
 
         if ((strncmp(resourceName, inBuffer, strlen(resourceName)) == 0) && inBuffer[strlen(resourceName)] == '_')
```
gerickson: Is the compiler smart enough in all cases to optimize all of the 'strlen(resourceName)' calls into one. At minimum, it might certainly make the code clearer if that was all calculated once and saved in a `const size_t`.

src/lib/profiles/data-management/Current/ResourceIdentifier.cpp
```diff
@@ -245,14 +248,22 @@ WEAVE_ERROR ResourceIdentifier::FromString(const char * inBuffer, size_t bufferL
...
 
-    memcpy(uintbuffer, inBuffer, 8);
-    uintbuffer[8] = '\0';
-    r_upper     = strtoul(uintbuffer, &endPtr, 16);
-    VerifyOrExit(strlen(endPtr) == 0, err = WEAVE_ERROR_INVALID_ARGUMENT);
-
-    memcpy(uintbuffer, inBuffer + 8, 8);
-    uintbuffer[8] = '\0';
-    r_lower     = strtoul(inBuffer + 8, &endPtr, 16);
+    if (bufferLen > 8)
```
gerickson: Nit: magic #: 8.

robszewczyk: Fixed.

src/lib/profiles/data-management/Current/ResourceIdentifier.cpp
robszewczyk: Done.

src/lib/profiles/data-management/Current/ResourceIdentifier.cpp
```diff
@@ -222,37 +222,50 @@ WEAVE_ERROR ResourceIdentifier::FromString(const char * inBuffer, size_t bufferL
...
-
-    memcpy(uintbuffer, inBuffer + 8, 8);
-    uintbuffer[8] = '\0';
-    r_lower     = strtoul(inBuffer + 8, &endPtr, 16);
+    if (bufferLen > hexDigitsInInt32)
+    {
+        memcpy(uintbuffer, inBuffer, hexDigitsInInt32);
+        uintbuffer[hexDigitsInInt32] = '\0';
+        r_upper                      = strtoul(uintbuffer, &endPtr, 16);
+        VerifyOrExit(strlen(endPtr) == 0, err = WEAVE_ERROR_INVALID_ARGUMENT);
```
jaylogue: Seems a little indirect.  Perhaps we should replace the expression with endPtr[0] == 0?

src/lib/profiles/data-management/Current/ResourceIdentifier.cpp
```diff
@@ -222,37 +222,50 @@ WEAVE_ERROR ResourceIdentifier::FromString(const char * inBuffer, size_t bufferL
...
+    else
+    {
+        r_upper = 0;
+    }
+    memset(uintbuffer, 0, sizeof(uintbuffer));
+    memcpy(uintbuffer, inBuffer, bufferLen > hexDigitsInInt32 ? hexDigitsInInt32 : bufferLen);
+    r_lower = strtoul(uintbuffer, &endPtr, 16);
     VerifyOrExit(strlen(endPtr) == 0, err = WEAVE_ERROR_INVALID_ARGUMENT);
 
     ResourceId   = ((uint64_t) r_upper) << 32 | ((uint64_t) r_lower);
```
jaylogue: I wonder if we really need to accommodate a variable length hex string.

The Weave / Phoenix Resource Ids document says: _The string form of the object id is always 16 hexadecimal digits, uppercase._  But perhaps this part of the spec was ignored.

src/test-apps/TestResourceIdentifier.cpp
```diff
@@ -410,12 +410,12 @@ static void CheckStringSerDes(nlTestSuite * inSuite, void * inContext)
...
-    const char * resource_self_str          = "RESERVED_DEVICE_SELF";
-    const char * resource_uninitialized_str = "RESERVED_NOT_SPECIFIED";
-    const char * resource_unknown_type_str  = "(C001)_18B4300000000001";
-    const char * resource_unknown_str       = "WIDGET_18B4300000000001";
+    const char resource_str[]               = "DEVICE_18B4300000000001";
+    const char resource2_str[]              = "USER_18B4300000000001";
+    const char resource_self_str[]          = "RESERVED_DEVICE_SELF";
+    const char resource_uninitialized_str[] = "RESERVED_NOT_SPECIFIED";
+    const char resource_unknown_type_str[]  = "(C001)_18B4300000000001";
+    const char resource_unknown_str[]       = "WIDGET_18B4300000000001";
```
jaylogue: Maybe add a test for object id < 16 hex characters?

robszewczyk: Added tests.  I've also updated the parsing code, it was incorrect previously for device IDs shorter than 8 hex characters.

PR #531, opened 2020-03-24
Cleanup/enforce formatting

robszewczyk:
This is a bit of a bit request for comments / opinions.  I've followed up on an earlier conversation with @turon on what OpenWeave code would look like if we started enforcing automated formatting.  I've noted that there are a number of places in Weave where the textual layout of the file is particularly carefully though out and geared towards readability of the source file and associated inline documentation.  This comes across particularly strongly in:

* error string formatting
* enum formatting with associated brief doxygen docs
* class / struct member documentation
* documentation blocks on consecutive and related structures, enums, and classes.

This pull request is a sweep through much of the openweave source code for places where the block oriented formatting significantly adds to the reliability.  I did not attempt as yet the sweep through `tools` and `test-apps`. With the changes in this PR, the clang-format produces the resulting source code layout as shown in https://github.com/robszewczyk/openweave-core/tree/cleanup/reformatted-source. If the automatically formatted source code results (broadly speaking) meet with the aesthetic and usability standards, I'd like to move in the direction of enforcing a uniform layout throughout the codebase.


src/adaptations/device-layer/include/Weave/DeviceLayer/ConnectivityManager.h
```diff
@@ -207,14 +207,24 @@ class ConnectivityManager
...
+    /**
+     * Interval at which the device polls its parent Thread router when
+     * when there are NO active Weave exchanges in progress. Only meaningful
+     * when the device is acting as a sleepy end node.
+     */
+    uint32_t InactivePollingIntervalMS;
+
+    /**
+     * Clear the structure, setting all members to 0.
+     */
```
jaylogue: What is the point of enforcing prefix documentation for inline definitions?  This is particularly onerous on readability.  

I see a lot of places in this PR where the requirement is suppressed.  If we don't like the way it looks, then why would we turn it on in the first place?

src/lib/support/verhoeff/Verhoeff10.cpp
```diff
@@ -30,6 +30,7 @@
 
 #ifndef VERHOEFF10_NO_MULTIPLY_TABLE
 
+// clang-format off
```
jaylogue: What motivates turning off formatting here?

src/adaptations/device-layer/include/Weave/DeviceLayer/ConnectivityManager.h
robszewczyk: This was one of the first files I changed, and I was still  evolving a style. In a few places earlier on I thought the right path forward was to switch to prefix and rely on the clang format down the road.  As the set of changes evolved, I've started using the `// clang-format off` directive more easily and I am happy to make those changes.

src/lib/profiles/network-provisioning/NetworkInfo.h
```diff
@@ -53,24 +53,26 @@ class NL_DLL_EXPORT NetworkInfo
     NetworkInfo();
     ~NetworkInfo();
 
-    ::nl::Weave::Profiles::NetworkProvisioning::NetworkType
-          NetworkType;              /**< The type of network. */
-    int64_t NetworkId;              /**< The network id assigned to the network by the device, -1 if not specified. */
+    /** The type of network. */
```
robszewczyk: This was an example where the existing infix notation did not work well.  I think the infix notation could have been fixed if we did not use fully qualified types, but as is the `type varName;` was just too long, especially with the `type` spilling into the infix doxygen. I'd be happy to refactor this back to infix with the shortened types, I think it is possible.

src/lib/support/verhoeff/Verhoeff10.cpp
robszewczyk: Clang-format will attempt to reformat the `sMultiplyTable` below to fill the allotted columns (I believe the fill line is set at around 132 columns).  This loses the readability of the table that is laid out to match Verhoeff encoding base.

PR #532, opened 2020-03-25
Fix printf buffer overflow

kghost:
printf format %*s is not limiting the string size, here we should use
%.*s to limit the output length.

Change-Id: Iea875e06e0df786b9a0a7f03f1a7bcaba2a1b9b4


jaylogue:
@gerickson -- Two possibilities: 1) This code isn't being analysed because its in the device layer, which isn't included in the standalone build, 2) The analyzers may not be aware that this is a printf-style macro.

PR #538, opened 2020-03-31
[Thread] Support Weave thread border routing

erjiaqing:
This PR adds related openthread api call to WarmSupport in LwIP.

- When add or removing thread addresses, will also add weave thread subnet mesh prefix to thread stack.
- Add fabric prefix to thread border router.

### DIFF
--- a/src/adaptations/device-layer/LwIP/WarmSupport.cpp
+++ b/src/adaptations/device-layer/LwIP/WarmSupport.cpp
@@ -36,6 +36,9 @@
 #include <Weave/DeviceLayer/ThreadStackManager.h>
 #include <Weave/DeviceLayer/OpenThread/OpenThreadUtils.h>
 #include <openthread/ip6.h>
+#if WARM_CONFIG_SUPPORT_THREAD_ROUTING
+#include <openthread/border_router.h>
+#endif // WARM_CONFIG_SUPPORT_THREAD_ROUTING
 #endif // WARM_CONFIG_SUPPORT_THREAD
 
 
@@ -55,7 +58,10 @@ namespace Platform {
 
 WEAVE_ERROR Init(WarmFabricStateDelegate * inFabricStateDelegate)
 {
-    // Nothing to do.
+#if WARM_CONFIG_SUPPORT_THREAD_ROUTING
+    // Inform Warm of current state of thread router.
+    Warm::ThreadRoutingStateChange(kInterfaceStateUp);
+#endif // WARM_CONFIG_SUPPORT_THREAD_ROUTING
     return WEAVE_NO_ERROR;
 }
 
@@ -304,36 +310,18 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
 {
     otError otErr;
     otNetifAddress otAddress;
-    otBorderRouterConfig brConfig;
 
     memset(&otAddress, 0, sizeof(otAddress));
-
-    ThreadStackMgrImpl().LockThreadStack();
-
     otAddress.mAddress = ToOpenThreadIP6Address(inAddress);
     otAddress.mPrefixLength = 64;
     otAddress.mValid = true;
     otAddress.mPreferred = true;
 
-    brConfig.mConfigure = false;
-    brConfig.mDefaultRoute = false;
-    brConfig.mDhcp = false;
-    brConfig.mOnMesh = true;
-    brConfig.mPreference = 0;
-    brConfig.mPreferred = true;
-    brConfig.mPrefix.mPrefix = otAddress.mAddress;
-    brConfig.mPrefix.mLength = 64;
-    brConfig.mRloc16 = otThreadGetRloc16(ThreadStackMgrImpl().OTInstance());
-    brConfig.mSlaac = false;
-    brConfig.mStable = true;
+    ThreadStackMgrImpl().LockThreadStack();
 
     if (inAdd)
     {
         otErr = otIp6AddUnicastAddress(ThreadStackMgrImpl().OTInstance(), &otAddress);
-        if (otErr == OT_ERROR_NONE)
-        {
-            otErr = otBorderRouterAddOnMeshPrefix(ThreadStackMgrImpl().OTInstance(), &brConfig);
-        }
     }
     else
     {
@@ -346,13 +334,6 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
         if (otErr == OT_ERROR_NOT_FOUND)
         {
             WeaveLogProgress(DeviceLayer, "otIp6RemoveUnicastAddress: already removed");
-            otErr = OT_ERROR_NONE;
-        }
-
-        otErr = otBorderRouterRemoveOnMeshPrefix(ThreadStackMgrImpl().OTInstance(), &brConfig.mPrefix);
-        if (otErr == OT_ERROR_NOT_FOUND)
-        {
-            WeaveLogProgress(DeviceLayer, "otBorderRouterRemoveOnMeshPrefix: already removed");
             otErr = OT_ERROR_NONE;
         }
     }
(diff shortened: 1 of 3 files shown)


src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -358,27 +383,66 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
 
 #if WARM_CONFIG_SUPPORT_THREAD_ROUTING
 
-#error "Weave Thread router support not implemented"
-
 PlatformResult StartStopThreadAdvertisement(InterfaceType inInterfaceType, const Inet::IPPrefix &inPrefix, bool inStart)
 {
-    // TODO: implement me
+    if (inInterfaceType != kInterfaceTypeThread) { return kPlatformResultFailure; }
```
yunhanw-google, line 388: The codestyle is not consistent with others in weave, please address it.
Please use one error return statement for function.
please use different error for both if.

src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -358,27 +383,66 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
...
 {
-    // TODO: implement me
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+    otError otErr;
+    otExternalRouteConfig routeConfig;
+
+    int ot_priority = OT_ROUTE_PREFERENCE_HIGH;
+    switch (inPriority)
+    {
+        case kRoutePriorityLow:
```
yunhanw-google, line 409: switch should align with  case, see this example
https://github.com/openweave/openweave-core/blob/master/src/lib/profiles/data-management/Current/SubscriptionClient.cpp#L171

src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -358,27 +383,66 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
...
+            ot_priority = OT_ROUTE_PREFERENCE_LOW;
+            break;
+        case kRoutePriorityMedium:
+            ot_priority = OT_ROUTE_PREFERENCE_MED;
+            break;
+        case kRoutePriorityHigh:
+            ot_priority = OT_ROUTE_PREFERENCE_HIGH;
+            break;
```
yunhanw-google, line 417: what is the default case? goto exit?

src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -304,18 +304,36 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
...
+    brConfig.mRloc16 = otThreadGetRloc16(ThreadStackMgrImpl().OTInstance());
+    brConfig.mSlaac = false;
+    brConfig.mStable = true;
 
     if (inAdd)
     {
         otErr = otIp6AddUnicastAddress(ThreadStackMgrImpl().OTInstance(), &otAddress);
+        if (otErr == OT_ERROR_NONE)
+        {
+            otErr = otBorderRouterAddOnMeshPrefix(ThreadStackMgrImpl().OTInstance(), &brConfig);
```
jaylogue, line 335: This doesn't follow the WARM spec.  In particular, a device that isn't built to be a border router, or isn't currently operating as a border router, shouldn't be adding an on-mesh prefix.  This behavior is supposed to happen in StartStopThreadAdvertisement().

src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -358,27 +383,66 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
...
+        otErr = otBorderRouterAddRoute(ThreadStackMgrImpl().OTInstance(), &routeConfig);
+    }
+    else
+    {
+        otErr = otBorderRouterRemoveRoute(ThreadStackMgrImpl().OTInstance(), &routeConfig.mPrefix);
+    }
+
+    err = MapOpenThreadError(otErr);
+    ThreadStackMgrImpl().UnlockThreadStack();
+    return (err == WEAVE_NO_ERROR) ? kPlatformResultSuccess : kPlatformResultFailure;
```
jaylogue, line 440: If this fails, how will the developer know what error happened?

src/adaptations/device-layer/LwIP/WarmSupport.cpp
erjiaqing: In wpantund, https://github.com/openthread/wpantund/blob/1f656564af7b0efb137bbada4c6c9ae7d99adf26/src/wpantund/NCPInstanceBase-Addresses.cpp#L635-L667

When a unicast address is added to interface, a corresponding OnMeshPrefix is also added to the interface. Should we follow this behavior?

Also, when trying to add a /48 OnMeshPrefix in StartStopThreadAdvertisement, which is triggered at https://github.com/openweave/openweave-core/blob/2bcd0349d0fd11f73fa715405308651746871511/src/warm/WarmCore.cpp#L835-L842, thread will drop the packets, but a /64 on mesh prefix works fine.

src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -358,27 +364,104 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
...
+            otErr = OT_ERROR_NONE;
+        }
+    }
+
+    ThreadStackMgrImpl().UnlockThreadStack();
+
+    err = MapOpenThreadError(otErr);
+
+exit:
+    return (err == WEAVE_NO_ERROR) ? kPlatformResultSuccess : kPlatformResultFailure;
```
jaylogue, line 410: If this fails, how will the developer know what happened?

src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -358,27 +364,104 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
...
-#error "Weave Thread router support not implemented"
-
 PlatformResult StartStopThreadAdvertisement(InterfaceType inInterfaceType, const Inet::IPPrefix &inPrefix, bool inStart)
 {
-    // TODO: implement me
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+    otError otErr;
+    otBorderRouterConfig brConfig;
+
+    if (inInterfaceType != kInterfaceTypeThread) { err = WEAVE_ERROR_INVALID_ARGUMENT; }
```
jaylogue, line 373: Use VerifyOrExit().

src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -358,27 +364,104 @@ PlatformResult AddRemoveThreadAddress(InterfaceType inInterfaceType, const Inet:
...
+    if ((inPrefix.Length & 7) != 0) { err = WEAVE_ERROR_INVALID_ADDRESS; }
```
jaylogue, line 374: Use VerifyOrExit().

src/adaptations/device-layer/LwIP/WarmSupport.cpp
```diff
@@ -55,7 +58,10 @@ namespace Platform {
 
 WEAVE_ERROR Init(WarmFabricStateDelegate * inFabricStateDelegate)
 {
-    // Nothing to do.
+#if WARM_CONFIG_SUPPORT_THREAD_ROUTING
+    // Inform Warm of current state of thread router.
+    Warm::ThreadRoutingStateChange(kInterfaceStateUp);
```
jaylogue, line 63: I don't think this should be hard-coded.   Even though WARM_CONFIG_SUPPORT_THREAD_ROUTING is 1, a device may not always be operating as a Thread router, based on configuration set by the application.  For example, the application could call `ThreadStackMgr().SetThreadDeviceType(kThreadDeviceType_FullEndDevice)` to force the device to be an end-device instead of a router.

I think the right place to call Warm::ThreadRoutingStateChange() is in [GenericConnectivityManagerImpl_Thread<ImplClass>::_OnPlatformEvent()](https://github.com/openweave/openweave-core/blob/47d46e76e643c67d226f866ad2f90bec5b49fc07/src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericConnectivityManagerImpl_Thread.ipp#L41). 

This method should watch for kThreadStateChange events where the RoleChanged flag == true.  Whenever the Thread role changes, the code should call ThreadStackMgrImpl().GetThreadDeviceType() to determine whether the device is currently operating as a router (kThreadDeviceType_Router) and then call Warm::ThreadRoutingStateChange() with kInterfaceStateUp or kInterfaceStateDown accordingly.

src/warm/WarmCore.cpp
```diff
@@ -836,7 +836,7 @@ static PlatformResult ThreadAdvertisementAction(ActionType inAction, bool inActi
 {
     Inet::IPPrefix prefix;
 
-    MakePrefix(inGlobalId, 0, kGlobalULAPrefixLength, prefix);
+    MakePrefix(inGlobalId, kWeaveSubnetId_ThreadMesh, kThreadULAAddressPrefixLength, prefix);
```
jaylogue, line 839: Now I understand your comment about packets being dropped when prefixlen != 64. This appears to be a bug in Warm that has gone unnoticed for a long time.

src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericConnectivityManagerImpl_Thread.ipp
```diff
@@ -62,6 +66,26 @@ void GenericConnectivityManagerImpl_Thread<ImplClass>::_OnPlatformEvent(const We
...
+
+#if WARM_CONFIG_SUPPORT_THREAD_ROUTING
+    const bool threadRoleChanged = (event->Type == DeviceEventType::kThreadStateChange &&
+                                    event->ThreadStateChange.RoleChanged);
+    if (threadRoleChanged)
+    {
+        otDeviceRole role;
+        ThreadStackMgrImpl().LockThreadStack();
+        role = otThreadGetDeviceRole(ThreadStackMgrImpl().OTInstance());
+        ThreadStackMgrImpl().UnlockThreadStack();
```
jaylogue, line 78: This class should not call OT directly.  Instead, it should call a method on ThreadStackMgrImpl() which then calls OT.  That way all OT-specific code is kept within the ThreadStackManager object.

I think the right API to call here is ThreadStackMgrImpl().GetThreadDeviceType(). This will tell us if the device has been configured to operate as a router.

jaylogue:
One small comment.  Otherwise this looks great.

src/adaptations/device-layer/include/Weave/DeviceLayer/internal/GenericConnectivityManagerImpl_Thread.ipp
```diff
@@ -29,6 +29,10 @@
 #include <Weave/DeviceLayer/internal/GenericConnectivityManagerImpl_Thread.h>
 #include <Warm/Warm.h>
 
+#if WARM_CONFIG_SUPPORT_THREAD_ROUTING
+#include <openthread/thread.h>
+#endif // WARM_CONFIG_SUPPORT_THREAD_ROUTING
```
jaylogue, line 34: We can remove this, right?

PR #540, opened 2020-04-01
Weave Data Management mobile support improvements

yunhanw-google:
-- Add onFlushUpdateComplete callback to notify failed path results to
application, for each flushed path, if systemError is WEAVE_ERROR_STATUS_REPORT_RECEIVED,
application would receive the device Error Exception including status
report, particular path and dataSink object, otherwise, application would receive the
local error exception, particular path and dataSink object. Add objective C, Java, and Python implementation for the above exceptions.
-- Add DeleteData API so that application can delete data in particular
path.
-- Add length check for GetString and GetStringArray.
-- When GetTraitSchemaEngine fails, it means user does not load the
correct profile id, update err with WEAVE_ERROR_INVALID_PROFILE_ID
-- When refresh fails, we use onError to pass the status report and
error.
-- Fix some style issues.
-- Update onFlushUpdateComplete and DeleteData APIs for objective C, Java, Python
-- Clean redundant code in NLGenericTraitUpdatableDataSink
-- Adjust the signature for getString, getBytes, and getStringArray, and
pass val as reference, and return error code for objectice C support.
-- Fix TLV Dump for empty string.

### DIFF
--- a/src/device-manager/WeaveDataManagementClient.cpp
+++ b/src/device-manager/WeaveDataManagementClient.cpp
@@ -498,7 +498,6 @@ WEAVE_ERROR GenericTraitUpdatableDataSink::GetBoolean(const char * apPath, bool 
     nl::Weave::TLV::TLVReader reader;
     PacketBuffer * pMsgBuf                = NULL;
     PropertyPathHandle propertyPathHandle = kNullPropertyPathHandle;
-    ;
     std::map<PropertyPathHandle, PacketBuffer *>::iterator it;
 
     err = GetSchemaEngine()->MapPathToHandle(apPath, propertyPathHandle);
@@ -537,7 +536,6 @@ WEAVE_ERROR GenericTraitUpdatableDataSink::GetBytes(const char * apPath, BytesDa
     nl::Weave::TLV::TLVReader reader;
     PacketBuffer * pMsgBuf                = NULL;
     PropertyPathHandle propertyPathHandle = kNullPropertyPathHandle;
-    ;
     std::map<PropertyPathHandle, PacketBuffer *>::iterator it;
 
     err = GetSchemaEngine()->MapPathToHandle(apPath, propertyPathHandle);
@@ -558,8 +556,12 @@ WEAVE_ERROR GenericTraitUpdatableDataSink::GetBytes(const char * apPath, BytesDa
     SuccessOrExit(err);
 
     apBytesData->mDataLen = reader.GetLength();
-    err                   = reader.GetDataPtr(apBytesData->mpDataBuf);
-    SuccessOrExit(err);
+    WeaveLogProgress(DataManagement, "GetBytes with length %d", apBytesData->mDataLen);
+    if (apBytesData->mDataLen != 0)
+    {
+        err = reader.GetDataPtr(apBytesData->mpDataBuf);
+        SuccessOrExit(err);
+    }
 
 exit:
     WeaveLogFunctError(err);
@@ -618,7 +620,6 @@ WEAVE_ERROR GenericTraitUpdatableDataSink::IsNull(const char * apPath, bool & aI
     nl::Weave::TLV::TLVReader reader;
     PacketBuffer * pMsgBuf                = NULL;
     PropertyPathHandle propertyPathHandle = kNullPropertyPathHandle;
-    ;
     std::map<PropertyPathHandle, PacketBuffer *>::iterator it;
 
     err = GetSchemaEngine()->MapPathToHandle(apPath, propertyPathHandle);
@@ -658,7 +659,6 @@ WEAVE_ERROR GenericTraitUpdatableDataSink::GetStringArray(const char * apPath, s
     nl::Weave::TLV::TLVReader reader;
     PacketBuffer * pMsgBuf                = NULL;
     PropertyPathHandle propertyPathHandle = kNullPropertyPathHandle;
-    ;
     TLVType OuterContainerType;
     std::map<PropertyPathHandle, PacketBuffer *>::iterator it;
 
@@ -686,8 +686,13 @@ WEAVE_ERROR GenericTraitUpdatableDataSink::GetStringArray(const char * apPath, s
     {
         int length               = reader.GetLength();
         const uint8_t * pDataBuf = NULL;
-        err                      = reader.GetDataPtr(pDataBuf);
-        SuccessOrExit(err);
+        WeaveLogProgress(DataManagement, "GetStringArray with length %d", length);
+        if (length != 0)
+        {
+            err = reader.GetDataPtr(pDataBuf);
+            SuccessOrExit(err);
+        }
+
         std::string val((char *) pDataBuf, length);
         aValueVector.push_back(val);
     }
@@ -734,6 +739,36 @@ WEAVE_ERROR GenericTraitUpdatableDataSink::Get(const char * apPath, T & aValue)
 
     err = reader.Get(aValue);
     SuccessOrExit(err);
+
+exit:
+    WeaveLogFunctError(err);
+    return err;
+}
+
+WEAVE_ERROR GenericTraitUpdatableDataSink::DeleteData(const char * apPath)
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+    nl::Weave::TLV::TLVReader reader;
+    PacketBuffer * pMsgBuf                = NULL;
+    PropertyPathHandle propertyPathHandle = kNullPropertyPathHandle;
+    std::map<PropertyPathHandle, PacketBuffer *>::iterator it;
+
+    err = GetSchemaEngine()->MapPathToHandle(apPath, propertyPathHandle);
+    SuccessOrExit(err);
+
+    it = mPathTlvDataMap.find(propertyPathHandle);
+    VerifyOrExit(it != mPathTlvDataMap.end(), err = WEAVE_ERROR_INVALID_TLV_TAG);
+
+    if (NULL != it->second)
+    {
+        PacketBuffer::Free(it->second);
+        it->second = NULL;
+        WeaveLogProgress(DataManagement, "Deleted data in mPathTlvDataMap for path %s", apPath);
+    }
+
+    mPathTlvDataMap.erase(it);
+
+    err = ClearUpdated(GetSubscriptionClient(), propertyPathHandle);
 
 exit:
     WeaveLogFunctError(err);
@@ -903,9 +938,9 @@ void WdmClient::ClientEventCallback(void * const aAppState, SubscriptionClient::
 {
     WEAVE_ERROR err              = WEAVE_NO_ERROR;
     WdmClient * const pWdmClient = reinterpret_cast<WdmClient *>(aAppState);
-
-    OpState savedOpState = pWdmClient->mOpState;
-    WeaveLogDetail(DataManagement, "WDM ClientEventCallback: current op is, %d", savedOpState);
+    DeviceStatus * deviceStatus  = NULL;
+    OpState savedOpState         = pWdmClient->mOpState;
+    WeaveLogDetail(DataManagement, "WdmClient ClientEventCallback: current op is, %d, event is %d", savedOpState, aEvent);
 
     switch (aEvent)
     {
@@ -984,31 +1019,46 @@ void WdmClient::ClientEventCallback(void * const aAppState, SubscriptionClient::
     case SubscriptionClient::kEvent_OnSubscriptionTerminated:
         WeaveLogDetail(DataManagement, "Client->kEvent_OnSubscriptionTerminated. Reason: %u, peer = 0x%" PRIX64 "\n",
                        aInParam.mSubscriptionTerminated.mReason, aInParam.mSubscriptionTerminated.mClient->GetPeerNodeId());
-        pWdmClient->mpSubscriptionClient->AbortSubscription();
-        err = WEAVE_ERROR_INCORRECT_STATE;
+        VerifyOrExit(kOpState_RefreshData == savedOpState, err = WEAVE_ERROR_INCORRECT_STATE);
+        deviceStatus                  = new DeviceStatus();
+        deviceStatus->StatusProfileId = aInParam.mSubscriptionTerminated.mStatusProfileId;
+        deviceStatus->StatusCode      = aInParam.mSubscriptionTerminated.mStatusCode;
+        deviceStatus->SystemErrorCode = aInParam.mSubscriptionTerminated.mReason;
+        err                           = aInParam.mSubscriptionTerminated.mReason;
         break;
     case SubscriptionClient::kEvent_OnUpdateComplete:
+        VerifyOrExit(kOpState_FlushUpdate == savedOpState, err = WEAVE_ERROR_INCORRECT_STATE);
         if ((aInParam.mUpdateComplete.mReason == WEAVE_NO_ERROR) &&
             (nl::Weave::Profiles::kWeaveProfile_Common == aInParam.mUpdateComplete.mStatusProfileId) &&
             (nl::Weave::Profiles::Common::kStatus_Success == aInParam.mUpdateComplete.mStatusCode))
         {
-            WeaveLogDetail(DataManagement, "Update: path result: success");
+            WeaveLogDetail(DataManagement, "Update: path result: success, tdh %" PRIu16 ", ph %" PRIu16 ".",
+                           aInParam.mUpdateComplete.mTraitDataHandle, aInParam.mUpdateComplete.mPropertyPathHandle);
         }
         else
         {
-            WeaveLogDetail(DataManagement, "Update: path failed: %s, %s, tdh %" PRIu16 ", will %sretry, discard failed change",
+
+            WeaveLogDetail(DataManagement,
+                           "Update: path failed: %s, %s, tdh %" PRIu16 ", ph %" PRIu16 ", %s, discard failed change",
                            ErrorStr(aInParam.mUpdateComplete.mReason),
                            nl::StatusReportStr(aInParam.mUpdateComplete.mStatusProfileId, aInParam.mUpdateComplete.mStatusCode),
-                           aInParam.mUpdateComplete.mTraitDataHandle, aInParam.mUpdateComplete.mWillRetry ? "" : "not ");
+                           aInParam.mUpdateComplete.mTraitDataHandle, aInParam.mUpdateComplete.mPropertyPathHandle,
+                           aInParam.mUpdateComplete.mWillRetry ? "will retry" : "will not retry");
+
+            err = pWdmClient->UpdateFailedPathResults(
+                pWdmClient, aInParam.mUpdateComplete.mTraitDataHandle, aInParam.mUpdateComplete.mPropertyPathHandle,
+                aInParam.mUpdateComplete.mReason, aInParam.mUpdateComplete.mStatusProfileId, aInParam.mUpdateComplete.mStatusCode);
         }
 
         break;
     case SubscriptionClient::kEvent_OnNoMorePendingUpdates:
-        // TODO: notify application with all status for updated paths
         WeaveLogDetail(DataManagement, "Update: no more pending updates");
         VerifyOrExit(kOpState_FlushUpdate == savedOpState, err = WEAVE_ERROR_INCORRECT_STATE);
         pWdmClient->mpSubscriptionClient->DiscardUpdates();
-        pWdmClient->mOnComplete.General(pWdmClient->mpContext, pWdmClient->mpAppReqState);
+        pWdmClient->mOnComplete.FlushUpdate(pWdmClient->mpContext, pWdmClient->mpAppReqState,
+                                            pWdmClient->mFailedFlushPathStatus.size(), pWdmClient->mFailedFlushPathStatus.data());
+        pWdmClient->mFailedFlushPathStatus.clear();
+        pWdmClient->mFailedPaths.clear();
         pWdmClient->mpContext = NULL;
         pWdmClient->ClearOpState();
         break;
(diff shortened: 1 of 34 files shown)


src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -2915,6 +2915,32 @@ WEAVE_ERROR SubscriptionClient::SetUpdated(TraitUpdatableDataSink * aDataSink, P
...
+    LockUpdateMutex();
+
+    schemaEngine = aDataSink->GetSchemaEngine();
+
+    err = mDataSinkCatalog->Locate(aDataSink, dataHandle);
+    SuccessOrExit(err);
+
+    isTraitInstanceInUpdate = mPendingUpdateSet.IsTraitPresent(dataHandle);
+
+    VerifyOrExit(isTraitInstanceInUpdate, WeaveLogDetail(DataManagement, "trait %d is not in update pending set, skip ClearUpdated", dataHandle));
```
jaylogue, line 2935: Do not put other actions in the second parameter to VeriryOrExit().  The second argument should only contain assignments to return values.

src/device-manager/WeaveDataManagementClient.cpp
```diff
@@ -988,27 +1013,31 @@ void WdmClient::ClientEventCallback(void * const aAppState, SubscriptionClient::
...
             (nl::Weave::Profiles::kWeaveProfile_Common == aInParam.mUpdateComplete.mStatusProfileId) &&
             (nl::Weave::Profiles::Common::kStatus_Success == aInParam.mUpdateComplete.mStatusCode))
         {
-            WeaveLogDetail(DataManagement, "Update: path result: success");
+            WeaveLogDetail(DataManagement, "Update: path result: success, tdh %" PRIu16 ", ph %" PRIu16 ".", aInParam.mUpdateComplete.mTraitDataHandle, aInParam.mUpdateComplete.mPropertyPathHandle);
         }
         else
         {
-            WeaveLogDetail(DataManagement, "Update: path failed: %s, %s, tdh %" PRIu16 ", will %sretry, discard failed change",
+            WeaveLogDetail(DataManagement, "Update: path failed: %s, %s, tdh %" PRIu16 ", ph %" PRIu16 ",will %s retry, discard failed change",
```
g-coppock, line 1025: The "will %s retry" formatting section will lead to a double space between words. Consider just subbing the whole string? e.g.

"%s" -> "will retry" : "will not retry"

src/device-manager/java/src/nl/Weave/DataManagement/WdmClient.java
```diff
@@ -67,7 +67,7 @@
 
     public interface CompletionHandler
     {
-        void onFlushUpdateComplete();
+        void onFlushUpdateComplete(int[] pathResults);
```
g-coppock, line 70: What are these results, exactly? It's unclear how this array should be interpreted.

yunhanw-google: For example,
In your one flush operation,
        localSettingsTrait.set("/1", "en-US");
        testCTrait.set("/1", false);
        testCTrait.setSigned("/2", 15);
        testCTrait.setUnsigned("/3/1", -5);
        testCTrait.set("/3/2", false);
        testCTrait.setUnsigned("/4", -5);
        TestResult = null;
        mockWdmClient.beginFlushUpdate();
        ExpectSuccess("FlushUpdate");
If all succeeds, you pathResults would be 
{0, 0, 0, 0, 0, 0}, which is exactly matching with your flushed path, and it is ErrorList for your flushed path. 
If the first fails, you pathResults would be 
{4041, 0, 0, 0, 0, 0},  4041 is the invalid profile id error code, it would tell you the trait may not be implemented in device side.
In application, you need to record the list regarding trait and tag path, you can use deleteData to clear the first path, or you can use clear to clear the whole trait.

g-coppock: What is the error code space? i.e. where is 4041 defined?

src/device-manager/WeaveDataManagementClient.cpp
yunhanw-google: yep

src/lib/profiles/data-management/Current/SubscriptionClient.cpp
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WdmClient.java
yunhanw-google: it should be weave error code, in my standalone mock test, it starts from 4000

yunhanw-google:
@jaylogue For update flushed path result, reason code should be enough? do you think we need to tell app the profile id + status code?, thanks

yunhanw-google:
> @jaylogue For update flushed path result, reason code should be enough? do you think we need to tell app the profile id + status code?, thanks

In Java, just return an array of Throwables (and have WeaveDeviceException and WeaveDeviceManagerException, 
either WeaveDeviceManagerException (uses the Weave error code) or WeaveDeviceException (uses profile ID and status code

or just 3 arrays(reason code, profile id and status code)

yunhanw-google:
> @jaylogue For update flushed path result, reason code should be enough? do you think we need to tell app the profile id + status code?, thanks

When error is  WEAVE_ERROR_STATUS_REPORT_RECEIVED, application may need more useful info, for exmaple, profile id + status code

yunhanw-google:
I am moving the effort from https://github.com/openweave/openweave-core/pull/527 to this PR since some large code refactor is happening here.

yunhanw-google:
Just validated it in python, objective c and java, please help further review. thanks

g-coppock:
Just making sure you'd seen the comment thread on #527, about using the specific exception types rather than an array of integer codes: https://github.com/openweave/openweave-core/pull/527#discussion_r403393258

yunhanw-google:
> Just making sure you'd seen the comment thread on #527, about using the specific exception types rather than an array of integer codes: [#527 (comment)](https://github.com/openweave/openweave-core/pull/527#discussion_r403393258)

Yes, I see that, I think it can be handled in application layer. For example, if we have 100 paths in this flush operation, then some of them are WEAVE_ERROR_STATUS_REPORT_RECEIVED, i am not sure how to use one call to pass some WeaveDeviceException (for profile/status errors), and some  WeaveDeviceManagerExceptions(notes that I need to keep the sequence, they must be in one array)

yunhanw-google:
@jaylogue @g-coppock 
For java, it seems it would be a array of jobject, something link [WeaveDeviceException, WeaveDeviceException, WeaveDeviceManagerExceptions, WeaveDeviceException] in jni layer, then java can check particular exception from this array.  note WeaveDeviceException (for profile/status errors) or WeaveDeviceManagerException (for errors)
In objective C, we cannot pass the mixed type of array, the best effort is to use NSArray to contruct only one type of array for NLProfileStatusError.
In python, from c++ to python, we cannot pass the mixed type of array, the best effort is just array of DeviceError.

The unified way is just 3 arrays for error code, profileid, and statusCode, or just array of statusReport

yunhanw-google:
offline chat with Jay, and addressing the comments, thanks

src/device-manager/java/Makefile.am
```diff
@@ -110,6 +110,8 @@ WeaveDeviceManager_jar_JAVA_SRCS                   =        \
     nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java \
     nl/Weave/DataManagement/WdmClientFactory.java           \
     nl/Weave/DataManagement/ResourceIdentifier.java         \
+	nl/Weave/DataManagement/WdmClientFlushUpdateException.java          \
```
g-coppock, line 113: nit: looks like the indentation is off here -- should be 2 spaces to the left

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
```diff
@@ -333,7 +340,7 @@ private void onRefreshDataComplete()
     }
 
     private void ensureNotClosed() {
-      if (mTraitInstancePtr == 0) {
+      if (mTraitInstancePtr== 0) {
```
g-coppock, line 343: Revert?

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientFlushUpdateDeviceException.java
```diff
@@ -0,0 +1,53 @@
...
+package nl.Weave.DataManagement;
+import nl.Weave.DeviceManager.WeaveDeviceException;
+
+/** Represents a exception that occurred on a Weave device during the WdmClient FlushUpdate operation.
+ */
+public class WdmClientFlushUpdateDeviceException extends WeaveDeviceException {
+    public WdmClientFlushUpdateDeviceException () {}
+
+    public WdmClientFlushUpdateDeviceException (int statusCode, int profileId, int sysErrorCode, String description, String pathStr, long dataSink) {
+        super(statusCode, profileId, sysErrorCode, description != null ? description : String.format("Error Code %d, Profile Id %d, StatusCode%d, path %s", sysErrorCode, profileId, statusCode, pathStr));
```
g-coppock, line 28: Space between "StatusCode" and "%d"?

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientFlushUpdateDeviceException.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public GenericTraitUpdatableDataSink getDataSink(WdmClient wdmClient)
+    {
+        return wdmClient.getDataSink(dataSinkPtr);
+    }
+
+    /**
+     * A flushed path
+     */
+    public String path;
```
g-coppock, line 47: If we keep the single constructor, both path and dataSinkPtr can be final.

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientFlushUpdateException.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public GenericTraitUpdatableDataSink getDataSink(WdmClient wdmClient)
+    {
+        return wdmClient.getDataSink(dataSinkPtr);
+    }
+
+    /**
+     * A flushed path
+     */
+    public String path;
```
g-coppock, line 47: path and dataSinkPtr can be final here

src/device-manager/java/Makefile.am
yunhanw-google: Done, thanks.

src/device-manager/java/src/nl/Weave/DataManagement/GenericTraitUpdatableDataSinkImpl.java
yunhanw-google: Done, thanks!

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientFlushUpdateDeviceException.java
```diff
@@ -0,0 +1,53 @@
...
+ *    limitations under the License.
+ */
+
+package nl.Weave.DataManagement;
+import nl.Weave.DeviceManager.WeaveDeviceException;
+
+/** Represents a exception that occurred on a Weave device during the WdmClient FlushUpdate operation.
+ */
+public class WdmClientFlushUpdateDeviceException extends WeaveDeviceException {
+    public WdmClientFlushUpdateDeviceException () {}
```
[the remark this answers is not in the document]
yunhanw-google: Just removed, thanks!

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientFlushUpdateDeviceException.java
yunhanw-google: Done, thanks!

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientFlushUpdateException.java
yunhanw-google: Done, thanks

src/device-manager/java/src/nl/Weave/DataManagement/WdmClientFlushUpdateDeviceException.java
yunhanw-google: Done, thanks

yunhanw-google:
@jaylogue @robszewczyk this pull is ready for review, thanks

src/device-manager/cocoa/NLWdmClientFlushUpdateError.h
```diff
@@ -0,0 +1,41 @@
...
+ *
+ */
+
+#import <Foundation/Foundation.h>
+#import "NLWeaveDeviceManagerTypes.h"
+#import "NLWeaveErrorCodes.h"
+#import "NLWeaveError.h"
+#import "NLGenericTraitUpdatableDataSink.h"
+
+@interface NLWdmClientFlushUpdateError : NLWeaveError
```
c-treyz, line 31: It's not super clear to me what the difference is between `NLWdmClientFlushUpdateError` and `NLWdmClientFlushUpdateDeviceStatusError`, might be a Weave convention I'm not familiar with, but if not, could be useful to add a documentation comment clarifying, maybe?

src/device-manager/cocoa/NLGenericTraitUpdatableDataSink.h
```diff
@@ -249,4 +249,6 @@ typedef void (^GenericTraitUpdatableDataSinkFailureBlock)(id owner, NSError * er
 /** Returns the version of the trait represented by this data sink. */
 - (WEAVE_ERROR)getVersion:(uint64_t *)val;
 
+/** Delete the trait property data. */
```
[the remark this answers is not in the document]
yunhanw-google: Done, thanks

src/device-manager/cocoa/NLGenericTraitUpdatableDataSink.mm
```diff
@@ -128,18 +128,18 @@ - (instancetype)init:(NSString *)name
...
     NLGenericTraitUpdatableDataSink * sink = (__bridge NLGenericTraitUpdatableDataSink *) reqState;
     [sink DispatchAsyncCompletionBlock:nil];
 }
 
-static void handleGenericUpdatableDataSinkError(
+static void onGenericUpdatableDataSinkError(
     void * dataSink, void * appReqState, WEAVE_ERROR code, nl::Weave::DeviceManager::DeviceStatus * devStatus)
 {
-    WDM_LOG_DEBUG(@"handleGenericUpdatableDataSinkError");
+    WDM_LOG_DEBUG(@"onGenericUpdatableDataSinkError");
```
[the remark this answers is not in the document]
yunhanw-google: agree, done, thanks

src/device-manager/WeaveDataManagementClient.cpp
```diff
@@ -984,31 +1011,45 @@ void WdmClient::ClientEventCallback(void * const aAppState, SubscriptionClient::
     case SubscriptionClient::kEvent_OnSubscriptionTerminated:
         WeaveLogDetail(DataManagement, "Client->kEvent_OnSubscriptionTerminated. Reason: %u, peer = 0x%" PRIX64 "\n",
                        aInParam.mSubscriptionTerminated.mReason, aInParam.mSubscriptionTerminated.mClient->GetPeerNodeId());
-        pWdmClient->mpSubscriptionClient->AbortSubscription();
-        err = WEAVE_ERROR_INCORRECT_STATE;
+        deviceStatus                  = new DeviceStatus();
+        deviceStatus->StatusProfileId = aInParam.mSubscriptionTerminated.mStatusProfileId;
+        deviceStatus->StatusCode      = aInParam.mSubscriptionTerminated.mStatusCode;
+        deviceStatus->SystemErrorCode = aInParam.mSubscriptionTerminated.mReason;
+        err                           = WEAVE_ERROR_WDM_REFRESH_DATA_FAILED;
```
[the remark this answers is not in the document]
yunhanw-google: yes, just add the check there. yes, it is only called when op is kOpState_RefreshData

src/device-manager/cocoa/NLWdmClient.mm
```diff
@@ -225,6 +227,48 @@ static void handleWdmClientError(
...
+    NLWdmClient * client = (__bridge NLWdmClient *) appReqState;
+
+    NSMutableArray * statusResultsList = [NSMutableArray arrayWithCapacity:pathCount];
+    NLGenericTraitUpdatableDataSink * dataSink = nil;
+    WDM_LOG_DEBUG(@"Failed path Count = %u\n", pathCount);
+
+    for (uint32_t i = 0; i < pathCount; i++) {
+        dataSink = [client getDataSink:(long long) statusResults[i].mpDataSink];
+        if (dataSink == nil) {
+            WDM_LOG_DEBUG(@"trait %d does not exist", i);
```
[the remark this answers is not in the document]
yunhanw-google: Yes, you understanding is correct. 

just update the PR, I would return nil to upper application layer since this is callback function and i cannot directly return the error and don't wanna overide the error code from lower layer, then application can check this dataSink and raise the exception finally. here I adjust the log message in unexpected, trait %d does not exist in traitMap

src/device-manager/cocoa/NLWdmClient.mm
```diff
@@ -382,6 +439,17 @@ - (void)removeDataSinkRef:(long long)traitInstancePtr;
     }
 }
 
+- (NLGenericTraitUpdatableDataSink *)getDataSink:(long long)traitInstancePtr;
+{
+    NLGenericTraitUpdatableDataSink * dataSink = nil;
+    NSString * address = [NSString stringWithFormat:@"%lld", traitInstancePtr];
+
+    if ([_mTraitMap objectForKey:address] != nil) {
```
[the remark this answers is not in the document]
yunhanw-google: yep, agree, updated, thanks

src/device-manager/cocoa/NLWdmClientFlushUpdateError.h
yunhanw-google: NLWdmClientFlushUpdateError is mobile client side error(local error), NLWdmClientFlushUpdateDeviceStatusError is remote device error

yunhanw-google: Yep, update the description in h file

src/device-manager/cocoa/NLWdmClient.mm
```diff
@@ -225,6 +227,48 @@ static void handleWdmClientError(
...
+                                report:[NSString stringWithUTF8String:nl::ErrorStr(statusResults[i].mErrorCode)]
+                                  path:[NSString stringWithUTF8String:statusResults[i].mpPath]
+                              dataSink:dataSink];
+                [statusResultsList addObject:weaveError];
+            }
+        }
+    }
+
+    NSLog(@"statusResultsList is: %@", statusResultsList);
+    [client dispatchAsyncCompletionBlockWithError:statusResultsList];
```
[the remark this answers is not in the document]
yunhanw-google: Thanks for the suggestions.

statusResult would be filled with error results when there are failed path, which would not have any success results.  If statusResult is empty array, it means no path error for flush update operation.

OnError is used to notify the protocol layer error.
OnComplete is used to notify the results when the operation is completed and there is no protocol layer error.
For flush update operations, for example, one particular path cannot be modified because of authentication issue or others, and other path has been handled correctly, the flush operation is still overally seen as successful operation, which has been designed/implemented from c++ wdm update protocol for the long time, therefore we should use onFlushUpdateComplete with particular path. This has been applied from c++ protocol layer, JAVA, python, and objective C. In this situation, you get onFlushUpdateComplete, and you read different errors, and can delete the local cached data for that particular path, you can pop or suppress the notification to UI, the overall flush operation is still seen as valid, the flow can go forward.

From objective c application layer, you can do something as below.

    [wdmClient flushUpdate:^(id owner, id result1)
     {
         flushDone = YES;
         NSLog(@"error%@",result1);
 
         for (id object in result1)
         {
             NLWdmClientFlushUpdateDeviceStatusError *wdmClientDeviceStatus;
             NLWdmClientFlushUpdateError *wdmClientError;
             if ([object isKindOfClass:[NLWdmClientFlushUpdateDeviceStatusError class]]) {
                 // You have a mutable array!
                 wdmClientDeviceStatus = object;
                 NSLog(@"path%@",wdmClientDeviceStatus.path);
                 //[wdmClientDeviceStatus.dataSink deleteData:wdmClientDeviceStatus.path];
                 
             } else {
                 wdmClientError = object;
                 NSLog(@"path%@",wdmClientError.path);
                 [wdmClientError.dataSink deleteData:wdmClientError.path];
                 // You don't have a mutable array!
             }
 

         }
     }
                   failure:^(id owner, NSError *error)
     {
         flushDone = YES;
         gotError = YES;
         NSLog(@"Flush Update Error:%@", error);
         NSDictionary* params = [NSDictionary dictionaryWithObjectsAndKeys:@"Flush Update Error", @"Title", error, @"Error", nil];
         [self performSelectorOnMainThread:@selector(createErrorSheet:) withObject:params waitUntilDone:YES];
     }];

Thanks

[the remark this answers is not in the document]
c-treyz: Okay, I see what you mean. From the client perspective, it's not clear to me that the distinction between a protocol layer error and an application/operation layer error is so obvious – either way, the client is going to need to determine how to handle the failure in either of those cases, which would often have a decent amount of overlap. But it sounds like it's just an OpenWeave standard that I'm not familiar with and is outside the scope of this CL.

In that case, just a couple small nits:
1. I think it would be quite helpful to have a documentation comment for |flushUpdate| in the header explaining the expected values/types that will be passed to the `WdmClientCompletionBlock`(and the meaning, ie. that if the success block is called with an empty array, that indicates that all paths were flushed successfully)
2. Renaming `-[NLWdmClient dispatchAsyncCompletionBlockWithError:]` so it's a little more clear why it's not calling the failure block – maybe `-[NLWdmClient dispatchAsyncCompletionBlockWithResults:]`? WDYT?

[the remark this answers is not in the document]
yunhanw-google: Thanks, just address the comments:
a. add detailed descriptions on FlushUpdate API in header.
b. renaming toward dispatchAsyncCompletionBlockWithResults.

Thanks!

src/device-manager/python/openweave/WdmClient.py
```diff
@@ -26,11 +26,42 @@
 from .WeaveStack import *
 from .GenericTraitUpdatableDataSink import *
 from .ResourceIdentifier import *
+from .WeaveUtility import WeaveUtility
 
-__all__ = [ 'WdmClient' ]
+__all__ = [ 'WdmClient', 'WdmFlushUpdateStatusStruct', 'WdmClientFlushUpdateError', 'WdmClientFlushUpdateDeviceError' ]
+
+WEAE_ERROR_STATUS_REPORT = 4044
```
robszewczyk, line 33: Misspelled: `WEAVE`

src/lib/core/WeaveTLVDebug.cpp
```diff
@@ -140,14 +140,20 @@ static void DumpHandler(DumpWriter aWriter, const char *aIndent, const TLVReader
             break;
 
         case kTLVType_UTF8String:
-            err = temp.GetDataPtr(strbuf);
-            VerifyOrExit(err == WEAVE_NO_ERROR, aWriter("Error in kTLVType_UTF8String"));
+            if (len > 0)
+            {
+                err = temp.GetDataPtr(strbuf);
+                VerifyOrExit(err == WEAVE_NO_ERROR, aWriter("Error in kTLVType_UTF8String"));
+            }
```
robszewczyk, line 147: set `strbuf = ""` in the else clause?

src/device-manager/python/openweave/WdmClient.py
yunhanw-google: Done, thanks

PR #542, opened 2020-04-06
Bug/broken serialization

robszewczyk:
Fix bugs in generated serialization code

* incorrect offset of nullified fields
* incomplete deallocation of generated structures.

### DIFF
--- a/src/lib/support/SerializationUtils.cpp
+++ b/src/lib/support/SerializationUtils.cpp
@@ -998,12 +998,23 @@ exit:
 static WEAVE_ERROR FindNullifiedFieldsArray(void *aStructureData, const SchemaFieldDescriptor *aSchemaDescriptor, uint8_t *&aNullifiedFields)
 {
     uint32_t offset;
-    WEAVE_ERROR err;
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
     const FieldDescriptor *lastFieldDescriptor = &(aSchemaDescriptor->mFields[aSchemaDescriptor->mNumFieldDescriptorElements - 1]);
+    bool lastFieldIsArray = false;
 
-    err = GetArrayElementSize(offset, lastFieldDescriptor, lastFieldDescriptor->GetType());
-    SuccessOrExit(err);
-
+    if (aSchemaDescriptor->mNumFieldDescriptorElements >= 2)
+    {
+        lastFieldIsArray = (aSchemaDescriptor->mFields[aSchemaDescriptor->mNumFieldDescriptorElements - 2].GetType() == SerializedFieldTypeArray);
+    }
+    if (lastFieldIsArray)
+    {
+        offset = sizeof(void *);
+    }
+    else
+    {
+        err = GetArrayElementSize(offset, lastFieldDescriptor, lastFieldDescriptor->GetType());
+        SuccessOrExit(err);
+    }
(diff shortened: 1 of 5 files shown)


robszewczyk:
Reviewers: Please note there are two somewhat unrelated changes here: one fixes issues arising in serialization of structures where the last field is an array, the second is strictly related to memory leaks in deallocation of deserialized structures.  Both issues ended up being shoved into the same PR: in devising the tests, I ran the `TestEventLogging` test suite with address sanitizer, and that uncovered leaks in both the core deallocation code for deserialized structures and in the test code.

src/lib/support/SerializationUtils.cpp
```diff
@@ -998,12 +998,23 @@ WEAVE_ERROR ReadDataForType(TLVReader &aReader, void *aStructureData, const Fiel
...
-    err = GetArrayElementSize(offset, lastFieldDescriptor, lastFieldDescriptor->GetType());
-    SuccessOrExit(err);
-
+    if (aSchemaDescriptor->mNumFieldDescriptorElements >= 2)
+    {
+        lastFieldIsArray = (aSchemaDescriptor->mFields[aSchemaDescriptor->mNumFieldDescriptorElements - 2].GetType() == SerializedFieldTypeArray);
+    }
+    if (lastFieldIsArray)
+    {
+        offset = sizeof(void *);
```
robszewczyk, line 1011: Per chat with Lukasz: absent some archaic decorations on pointers (`far`, `near`, `huge`), all data pointers in C are the same size.  C99 makes allowances for the function pointers being of different size than `void *` but that distinction has no impact here.

src/lib/support/SerializationUtils.cpp
```diff
@@ -998,12 +998,23 @@ WEAVE_ERROR ReadDataForType(TLVReader &aReader, void *aStructureData, const Fiel
 static WEAVE_ERROR FindNullifiedFieldsArray(void *aStructureData, const SchemaFieldDescriptor *aSchemaDescriptor, uint8_t *&aNullifiedFields)
 {
     uint32_t offset;
-    WEAVE_ERROR err;
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
     const FieldDescriptor *lastFieldDescriptor = &(aSchemaDescriptor->mFields[aSchemaDescriptor->mNumFieldDescriptorElements - 1]);
+    bool lastFieldIsArray = false;
```
jaylogue, line 1003: Can you explain why the last field needs to be special cased?

robszewczyk: My apologies, this should have been in the commit message.

The serialization code assumes the bitmap of nullified fields, when *present*, follows the last element in the structure.   Arrays in this serialized representation are stored as pointers in the parent structure, i.e. `repeated TYPE fieldName = n;` is represented in the generated code as a `size_t fieldName_num_elements; TYPE * fieldName_buffer;`.  To calculate where the structure ends (excluding the nullified fields bitmap), the code uses the following calculations:
* last field is a non-array: we take the offset of the last field and add the size the field type would occupy in an array.  This accounts for packing/alignment issues around the representation for `TYPE`
* last field is an array: we identify that the field is an array by looking at the second last field in the schema descriptor.  The last field in the schema descriptor is the type of the elements in the array, but the array itself is represented in the container structure as a pointer.  Given that data pointers are of the same size as `void *`, we use the `sizeof(void *)` as the size of the array in the container structure.

PR #544, opened 2020-04-07
On sockets platforms, starting a timer should wake up layer's event loop

robszewczyk:
Calling `nl::Weave::System::Timer::Start` does not wake the Weave
thread. This is arguably an understood limitation stemming from Weave
code being geared towards a single-threaded execution.  However, on a
number of platforms -- such as Java -- we admit to having pieces of
the Weave stack run within different threads but enclosed within the
correct lock around the Weave stack.  This results in issues where the
Java caller schedules a timer: the main Weave thread may be sleeping
in the event loop and does not adjust the wake time in response to the
new timer being scheduled.

The construction proposed herein parallels the semantics
`nl::Weave::System::Timer::Start` offers on LwIP systems and parallels
the semantics offered by `ScheduleWork`.

On systems where POSIX locking is already enabled, calling
`WakeSelect` from the layer's event loop does not actually involve a
write to a pipe, the wake is only invoked when the `Timer::Start` is
invoked from a different thread.

### DIFF
--- a/src/system/SystemTimer.cpp
+++ b/src/system/SystemTimer.cpp
@@ -125,9 +125,7 @@ bool Timer::IsEarlierEpoch(const Timer::Epoch &inFirst, const Timer::Epoch &inSe
  */
 Error Timer::Start(uint32_t aDelayMilliseconds, OnCompleteFunct aOnComplete, void* aAppState)
 {
-#if WEAVE_SYSTEM_CONFIG_USE_LWIP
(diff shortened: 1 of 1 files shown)


gerickson:
> Port to CHIP? @turon?

See https://github.com/project-chip/connectedhomeip/issues/274.

turon:
@gerickson, sure, I can post it -- I can't assign myself issues in CHIP however.

PR #548, opened 2020-04-09
Improve the search for OpenSSL on MacOS

robszewczyk:
MacOS migrated to LibreSSL sometime ago, and the preferred
way to use OpenSSL on MacOS is via `brew`.  This commit alters
the search algorithm for the preferred openssl executable to
first check the directory where brew actually installed the
executable; this addresses dynamic link issues with some partial
openssl installations.

### DIFF
--- a/src/test-apps/TestWeaveTool.py
+++ b/src/test-apps/TestWeaveTool.py
@@ -23,6 +23,7 @@
 #
 
 import os
+import platform
 import sys
 import unittest
 import tempfile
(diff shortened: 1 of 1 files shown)


robszewczyk:
The motivation for this pull request is: after running a `brew reinstall openssl`, the `openssl` in the `PATH` (usually located in `/usr/local/bin/openssl` no longer works, instead it crashes with:

    dyld: Library not loaded: /usr/local/lib/libcrypto.1.1.dylib
      Referenced from: /usr/local/bin/openssl
      Reason: image not found
    Abort trap: 6

The patch pulls in the `openssl` executable from the `brew` path, in exactly the same way as the algorithm in the `Makefile-Standalone`

PR #549, opened 2020-04-09
Use libc++ as default Android STL when make Makefile-Android

yunhanw-google:
-- Internal android weave daily build fails in
src/lib/../../src/device-manager/WeaveDataManagementClient.cpp:39:18:
fatal error: string: No such file or directory
  #include <string>
-- At default, it uses system, but it don't have c++ string,
/usr/local/android/android-ndk-r14/system/include.
-- if using libc++, it has
/usr/local/android/android-ndk-r14/sources/cxx-stl/llvm-libc++/include/string.
-- Therefore, use Android NDK C++ runtime to build against libc++
-- checking latest android ndk, the same applies.

### DIFF
--- a/Makefile-Android
+++ b/Makefile-Android
@@ -36,16 +36,12 @@ ANDROID_HOME                    ?= $(ANDROID_ROOT)/android-sdk-linux-r23.0.2
 ANDROID_NDK_HOME                ?= $(ANDROID_ROOT)/android-ndk-r10
 ANDROID_API                     ?= $(ANDROID_RELEASE)
 ANDROID_ABI                     ?= armeabi-v7a arm64-v8a x86 x86_64
-ANDROID_STL                     ?= $(ANDROID_CXX)
+ANDROID_STL                     ?= libc++
 DEBUG                           ?= 0
 ENABLE_TARGETED_LISTEN          ?= 0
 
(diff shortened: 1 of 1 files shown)


Makefile-Android
```diff
@@ -44,7 +44,7 @@ ENABLE_TARGETED_LISTEN          ?= 0
 ANDROID_RELEASE                 ?= 22
 
 # Deprecated -- Use ANDROID_STL
```
gerickson, line 46: Does this comment need to change to reflect the commit / description for this PR?

robszewczyk: I've looked at the Makefile, and the fix is correct, and comment may remain unchanged.  The "deprecated" annotation refers to the use of the variable: the variable was deprecated (should not be used to modify behavior by external programs) but is still used as a base value for `ANDROID_STL`.  Given the passage of time, I would be satisfied if the `ANDROID_CXX` was removed altogether, and the default value of `ANDROID_STL` were changed to `libc++`

robszewczyk: FWIW, I'm testing the internal build with the above suggestion implemented; if nothing relies on `ANDROID_CXX` we should remove the deprecated variable

yunhanw-google: Yep, I have updated PR and removed ANDROID_CXX.

PR #555, opened 2020-04-15
Fixed Function Implementation that Converts DER to Weave Signature.

emargolis:
In the variant of the function where output is EncodedECDSASignature
structure, the R and S values of the signature should be copied
and not just referenced.

### DIFF
--- a/src/lib/profiles/security/WeaveSig.cpp
+++ b/src/lib/profiles/security/WeaveSig.cpp
@@ -1,5 +1,6 @@
 /*
  *
+ *    Copyright (c) 2020 Google LLC.
  *    Copyright (c) 2013-2017 Nest Labs, Inc.
  *    All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


src/lib/profiles/security/WeaveSig.cpp
```diff
@@ -514,11 +514,20 @@ WEAVE_ERROR ConvertECDSASignature_DERToWeave(const uint8_t * sigBuf, uint8_t sig
     ASN1_PARSE_ENTER_SEQUENCE {
         // r INTEGER
         ASN1_PARSE_ELEMENT(kASN1TagClass_Universal, kASN1UniversalTag_Integer);
-        sig.R = const_cast<uint8_t *>(reader.Value);
+
+        VerifyOrExit(sig.R != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
+        VerifyOrExit(sig.RLen >= reader.ValueLen, err = WEAVE_ERROR_BUFFER_TOO_SMALL);
+
+        memcpy(sig.R, const_cast<uint8_t *>(reader.Value), reader.ValueLen);
```
jaylogue, line 521: Is this the change we want to make?  The original purpose of this function was to parse a DER form of signature and place pointers to the R and S values into a EncodedECDSASignature structure.  If desired, it was the job of the caller to copy these values (typically by calling EncodeWeaveECDSASignature) before the input buffer was destroyed.

Since this function is only used by code outside of openweave-core, I can't see how it is being used.  Is there a reason why the caller can't do the copying?  Or, is there a reason why the caller isn't calling the function below to directly encode the signature in Weave format?

emargolis: This function is used outside of openweave-core by the platform specific implementation of GenerateECDSASignatureFunct function, which is called in GenerateOperationalDeviceCert().

As you suggested this can be implemented by GenerateECDSASignatureFunct. This function can use its local EncodedECDSASignature structure and then copy values to the output EncodedECDSASignature structure.

I can do that if this is the way we want to proceed.

Side note: maybe worth adding Copy() method to the EncodedECDSASignature() class.

jaylogue: I see.  I think the name of this function is misleading, which leads to the confusion.  The purpose of this function is really to _decode_ a DER signature, not convert it.  Thus, it is very much like DecodeWeaveECDSASignature() below, with a different input format.

DecodeWeaveECDSASignature() does not copy the signature values.  However I do see the need to do that in the case of GenerateOperationalDeviceCert.

I suggest keeping the function implementation as-is, but renaming it to DecodeCopyECDSASignature_DER().  This makes it clear that function both decodes and copies the data.

Maybe in the future we'll rename DecodeWeaveECDSASignature() to follow the same pattern; e.g. DecodeECDSASignature_Weave().  No need to do that now, however.

emargolis: updated

PR #556, opened 2020-04-17
Improved logging in python

jaylogue:
- **Reworked the way developer logging happens when using the openweave wrappers for python.** With this change, all WeaveLog...() calls in C++ code result in a callback to a logging function in python.  The default implementation of this function directs log messages to a standard python logger object associated with the python WeaveStack object.  By default, the python logger is configured to output messages to stdout, in a format identical to that used by the C++ code.  However, developers are free to reconfigure the logger object, or replace it altogether, giving them complete control over the handling of C++ generated logs in python applications.

- **Added support for dynamic logging behavior.**  Modified the standard implementation of nl::Weave::Logging::Log() so that the default logging behavior can be overridden at runtime. Applications can now use the SetLogFunct() function to specify a pointer to a function to be called for every logged message.  Setting the function pointer to NULL (the default value) results in the original behavior of logging to stdout via printf().

### DIFF
--- a/src/device-manager/python/WeaveDeviceManager-ScriptBinding.cpp
+++ b/src/device-manager/python/WeaveDeviceManager-ScriptBinding.cpp
@@ -29,6 +29,7 @@
 #include <errno.h>
 #include <unistd.h>
 #include <fcntl.h>
+#include <sys/time.h>
 
 #include <SystemLayer/SystemLayer.h>
 #include <SystemLayer/SystemError.h>
@@ -69,6 +70,7 @@ using DeviceDescription::IdentifyDeviceCriteria;
 extern "C" {
 typedef void * (*GetBleEventCBFunct)(void);
 typedef void (*ConstructBytesArrayFunct)(const uint8_t *dataBuf, uint32_t dataLen);
+typedef void (*LogMessageFunct)(uint64_t time, uint64_t timeUS, const char *moduleName, uint8_t category, const char *msg);
 }
 
 enum BleEventType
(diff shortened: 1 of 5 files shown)


jaylogue:
Latest version has some minor changes to make it easier for applications to setup customize logging.

yunhanw-google:
Overall it looks perfect!!

Currently during CI validation, we use make -f Makefile-Standalone TIMESTAMP=1, then weave c++ log with timestamp would automatically go to stdout
the test case is openweave-core/src/test-apps/happy/tests/standalone/pairing/test_weave_pairing_01.py, the log is in /tmp/xxxxxWEAVE-PAIRING-MOBILEnode02.out
With your change, the weave c++ log message has piped to python, but at default, timestamp is disabled. Could you enable timestamp at default,
def __init__(self, datefmt=None, logModule=True, logLevel=False, **logTimestamp=True**, logMSecs=True)
https://github.com/openweave/openweave-core/pull/556/commits/540be9be4452ba108f22860656d761c71b4c473c#diff-870dcb3a4097be4f5a8c1d196c0db469R119

Or add 
logFormatter = WeaveLogFormatter(logLevel=True, logTimestamp=True) in 
https://github.com/openweave/openweave-core/blob/master/src/test-apps/happy/lib/WeaveDeviceManager.py

Thanks

jaylogue:
> Could you enable timestamp at default

@yunhanw-google I cannot do that, as that would enable timestamps everywhere, including in the device manager shell.

I have pushed another commit that enables timestamps for the pairing tests.

yunhanw-google:
> > Could you enable timestamp at default
> 
> @yunhanw-google I cannot do that, as that would enable timestamps everywhere, including in the device manager shell.
> 
> I have pushed another commit that enables timestamps for the pairing tests.

Yep, fully agree on this. Just review and validate the code, all looks good, thank you!

PR #561, opened 2020-04-24
Fix CommandSender::SendCommand with binding override

holbrookt:
Fixes a bug in CommandSender where providing SendCommand with a binding
argument doesn't override the default binding.

### DIFF
--- a/src/lib/profiles/data-management/Current/CommandSender.cpp
+++ b/src/lib/profiles/data-management/Current/CommandSender.cpp
@@ -267,7 +267,7 @@ WEAVE_ERROR CommandSender::SendCommand(PacketBuffer *aRequestBuf, Binding *aBind
         kMaxCommandRequestHeaderSize = sizeof(headerEncoding)
     };
 
-    VerifyOrExit(binding != NULL, err = WEAVE_ERROR_INCORRECT_STATE);
+    VerifyOrExit(binding != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
     VerifyOrExit(aSendParams.ProfileId != 0, err = WEAVE_ERROR_INVALID_ARGUMENT);
 
     // If the user didn't pass in a request buffer that we can re-use to blit the header, allocate our own.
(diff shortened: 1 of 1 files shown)


src/lib/profiles/data-management/Current/CommandSender.cpp
```diff
@@ -267,7 +267,7 @@ WEAVE_ERROR CommandSender::SendCommand(PacketBuffer *aRequestBuf, Binding *aBind
         kMaxCommandRequestHeaderSize = sizeof(headerEncoding)
     };
 
-    VerifyOrExit(mBinding != NULL || aBinding != NULL, err = WEAVE_ERROR_INVALID_ARGUMENT);
+    VerifyOrExit(binding != NULL, err = WEAVE_ERROR_INCORRECT_STATE);
```
holbrookt, line 270: Is this the right error for this case?

pidarped: The error state seems ok to me, although an INCORRECT_STATE would have been more appropriate if the passed Binding was also assigned to the member.
The Binding member variable is set at Initialization. But this API allows an override to that. However, I don't see the member also getting set to the passed Binding variable. Seems like the Binding that is passed to this API is a temporary Binding to use, and a later fallback to the member Binding is allowed? Or should overriding by a passed Binding automatically also set the member too? I would have preferred an API to set the member Binding too so that the application can change it later. Jay, thoughts?

jaylogue: If you called Init() with aBinding=NULL, then you are indirectly stating that you will never call SendCommand() with aBinding=NULL. Hence, I think INVALID_ARGUMENT is correct. There's nothing wrong with the state of the CommandSender, you just passed an invalid argument.

holbrookt: That makes sense to me, Jay. I'll change it.

Pradip - the intention with this API is that the user can override the default Binding just for this one call to SendCommand. If the user doesn't pass a Binding to SendCommand, it is assumed that the default Binding will be used. It would be incorrect to overwrite mBinding.

pidarped: Yes, I understand that the Binding argument being passed to the SendCommand call allows the application to use it temporarily for that call. I was wondering if it might be worth having an additional API to set the member Binding so that the application can change it after initialization. Currently, if you have to use a different Binding than the initialized member, you would have to pass it every time to the SendCommand call.

Issue #564, opened 2020-04-28
nrf5-app.mk does not consider JLINK_SN on targets flash_softdevice and erase


pierredelisle:
"-f nrf52" should be replaced with $(NRFJPROG_FLAGS).

PR #565, opened 2020-04-28
Fix for issue #564: nrf5-app.mk does not consider JLINK_SN on targets…

pierredelisle:
… flash_softdevice and erase

### DIFF
--- a/build/nrf5/nrf5-app.mk
+++ b/build/nrf5/nrf5-app.mk
@@ -274,14 +274,14 @@ endef
 flash-softdevice flash_softdevice :
 	@if test -z "$(SOFTDEVICE_IMAGE)"; then echo "SoftDevice image not found in $(SOFTDEVICE_IMAGE_DIR)"; false; fi
 	@echo "FLASH $(SOFTDEVICE_IMAGE)"
-	$(NO_ECHO)$(NRFJPROG) -f nrf52 --program $(SOFTDEVICE_IMAGE) --sectorerase
+	$(NO_ECHO)$(NRFJPROG) $(NRFJPROG_FLAGS) --program $(SOFTDEVICE_IMAGE) --sectorerase
 	@echo "RESET DEVICE"
-	$(NO_ECHO)$(NRFJPROG) -f nrf52 --reset
+	$(NO_ECHO)$(NRFJPROG) $(NRFJPROG_FLAGS) --reset
 
 # Erase device
 erase :
 	@echo "ERASE DEVICE"
-	$(NO_ECHO)$(NRFJPROG) -f nrf52 --eraseall
+	$(NO_ECHO)$(NRFJPROG) $(NRFJPROG_FLAGS) --eraseall
 
 # Clean build output
 clean ::


build/nrf5/nrf5-app.mk
```diff
@@ -274,14 +274,14 @@ endef
 flash-softdevice flash_softdevice :
 	@if test -z "$(SOFTDEVICE_IMAGE)"; then echo "SoftDevice image not found in $(SOFTDEVICE_IMAGE_DIR)"; false; fi
 	@echo "FLASH $(SOFTDEVICE_IMAGE)"
-	$(NO_ECHO)$(NRFJPROG) -f nrf52 --program $(SOFTDEVICE_IMAGE) --sectorerase
+	$(NO_ECHO)$(NRFJPROG) $(NRFJPROG_FLAGS) --program $(SOFTDEVICE_IMAGE) --sectorerase
 	@echo "RESET DEVICE"
-	$(NO_ECHO)$(NRFJPROG) -f nrf52 --reset
+	$(NO_ECHO)$(NRFJPROG) $(NRFJPROG_FLAGS) --reset
```
suryanshup, line 279: In the flash-app target, I see $$ being used for $$(NRFJPROG_FLAGS).  Can we keep it consistent here ?

pierredelisle: As explained by Jay:
------
As to why $$ in some places, the answer is subtle, but very important.  If you look closely at the lines you cited, you will see they are encompassed in a define / endef statement:

# ==================================================
# Late-bound rules for building the application
# ==================================================

define AppBuildRules

[...]

# Rule to flash a pre-built application
flash-app flash_app :
@echo "FLASH $$(APP_HEX)"
$(NO_ECHO)$$(NRFJPROG) $$(NRFJPROG_FLAGS) --program $$(APP_HEX) --sectorerase
@echo "RESET DEVICE"
$(NO_ECHO)$$(NRFJPROG) $$(NRFJPROG_FLAGS) --reset

[...]

endif

This means that the lines you cited are not actually declaring make rules.  Instead, they are simply part of a long, multi-line variable definition that is setting the value of the AppBuildRules variable.  Later, in the execution of the main Makefile, the contents of this variable get evaluated as a make script [using $(eval)].  This causes the rules expressed in the variable text to be declared as make rules.

Per the normal rules for defining make variables, $'s in the value of the variable definition undergo substitution at the time the variable is used.  In the case of AppBuildRules, that is the time at which its contents are evaluated.  However, in some cases we want variables in the AppBuildRules text to undergo substitution at the time the declared rules are actually run.  For these cases we use $$.  This has the following effect: when AppBuildRules is evaluated, make replaces $$(var) with $(var) (as shown for PRICE above). Later, when the rules in AppBuildRules actually run, the resultant $(var) is further replaced with whatever the contents of var is at that time.

So $(var) inside a define / endif block means expand at the time the rules are evaluated; $$(var) means expand at the time the rules are run.

As to why $$(NRFJPROG) in some places and $(NRFJPROG) in others: if you look closely, the latter are not within a define / endif block.

PR #566, opened 2020-04-28
DeviceLayer: implement CASE auth delegate

kghost:
Currently, only legacy CASE auth delegate is implemented. The patch
implements the new CASE auth delegate interface for device layer.

Change-Id: Ifb6a4dcdcd29a75eca47bd84158345bb2253ebff

### DIFF
--- a/src/adaptations/device-layer/CASEAuth.cpp
+++ b/src/adaptations/device-layer/CASEAuth.cpp
@@ -28,6 +28,7 @@
 #include <Weave/Core/WeaveTLV.h>
 #include <Weave/Profiles/WeaveProfiles.h>
 #include <Weave/Profiles/security/WeaveSecurity.h>
+#include <Weave/Profiles/security/WeaveSig.h>
 #include <Weave/Profiles/security/WeaveCert.h>
 #include <Weave/Profiles/security/WeaveCASE.h>
 #include <Weave/Profiles/service-provisioning/ServiceProvisioning.h>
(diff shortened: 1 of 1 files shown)


jaylogue:
I think this needs to be coordinated with @emargolis .  He undoubtedly has a commit that makes the same change.

src/adaptations/device-layer/CASEAuth.cpp
```diff
@@ -67,14 +68,45 @@ class CASEAuthDelegate : public WeaveCASEAuthDelegate
...
 
-    virtual WEAVE_ERROR GetNodeCertInfo(bool isInitiator, uint8_t *buf, uint16_t bufSize, uint16_t& certInfoLen);
-    virtual WEAVE_ERROR GetNodePrivateKey(bool isInitiator, const uint8_t *& weavePrivKey, uint16_t & weavePrivKeyLen);
-    virtual WEAVE_ERROR ReleaseNodePrivateKey(const uint8_t *weavePrivKey);
-    virtual WEAVE_ERROR GetNodePayload(bool isInitiator, uint8_t *buf, uint16_t bufSize, uint16_t& payloadLen);
-    virtual WEAVE_ERROR BeginCertValidation(bool isInitiator, WeaveCertificateSet& certSet, ValidationContext& validContext);
-    virtual WEAVE_ERROR HandleCertValidationResult(bool isInitiator, WEAVE_ERROR& validRes, WeaveCertificateData *peerCert,
-            uint64_t peerNodeId, WeaveCertificateSet& certSet, ValidationContext& validContext);
-    virtual WEAVE_ERROR EndCertValidation(WeaveCertificateSet& certSet, ValidationContext& validContext);
+#if !WEAVE_CONFIG_LEGACY_CASE_AUTH_DELEGATE
```
jaylogue, line 71: If we do go with this commit, there's no need to support both legacy and new interfaces.  Just force a compile error if WEAVE_CONFIG_LEGACY_CASE_AUTH_DELEGATE is set.

PR #574, opened 2020-04-30
Make package name configurable to be able to distinguish between inte…

githanwang1:
…rnal and external builds

Make the default package name configurable but make the default 'openweave' to distinguish between internal and public package. Please see internal ticket for more context (can cc as requested).

### DIFF
--- a/src/device-manager/python/Makefile.am
+++ b/src/device-manager/python/Makefile.am
@@ -130,20 +130,20 @@ EXTRA_DIST +=										\
     $(NULL)
 
 .openweave-wheel: $(srcdir)/build-openweave-wheel.py weave-device-mgr _WeaveDeviceMgr.la $(shell find $(srcdir) -type f -name '*.py')
-	rm -f openweave-*.whl
+	rm -f *.whl
 	$(PYTHON) $(srcdir)/build-openweave-wheel.py
 	touch .openweave-wheel
 
 install-exec-local: .openweave-wheel
 	mkdir -p $(DESTDIR)$(pyexecdir)
-	rm -f $(DESTDIR)$(pyexecdir)/openweave-*.whl
-	cp openweave-*.whl $(DESTDIR)$(pyexecdir)
+	rm -f $(DESTDIR)$(pyexecdir)/*.whl
+	cp *.whl $(DESTDIR)$(pyexecdir)
 
 uninstall-local:
-	rm -f $(DESTDIR)$(pyexecdir)/openweave-*.whl
+	rm -f $(DESTDIR)$(pyexecdir)/*.whl
 
 clean-local:
-	rm -f .openweave-wheel openweave-*.whl
+	rm -f .openweave-wheel *.whl
 
 endif # WEAVE_BUILD_INSTALLABLE_PYTHON_PACKAGE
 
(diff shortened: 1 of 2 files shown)


src/device-manager/python/Makefile.am
```diff
@@ -130,20 +130,20 @@ EXTRA_DIST +=										\
...
-	rm -f openweave-*.whl
+	rm -f *.whl
 	$(PYTHON) $(srcdir)/build-openweave-wheel.py
 	touch .openweave-wheel
 
 install-exec-local: .openweave-wheel
 	mkdir -p $(DESTDIR)$(pyexecdir)
-	rm -f $(DESTDIR)$(pyexecdir)/openweave-*.whl
-	cp openweave-*.whl $(DESTDIR)$(pyexecdir)
+	rm -f $(DESTDIR)$(pyexecdir)/*.whl
```
jaylogue, line 139: This will delete all .whl files in the destination directory.  That could wipe out other unrelated .whl files.

jaylogue:
This looks like a feature that belongs in the Google internal fork of OpenWeave.

src/device-manager/python/Makefile.am
githanwang1: True, just assumed that there is only the openweave whl in that directory but we can probably restrict it to the specific package name.

githanwang1:
> This looks like a feature that belongs in the Google internal fork of OpenWeave.

Wasn't sure about where to make the change for a shared file but can definitely try make the change in internal weave. In that case, probably easier to just hardcode the package name, but it'll be cleaner in a way.

jaylogue:
@githanwang1 : I'm not against making the build-openweave-wheel.py script customizable.  What I don't want to do is wire that logic into the OpenWeave build system.

I think there's a middle ground here: We can modify build-openweave-wheel.py to accept a package name as a command line option (--pkgname foo).   Then, after the OpenWeave build completes, your internal build script can run build-openweave-wheel.py again, supplying the package name it desires.  

We can also add other options to further customize the behavior of the script if desired.  E.g. we could have an option that specifies where to put the output file.

This approach could also be a means to build both python2 and python3 wheel files in the same build.

githanwang1:
Thanks for suggesting alternatives but not sure if I'm following correctly. Doesn't the makefile have to adjusted regardless based on the configured package name? Is the concern about modifying the makefiles (as part of the build system) vs modifying the build-openweave-wheel.py file?

The internal fork suggestion you mentioned earlier definitely seems tempting, since it's a simpler solution but maybe not as valuable.

jaylogue:
I have been assuming that your internal build process involved some form of outer build script that called the OpenWeave Makefile (e.g. Makefile-Standalone).  Please correct me if I'm wrong.

Based on this, I was suggesting that we modify build-openweave-wheel.py so that it can be used as a stand-alone tool outside of the OpenWeave build process, in addition to being invoked during the build.

Then we modify your outer build script to invoke build-openweave-wheel.py a second time, after the OpenWeave build completes, passing it the necessary arguments to produce the package you desire.

githanwang1:
Ah, it seems like I might need further clarification. What is meant by the OpenWeave build process? 
Yes, we call `make -f Makefile-Standalone` which ends up invoking the OpenWeave build process (but I'm not sure where the separation is). I can see your point now that if we can get build-openweave-wheel.py to work standalone, there is no need to modify the makefiles as well, since we aren't calling make.

jaylogue:
Well, to be a little more clear, you will still need to invoke make -f Makefile-Standalone, because you need the C++ code needs to be built into a shared library.  But after that completes, you can invoke build-openweave-wheel.py to package up the shared library, along with the python code, into a .whl file configured as you like.

PR #577, opened 2020-05-07
Access Control relaxation for fetching device credentials over a PASE session

pidarped:
These relaxed access controls would allow a mobile client device that
is connected over PASE during first-device pairing to access device
credentials only over that same session, and cache them for
assisting other devices during subsequent pairing attempts.

### DIFF
--- a/src/lib/profiles/network-provisioning/NetworkProvisioning.cpp
+++ b/src/lib/profiles/network-provisioning/NetworkProvisioning.cpp
@@ -66,6 +66,8 @@ NetworkProvisioningServer::NetworkProvisioningServer()
  */
 WEAVE_ERROR NetworkProvisioningServer::Init(WeaveExchangeManager *exchangeMgr)
 {
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+
     ExchangeMgr = exchangeMgr;
     FabricState = exchangeMgr->FabricState;
     mCurOp = NULL;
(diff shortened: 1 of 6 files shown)


src/lib/core/WeaveServerBase.cpp
```diff
@@ -191,6 +191,55 @@ WEAVE_ERROR WeaveServerBase::SendStatusReport(ExchangeContext *ec, uint32_t stat
...
 
+/**
+ * Indicates if the session with the given node id and the session key id is
+ * authorized to retrieve secret credential information.
+ *
+ * @return Returns 'true' if the given peer is privileged, else
+ * 'false'.
+ */
+
+bool WeaveServerBase::IsSessionPrivileged(uint16_t keyId, uint64_t peerNodeId) const
```
jaylogue, line 202: I think this notion of a "privileged" session is too generic, and should not be incorporated into WeaveServerBase.  Since the current goal is to enable some very specific privileges in the context of network provisioning and fabric provisioning, I think the logic behind tracking and acting on these privileges should be in the respective server classes, and named specifically to their purpose.

For example, for network provisioning, I suggest we use names such as HasCredentialAccessPrivilege(), GrantCredentialAccessProvilege(), ClearCredentialAccessPrivilege() and mCredentialAccessSession.  A similar pattern would apply in the FabricProvisioningServer; e.g. HasFabricConfigAccessPrivilege(), etc.

pidarped: I had a bit of doubt when putting it in WeaveServerBase for the same reason but kept them there to, essentially, re-use the code for the derived provisioning servers. Will move them to the respective server classes and update.

src/lib/profiles/network-provisioning/NetworkProvisioning.cpp
```diff
@@ -369,6 +427,14 @@ void NetworkProvisioningServer::HandleRequest(ExchangeContext *ec, const IPPacke
         VerifyOrExit(dataLen >= 1, err = WEAVE_ERROR_INVALID_MESSAGE_LENGTH);
         err = delegate->HandleAddNetwork(payload);
         payload = NULL;
+
+        if (err == WEAVE_NO_ERROR)
+        {
+            // Authorize the current session for privileged access to secret
+            // credential information.
+            server->GrantCredentialAccessPrivilege(msgInfo->KeyId, msgInfo->SourceNodeId);
```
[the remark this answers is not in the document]
pidarped: Had previously looked at a delegate implementation where the SendAddNetworkComplete() was sent synchronously within the HandleAddNetwork() function, but Jay clarified offline that it need not be the case always. Moved the GrantAccess to inside SendAddNetworkComplete.

PR #579, opened 2020-05-09
Relax Check for TLV Path Element in the VerifyElement().

emargolis:
Elements in the list (path) can also be of an AnonymousTag.


src/lib/core/WeaveTLVReader.cpp
```diff
@@ -1382,7 +1382,6 @@ WEAVE_ERROR TLVReader::VerifyElement()
                 return WEAVE_ERROR_INVALID_TLV_TAG;
             break;
         case kTLVType_Structure:
-        case kTLVType_Path:
```
jaylogue, line 1385: We need as case statement for kTLVType_Path that just breaks.  Otherwise the code will return WEAVE_ERROR_INCORRECT_STATE.

jaylogue:
Interesting: TLVWriter allows writing a path with an anonymous tag, but TLVReader doesn't allow reading one.  Not surprisingly, the TestTLV unit test checks the ability to write such a path, but doesn't check reading it.

src/lib/core/WeaveTLVReader.cpp
emargolis: fixed

emargolis: fixed

emargolis:
@jaylogue an update in the TLVWriter was also needed. Please take a look

@gerickson updated unit test to verify new behavior

PR #581, opened 2020-05-09
Updated Weave TLV Format specification to revision 5

jaylogue:
- Renamed path type to list and clarified text in corresponding section.
- Added section on canonical ordering of tags.
- Added section describing anonymous tags.
- Clarified descriptions of structure and array types.
- Clarified encoding of floating point values with reference to RFC-4506.
- Moved revision history to end of document.


gerickson:
Any opposition to converting this to AsciiDoc based on the work I did [regarding the same for CHIP](https://github.com/project-chip/connectedhomeip/blob/master/docs/specs/chip-tlv-format.adoc)?

PR #583, opened 2020-05-11
Add virtual d'tors, specify cast

EliRibble:
This resolves some compilation errors when cross-compiling with clang in our project. No behavior is changed, just clarified to avoid compiler warnings that we have set as errors.


EliRibble:
Thanks for the quick response. I'll split out the cast changes so that the virtual d'tor work can be removed since it isn't wanted.

PR #584, opened 2020-05-12
[WIP] Addition of a new helper class to deal with WDL maps.

mrjerryjohns:
**_This is a sketch of a possible helper class to deal with WDL maps written with a specific Nest product team in mind. This likely won't be hosted in openweave-core, but given the ease with which unit-test code can be created here, I decided to put it here to provide examples of it's use_**

This templated class provides a more user-friendly type when interacting with WDL maps. WDL maps are constrained in their keys to having 16-bit numerical keys. As such, these maps cannot be thought of as traditional maps that have the key being the logical/primary key used to index the data in the collection. Rather, the key is merely a unique number assigned to each item, lacking any semantic or logical meaning. The actual logical key is embedded in the item itself, and be any WDL type. In this sense, WDL maps can be thought of as spare arrays.

Applications will typically want to interact with the collection locally on devices more naturally as a keyed collection, with the data index using the logical key. This requires them to maintain a two-map solution: one to store the original data in the trait, and another to translate from the logical key to the map key.

This templated class utilizes boost's multi_index_container, which permits indexing the data against multiple keys while still retaining efficient storage and look-up of the data. The class wraps the provided boost type and provides helper methods to interact with each key's sub-table. 

Specializing the template requires passing in both the logical key type as well as the value type as well.

Requirements on the types include:
     Logical Key Type: Has both '<' and '==' operators implemented.
     Value Type: Has the '==' operator implemented.

The class also provides methods to compare against another instance of the collection to easily figure out the set of items added, removed or modified against the logical key. This allows for comparisons

### DIFF
--- a/src/lib/profiles/data-management/Current/WdmDictionary.h
+++ b/src/lib/profiles/data-management/Current/WdmDictionary.h
@@ -1,0 +1,230 @@
+/*
+ *
+ *    Copyright (c) 2018 Google LLC.
+ *    Copyright (c) 2013-2017 Nest Labs, Inc.
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ *    @file
+ *      This file provides a more logical container representation of a WDM dictionary
+ *      that permits both indexing against the dictionary key as well as a more logical primary
+ *      key
+ */
+
+#ifndef _WEAVE_DATA_MANAGEMENT_WDM_DICTIONARY_H
+#define _WEAVE_DATA_MANAGEMENT_WDM_DICTIONARY_H
+
+#include <Weave/Profiles/data-management/Current/WdmManagedNamespace.h>
+
+#include <Weave/Core/WeaveCore.h>
+#include <Weave/Core/WeaveMessageLayer.h>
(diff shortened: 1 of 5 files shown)


src/lib/profiles/data-management/Current/WdmDictionary.h
```diff
@@ -0,0 +1,230 @@
...
+namespace Profiles {
+namespace WeaveMakeManagedNamespaceIdentifier(DataManagement, kWeaveManagedNamespaceDesignation_Current) {
+
+//
+// This templated class provides a more user-friendly type when interacting with WDL maps.
+// WDL maps are constrained in their keys to having 16-bit numerical keys. As such, these maps
+// cannot be thought of as traditional maps that have the key being the logical/primary key used to 
+// index the data in the collection. Rather, the key is merely a unique number assigned to each item,
+// lacking any semantic or logical meaning. The actual logical key is embedded in the item itself, and
+// be any WDL type. In this sense, WDL maps can be thought of as spare arrays.
```
gerickson, line 57: spare -> sparse?

src/lib/profiles/data-management/Current/WdmDictionary.h
```diff
@@ -0,0 +1,230 @@
...
+class WdmDictionary {
+public:
+    //
+    // This is the Item that is used in the collection, housing both the 16-bit dictionary
+    // key as well as the logical KeyT typed key.
+    // It also contains the value as well.
+    //
+    struct Item {
+        Item(uint16_t dictKey) { _dict_key = dictKey; _logical_key = 0; }
+        Item() { _dict_key = 0; _logical_key = 0; }
```
gerickson, line 86: Style nit: `Item()` -> `Item(void)`

src/lib/profiles/data-management/Current/WdmDictionary.h
```diff
@@ -0,0 +1,230 @@
...
+
+        bool operator<(const Item& i1) const {
```
gerickson, line 88: Style nit `operator<(` -> `operator <(`.

gerickson:
It's going to take me a bit to wade into the STL/Boost (since it's always a difficult read), but the concept seems sound. Maybe we create a new repo, openweave-utilities, that has some optional, but-in-some-contexts-helpful utilities like this that developers can avail themselves of?

PR #587, opened 2020-05-13
Replace R/O Weave TLV Format Specification PDF with a R/W AsciiDoc

gerickson:


Vyrastas:
Why are you using Asciidoc instead of PDF?  To allow third-party contributions?

gerickson:
> Why are you using Asciidoc instead of PDF? To allow third-party contributions?

So the source of truth is public rather than behind a Google Doc domain only accessible by Google employees and can, as a result, be editable by anyone.

This also makes it easier to sync changes to CHIP (since this is leveraged wholesale as-is there).

Vyrastas:
> So the source of truth is public rather than behind a Google Doc domain only accessible by Google employees and can, as a result, be editable by anyone.
> 
> This also makes it easier to sync changes to CHIP (since this is leveraged wholesale as-is there).

Do you still want to post this to openweave.io, then?  They don't have native support for Asciidoc so I'd have to convert it to HTML first on my end, but we also don't need to be maintaining multiple copies of things, we can just link to the docs here if that makes more sense.

gerickson:
> > So the source of truth is public rather than behind a Google Doc domain only accessible by Google employees and can, as a result, be editable by anyone.
> > This also makes it easier to sync changes to CHIP (since this is leveraged wholesale as-is there).
> 
> Do you still want to post this to openweave.io, then? They don't have native support for Asciidoc so I'd have to convert it to HTML first on my end, but we also don't need to be maintaining multiple copies of things, we can just link to the docs here if that makes more sense.

Assuming @jaylogue is good with this proposed change, I defer to you @Vyrastas on what you think the right approach is for supporting openweave.io.

jaylogue:
I see a few issues with the formatting:

- The tables use alternating shading for rows, without a distinct color / shading for the header row.  This is confusing for small tables.

- In the PDF, the control byte tables used blank fields for the "n/a" bits of the control byte (those not germane to the purpose of the table).  This styling proved to be too subtle (people were confused), and I'm afraid of a similar problem here as well.  I suggest adding another column header row above the others indicating that the 8 bit columns are a depiction of the control byte.  Then placing something like an 'x' in each unused column.  E.g.:

```
 |     Control Byte       |                           |
 |        |  Elem Type    |                           |
 | 7  6  5  4  3  2  1  0 |       Description         |
 ------------------------------------------------------
 | x  x  x  0  0  0  0  0 |       Signed Integer      |
                         ...
```

gerickson:
> I see a few issues with the formatting:
> 
> * The tables use alternating shading for rows, without a distinct color / shading for the header row.  This is confusing for small tables.
> * In the PDF, the control byte tables used blank fields for the "n/a" bits of the control byte (those not germane to the purpose of the table).  This styling proved to be too subtle (people were confused), and I'm afraid of a similar problem here as well.  I suggest adding another column header row above the others indicating that the 8 bit columns are a depiction of the control byte.  Then placing something like an 'x' in each unused column.  E.g.:
> 
> ```
>  |     Control Byte       |                           |
>  |        |  Elem Type    |                           |
>  | 7  6  5  4  3  2  1  0 |       Description         |
>  ------------------------------------------------------
>  | x  x  x  0  0  0  0  0 |       Signed Integer      |
>                          ...
> ```

Regarding the shading, I've tried to find ways to control it; however, AsciiDoc doesn't appear to offer an easy way to do so. It renders differently depending the renderer.

I can make the change to further clarify the control byte.

jaylogue:
> Regarding the shading, I've tried to find ways to control it; however, AsciiDoc doesn't appear to offer an easy way to do so. It renders differently depending the renderer.

I think this gets to the issue of how we publish this.  Its unclear to me how much control over rending AsciiDoc you have on github.  But I believe when it come to publishing a spec, we cannot be beholden to a simplistic rendering of the document as appears to be the case here.  We need to find a way to have more control over the presentation format that users see.

One way would be to maintain the source document in AsciiDoc on github, but publish a PDF rendering with the desired format for users/readers.  This would be what we link to from openweave.io.  This has the added benefit producing a document that is readily downloadable and usable off-line.

gerickson:
> > I see a few issues with the formatting:
> > 
> > * The tables use alternating shading for rows, without a distinct color / shading for the header row.  This is confusing for small tables.
> > * In the PDF, the control byte tables used blank fields for the "n/a" bits of the control byte (those not germane to the purpose of the table).  This styling proved to be too subtle (people were confused), and I'm afraid of a similar problem here as well.  I suggest adding another column header row above the others indicating that the 8 bit columns are a depiction of the control byte.  Then placing something like an 'x' in each unused column.  E.g.:
> > 
> > ```
> >  |     Control Byte       |                           |
> >  |        |  Elem Type    |                           |
> >  | 7  6  5  4  3  2  1  0 |       Description         |
> >  ------------------------------------------------------
> >  | x  x  x  0  0  0  0  0 |       Signed Integer      |
> >                          ...
> > ```
> 
> Regarding the shading, I've tried to find ways to control it; however, AsciiDoc doesn't appear to offer an easy way to do so. It renders differently depending the renderer.
> 
> I can make the change to further clarify the control byte.

@jaylogue, the element type and tag control tables have been updated.

gerickson:
> > Regarding the shading, I've tried to find ways to control it; however, AsciiDoc doesn't appear to offer an easy way to do so. It renders differently depending the renderer.
> 
> I think this gets to the issue of how we publish this. Its unclear to me how much control over rending AsciiDoc you have on github. But I believe when it come to publishing a spec, we cannot be beholden to a simplistic rendering of the document as appears to be the case here. We need to find a way to have more control over the presentation format that users see.
> 
> One way would be to maintain the source document in AsciiDoc on github, but publish a PDF rendering with the desired format for users/readers. This would be what we link to from openweave.io. This has the added benefit producing a document that is readily downloadable and usable off-line.

At this point, I think the rendering differences between the PDF, what GitHub renders, and what my AsciiDoctor.js Preview plug-in on Chrome show are pretty minor. I think the benefit of open outweighs the slight formatting control. Either way, this is an order of magnitude better than IETF RFCs in ASCII.

jaylogue:
To be clear, there was nothing "closed" about my proposal.  The master document would still be checked in and editable.  Only there would be a rendering step (probably automatable) that would produce the final presentation.

I think its pretty clear that we need better rendering control, at least in some contexts.  E.g. it would not be acceptable, IMO, to dumb-down the rendering of the TVL Schema spec to AsciiDoc+github format.  However it might be possible to express that document in AsciiDoc, presuming we have control over the presentation.

Vyrastas:
Given the discussion, whatever format you want to present to a reader is what I'd put on openweave.io.  If you're happy with the rendered AsciiDoc format for people who just want to read the doc, I'd convert what you host here to HTML for displaying there, rather than linking to the source.  Otherwise we can continue with PDF if someone is still going to produce that.

gerickson:
@Vyrastas and @jaylogue, how would you like to move forward with this pull request?

Vyrastas:
If this enables you to sync content with CHIP much easier, then I would submit it in this format and we can sort out the presentation on openweave.io separately.  We can also keep the PDF version here too, rather than replacing it.

gerickson:
> If this enables you to sync content with CHIP much easier, then I would submit it in this format and we can sort out the presentation on openweave.io separately. We can also keep the PDF version here too, rather than replacing it.

Added the PDF version back.

jaylogue:
Looks like the latest push mixed in some other changes.

gerickson:
> I'm fine with merging this change. I have been investigating the use of asciidoctor to render PDFs from asciidoc files and it looks quite promising. I'm thinking at some point we can add makefiles in the various doc directories to generate the .pdf files from the .adoc files using asciidoctor-pdf and an appropriate theme file. This should give us reasonable control over the formatting.

Perfect, thanks.

PR #590, opened 2020-05-13
Specifically cast certDER to void*.

EliRibble:
clang complained about the implicit conversion - this more clearly
shows what is being done to the pointer.

Bug: 155214583
Change-Id: I90f01a46c4f0eb08388a0c6dcf9c4dbdd5a73d75


EliRibble:
Looks like this passed review - what does it take to get this merged?

PR #592, opened 2020-05-15
Persist Event

didishe90:
-Serializes and persit events stored in WeaveCircularTLVBuffer and associated states
-Use ShutdownInProgress flag to suspend logging manager during shutdown process

### DIFF
--- a/src/lib/core/WeaveCircularTLVBuffer.h
+++ b/src/lib/core/WeaveCircularTLVBuffer.h
@@ -70,6 +70,8 @@ public:
     inline size_t AvailableDataLength(void) const { return mQueueSize - mQueueLength; };
     inline size_t GetQueueSize(void) const { return mQueueSize; };
     inline uint8_t *GetQueue(void) const { return mQueue; };
+    inline void SetQueueHead(uint8_t *aQueueHead) { mQueueHead = aQueueHead; };
+    inline void SetQueueLength(size_t aQueueLength) { mQueueLength = aQueueLength; };
 
     WEAVE_ERROR EvictHead(void);
 
--- a/src/lib/core/WeaveTLV.h
+++ b/src/lib/core/WeaveTLV.h
@@ -222,6 +222,7 @@ public:
     // *** See WeaveTLVWriter.cpp file for API documentation ***
 
     void Init(uint8_t *buf, uint32_t maxLen);
+    void InitMalloced(uint8_t *& outBuf, uint32_t initialBufSize = 256, uint32_t maxLen = 0xFFFFFFFFUL);
     void Init(PacketBuffer *buf, uint32_t maxLen = 0xFFFFFFFFUL);
     void Init(PacketBuffer *buf, uint32_t maxLen, bool allowDiscontiguousBuffers);
 
@@ -247,6 +248,8 @@ public:
     WEAVE_ERROR Put(uint64_t tag, double v);
     WEAVE_ERROR PutBoolean(uint64_t tag, bool v);
     WEAVE_ERROR PutBytes(uint64_t tag, const uint8_t *buf, uint32_t len);
+    WEAVE_ERROR StartPutBytes(uint64_t tag, uint32_t totalLen);
+    WEAVE_ERROR ContinuePutBytes(const uint8_t *buf, uint32_t len);
     WEAVE_ERROR PutString(uint64_t tag, const char *buf);
     WEAVE_ERROR PutString(uint64_t tag, const char *buf, uint32_t len);
     WEAVE_ERROR PutStringF(uint64_t tag, const char *fmt, ...);
@@ -282,6 +285,9 @@ public:
     // PacketBuffers.
     static WEAVE_ERROR GetNewPacketBuffer(TLVWriter& writer, uintptr_t& bufHandle, uint8_t *& bufStart, uint32_t& bufLen);
     static WEAVE_ERROR FinalizePacketBuffer(TLVWriter& writer, uintptr_t bufHandle, uint8_t *bufStart, uint32_t dataLen);
+
+    // Implementations of the GetNewBufferFunct that support writing into a dynamic buffer
+    static WEAVE_ERROR GetNewBuffer_Malloced(TLVWriter& writer, uintptr_t& bufHandle, uint8_t *& bufStart, uint32_t& bufLen);
 
 #if WEAVE_CONFIG_PROVIDE_OBSOLESCENT_INTERFACES
     static WEAVE_ERROR GetNewInetBuffer(TLVWriter& writer, uintptr_t& bufHandle, uint8_t *& bufStart, uint32_t& bufLen);
--- a/src/lib/core/WeaveTLVWriter.cpp
+++ b/src/lib/core/WeaveTLVWriter.cpp
@@ -32,6 +32,7 @@
 #include <Weave/Core/WeaveEncoding.h>
 #include <Weave/Core/WeaveTLV.h>
 #include <Weave/Support/CodeUtils.h>
+#include <stdlib.h>
 
 namespace nl {
 namespace Weave {
@@ -175,6 +176,30 @@ void TLVWriter::Init(uint8_t *buf, uint32_t maxLen)
     ImplicitProfileId = kProfileIdNotSpecified;
     GetNewBuffer = NULL;
     FinalizeBuffer = NULL;
+}
+
+/**
+ * Initializes a TLVWriter object to write into a dynamic buffer.
+ *
+ * @param[in]   buf     A reference to a pointer that will receive the allocated buffer.
+ * @param[in]   maxLen  The maximum number of bytes that should be written to the output buffer.
+ * @param[in]   initialBufSize
+ *                      The initial number of bytes that should be allocated to the buffer.
+ */
+void TLVWriter::InitMalloced(uint8_t *& outBuf, uint32_t initialBufSize, uint32_t maxLen)
+{
+    mBufHandle = (uintptr_t)&outBuf;
+    mBufStart = mWritePoint = outBuf = (uint8_t *)malloc(initialBufSize);
+    mMaxLen = maxLen;
+    mRemainingLen = initialBufSize;
+    mLenWritten = 0;
+    mContainerType = kTLVType_NotSpecified;
+    SetContainerOpen(false);
+    SetCloseContainerReserved(true);
+
+    ImplicitProfileId = kProfileIdNotSpecified;
+    FinalizeBuffer = NULL;
+    GetNewBuffer = GetNewBuffer_Malloced;
 }
 
 /**
@@ -660,6 +685,77 @@ WEAVE_ERROR TLVWriter::Put(uint64_t tag, double v)
 WEAVE_ERROR TLVWriter::PutBytes(uint64_t tag, const uint8_t *buf, uint32_t len)
 {
     return WriteElementWithData(kTLVType_ByteString, tag, (const uint8_t *) buf, len);
+}
+
+/**
+ * Encodes a TLV byte string in multiple chunks. This should be used with ContinuePutBytes.
+ *
+ * @param[in]   tag             The TLV tag to be encoded with the value, or @p AnonymousTag if the
+ *                              value should be encoded without a tag.  Tag values should be
+ *                              constructed with one of the tag definition functions ProfileTag(),
+ *                              ContextTag() or CommonTag().
+ * @param[in]   totalLen        The total number of bytes to be encoded.
+ *
+ * @retval #WEAVE_NO_ERROR      If the method succeeded.
+ * @retval #WEAVE_ERROR_TLV_CONTAINER_OPEN
+ *                              If a container writer has been opened on the current writer and not
+ *                              yet closed.
+ * @retval #WEAVE_ERROR_INVALID_TLV_TAG
+ *                              If the specified tag value is invalid or inappropriate in the context
+ *                              in which the value is being written.
+ * @retval #WEAVE_ERROR_BUFFER_TOO_SMALL
+ *                              If writing the value would exceed the limit on the maximum number of
+ *                              bytes specified when the writer was initialized.
+ * @retval #WEAVE_ERROR_NO_MEMORY
+ *                              If an attempt to allocate an output buffer failed due to lack of
+ *                              memory.
+ * @retval other                Other Weave or platform-specific errors returned by the configured
+ *                              GetNewBuffer() or FinalizeBuffer() functions.
+ *
+ */
+WEAVE_ERROR TLVWriter::StartPutBytes(uint64_t tag, uint32_t totalLen)
+{
+    TLVFieldSize lenFieldSize;
+    TLVType type = kTLVType_ByteString;
+
+    if (totalLen <= UINT8_MAX)
+        lenFieldSize = kTLVFieldSize_1Byte;
+    else if (totalLen <= UINT16_MAX)
+        lenFieldSize = kTLVFieldSize_2Byte;
+    else
+        lenFieldSize = kTLVFieldSize_4Byte;
+
+    WEAVE_ERROR err = WriteElementHead((TLVElementType) (type | lenFieldSize), tag, totalLen);
+
+    return err;
+}
+
+/**
+ * Encodes a TLV byte string value. This should be used with StartPutBytes.
+ *
+ * @param[in]   buf             A pointer to a buffer containing the bytes string to be encoded.
+ * @param[in]   len             The number of bytes to be encoded.
+ *
+ * @retval #WEAVE_NO_ERROR      If the method succeeded.
+ * @retval #WEAVE_ERROR_TLV_CONTAINER_OPEN
+ *                              If a container writer has been opened on the current writer and not
+ *                              yet closed.
+ * @retval #WEAVE_ERROR_INVALID_TLV_TAG
+ *                              If the specified tag value is invalid or inappropriate in the context
+ *                              in which the value is being written.
+ * @retval #WEAVE_ERROR_BUFFER_TOO_SMALL
+ *                              If writing the value would exceed the limit on the maximum number of
+ *                              bytes specified when the writer was initialized.
+ * @retval #WEAVE_ERROR_NO_MEMORY
+ *                              If an attempt to allocate an output buffer failed due to lack of
+ *                              memory.
+ * @retval other                Other Weave or platform-specific errors returned by the configured
+ *                              GetNewBuffer() or FinalizeBuffer() functions.
+ *
+ */
+WEAVE_ERROR TLVWriter::ContinuePutBytes(const uint8_t *buf, uint32_t len)
+{
+    return WriteData(buf, len);
 }
 
 /**
@@ -1869,6 +1965,32 @@ WEAVE_ERROR TLVWriter::GetNewPacketBuffer(TLVWriter& writer, uintptr_t& bufHandl
         bufStart = NULL;
         bufLen = 0;
     }
+
+    return WEAVE_NO_ERROR;
+}
+
+/**
+ * An implementation of a TLVWriter GetNewBuffer function for writing to a dynamic buffer.
+ *
+ * The GetNewBuffer_Malloced() function supplies new output space to a TLVWriter by doubling
+ * the size of the underlying dynamic buffer as needed to store the encoding.  The function
+ * is designed to be assigned to the TLVWriter GetNewBuffer function pointer.
+ *
+ * See the GetNewBufferFunct type definition for additional information on the API of the
+ * GetNewBuffer_Malloced() function.
+ */
+WEAVE_ERROR TLVWriter::GetNewBuffer_Malloced(TLVWriter& writer, uintptr_t& bufHandle, uint8_t *& bufStart, uint32_t& remainingLen)
+{
+    size_t prevSize = (writer.mWritePoint - *(uint8_t **)bufHandle) + writer.mRemainingLen;
+
+    size_t newSize = prevSize * 2;
+    if (newSize > writer.mMaxLen)
+    {
+        newSize = writer.mMaxLen;
+    }
+    *(uint8_t **)bufHandle = (uint8_t *)realloc(*(uint8_t **)bufHandle, newSize);
+    bufStart = *(uint8_t **)bufHandle + prevSize;
+    remainingLen = newSize - prevSize;
 
     return WEAVE_NO_ERROR;
 }
--- a/src/lib/profiles/data-management/Current/EventLoggingTags.h
+++ b/src/lib/profiles/data-management/Current/EventLoggingTags.h
@@ -195,6 +195,30 @@ enum
     kTag_Args  = 2  ///< An array of arguments to be sent along with the token message.
 };
 
+/**
+ * @brief
+ *  Tags for persisting events
+ */
+enum
+{
+    kTagNum_SerializedEventState                    = 1,
+    kTagNum_SerializedEventSet                      = 2,
+    kTag_PersistEvent_ImportanceLevel               = 3,
+    kTag_PersistEvent_EventData                     = 4,
+    kTag_PersistEvent_FirstEventId                  = 5,
+    kTag_PersistEvent_LastEventId                   = 6,
+    kTag_PersistEvent_FirstEventTimestamp           = 7,
+    kTag_PersistEvent_LastEventTimestamp            = 8,
+    kTag_PersistEvent_EventIdCounter                = 9
+
+#if WEAVE_CONFIG_EVENT_LOGGING_UTC_TIMESTAMPS
+    ,
+    kTag_PersistEvent_FirstEventUTCTimestamp        = 10,
+    kTag_PersistEvent_LastEventUTCTimestamp         = 11,
+    kTag_PersistEvent_UTCInitialized                = 12
+#endif
+};
+
 } // namespace WeaveMakeManagedNamespaceIdentifier(DataManagement, kWeaveManagedNamespaceDesignation_Current)
 } // namespace Profiles
 } // namespace Weave
--- a/src/lib/profiles/data-management/Current/LoggingManagement.cpp
+++ b/src/lib/profiles/data-management/Current/LoggingManagement.cpp
@@ -27,6 +27,7 @@
 
 #include <Weave/Profiles/data-management/Current/WdmManagedNamespace.h>
 #include <Weave/Profiles/data-management/DataManagement.h>
+#include <Weave/Profiles/WeaveProfiles.h>
 
 #include <Weave/Profiles/bulk-data-transfer/Development/BulkDataTransfer.h>
 #include <Weave/Profiles/bulk-data-transfer/Development/BDXMessages.h>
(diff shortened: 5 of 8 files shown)


src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +437,210 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+ *
+ * Serializes the events in WeaveCircularTLVBuffer and the associated states into the
+ * supplied buffer and set mShutdownInProgress flag to true such that no further
+ * events can be added to the buffer.
+ *
+ * This method is intended to be used by devices that do not retain RAM while sleeping,
+ * allowing them to persist events before going to sleep and thereby prevent lost of
+ * events
+ */
+WEAVE_ERROR LoggingManagement::PersistEvent(uint8_t * buf, uint32_t bufSize, uint32_t & serializedLen)
```
jaylogue, line 451: This method should be named SerializeEvents() because its purpose is to serialize the currently stored events.  Whether the output will be persisted is immaterial.

Also, the method should take a reference to a TLVWriter.  This way the caller gets to decide where the data is encoded to.

Note that it is not very obvious how large the output should be.  Thus the caller will probably want to supply a TVLWriter that dynamically grows the underlying buffer.

mrjerryjohns, line 451: An event-by-event blit seem highly in-inefficient and un-necessary - why aren't we just global copy/write of the entire buffer contents (ptr + size)?

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +437,210 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+exit:
+    Platform::CriticalSectionExit();
+    return err;
+}
+
+/**
+ * @brief Load previously persisted Weave event.
+ *
+ */
+WEAVE_ERROR LoggingManagement::LoadEvent(uint8_t * serializedEvent, uint32_t serializedEventLen)
```
mrjerryjohns, line 527: Similar comment to above.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +437,210 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+    err = writer.StartContainer(ProfileTag(kWeaveProfile_Common, kTag_SerializedEvent), kTLVType_Structure, rootContainer);
+    SuccessOrExit(err);
+
+    for (int importanceType = kImportanceType_First; importanceType < kImportanceType_Last + 1; importanceType++)
+    {
+        // Get the specified event buffer.
+        CircularEventBuffer * circular_event_buffer = GetImportanceBuffer((ImportanceType)importanceType);
+
+        TLVType container;
+        err = writer.StartContainer(ContextTag(importanceType), kTLVType_Structure, container);
```
jaylogue, line 472: This should not use the important type as a tag.   In fact, the outer container should not be structure.  Instead, it should be an array of inner structures ("serialized event sets") with the importance expressed as a field within the inner structure.  E.g.:

```
serialized-event-state => ARRAY OF serialized-event-set

serialized-event-set => STRUCTURE
{
    importance-level [0]               : UNSIGNED INTEGER,
    event-data [1]                     : OCTET STRING,
    head-offset [2]                    : UNSIGNED INTEGER,
    length [3]                         : UNSIGNED INTEGER,
    first-event-id [4]                 : UNSIGNED INTEGER,
    last-event-id [5]                  : UNSIGNED INTEGER,
    first-event-timestamp [6]          : UNSIGNED INTEGER,
    last-event-timestamp [7]           : UNSIGNED INTEGER,
    first-event-utc-timestamp [8,opt]  : UNSIGNED INTEGER,
    first-event-utc-timestamp [9,opt]  : UNSIGNED INTEGER,
    last-event-utc-timestamp [10,opt]  : UNSIGNED INTEGER,
}
```

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +437,210 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+    TLVReader reader;
+    reader.Init(serializedEvent, serializedEventLen);
+
+    TLVType rootContainer;
+    err = reader.Next(kTLVType_Structure, ProfileTag(kWeaveProfile_Common, kTag_SerializedEvent));
+    SuccessOrExit(err);
+    err = reader.EnterContainer(rootContainer);
+    SuccessOrExit(err);
+
+    for (int importanceType = kImportanceType_First; importanceType < kImportanceType_Last + 1; importanceType++)
```
jaylogue, line 540: This will crash if kImportanceType_Last is greater than mMaxImportanceBuffer.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +437,210 @@ void LoggingManagement::DestroyLoggingManagement(void)
     Platform::CriticalSectionExit();
 }
 
+/**
+ * Persist the Weave events of all importance types.
+ *
+ * Serializes the events in WeaveCircularTLVBuffer and the associated states into the
+ * supplied buffer and set mShutdownInProgress flag to true such that no further
```
jaylogue, line 444: This method should not change the state of the LoggingManagement object.  Its purpose should be solely to persist event data.

src/lib/profiles/data-management/Current/LoggingManagement.h
```diff
@@ -292,6 +302,7 @@ class LoggingManagement
     uint32_t mThrottled;
     ImportanceType mMaxImportanceBuffer;
     bool mUploadRequested;
+    bool mShutdownInProgress;
```
jaylogue, line 305: I don't see any purpose for this field.  Why do we need a "shutdown in progress" state?  And why isn't this another value in the LoggingManagementStates enum?

src/lib/profiles/data-management/Current/EventLoggingTags.h
```diff
@@ -195,6 +195,30 @@ enum
...
 };
 
+/**
+ * @brief
+ *  Tags for persisting events
+ */
+enum
+{
+    kTag_SerializedEvent                         = 1,
+    kTag_PersistEvent_EventBuffer,
```
jaylogue, line 205: Always give specific values to tag enums.  Otherwise you run the risk of someone inserting a value, causing all the other values to change, and breaking compatibility with previously encoded data.

src/lib/core/WeaveCircularTLVBuffer.h
```diff
@@ -70,6 +70,8 @@ class NL_DLL_EXPORT WeaveCircularTLVBuffer
     inline size_t AvailableDataLength(void) const { return mQueueSize - mQueueLength; };
     inline size_t GetQueueSize(void) const { return mQueueSize; };
     inline uint8_t *GetQueue(void) const { return mQueue; };
+    inline void SetQueueHead(uint8_t *aQueueHead) { mQueueHead = aQueueHead; };
```
jaylogue, line 73: These are not intended to be publicly settable.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +437,210 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+    for (int importanceType = kImportanceType_First; importanceType < kImportanceType_Last + 1; importanceType++)
+    {
+        // Get the specified event buffer.
+        CircularEventBuffer * circular_event_buffer = GetImportanceBuffer((ImportanceType)importanceType);
+
+        TLVType container;
+        err = writer.StartContainer(ContextTag(importanceType), kTLVType_Structure, container);
+        SuccessOrExit(err);
+
+        err = writer.PutBytes(ContextTag(kTag_PersistEvent_EventBuffer), circular_event_buffer->mBuffer.GetQueue(), circular_event_buffer->mBuffer.GetQueueSize());
```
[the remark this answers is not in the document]
didishe90: Do you mean we should serialize only the actual event data, if that is it better to split the data by the 'turning point' of the ring buffer? Or do you mean that we check if the buffer is empty before serializing it?

src/lib/core/WeaveCircularTLVBuffer.h
didishe90: I guess this is probably a dumb question - then I should add CircularEventBuffer as a friend class to get access to private members of WeaveCircularTLVBuffer?

src/lib/profiles/data-management/Current/LoggingManagement.cpp
didishe90: Okay I looked at this code again... I'm not sure if my understanding is right - Seems like the number of buffers and their importance levels are hardcoded in nlError nl::Weave::Profiles::DataManagement::Platform::Init(). It not related to kImportanceType_Last, and not event mMaxImportanceBuffer? If this is true, should I pass the values as arguments and let platform decide how many buffers need to be persisted?

src/lib/core/WeaveCircularTLVBuffer.h
jaylogue: No.  Because we can't have WeaveCircularTLVBuffer reference classes in the WDM profile.

I think the right way to do this is to make CircularEventBuffer a subclass of WeaveCircularTLVBuffer and then make these fields protected.

I'm okay if you want to leave the code as-is for now.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
didishe90: This is probably another dumb question - I guess when you init the TVLWriter with a fixed sized buffer, it's not going to grow? Could you provide me some key words for "TVLWriter that dynamically grows the underlying buffer", so I can search in codebase or google it? Thanks.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +437,210 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+    TLVType rootContainer;
+    err = writer.StartContainer(ProfileTag(kWeaveProfile_Common, kTag_SerializedEvent), kTLVType_Structure, rootContainer);
+    SuccessOrExit(err);
+
+    for (int importanceType = kImportanceType_First; importanceType < kImportanceType_Last + 1; importanceType++)
+    {
+        // Get the specified event buffer.
+        CircularEventBuffer * circular_event_buffer = GetImportanceBuffer((ImportanceType)importanceType);
+
+        TLVType container;
```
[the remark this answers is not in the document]
didishe90: Okay if I call this function SerializeEvents(), then can I call the outer function ( the one I named PersistEvent()) - SerializeAllEvents()?

src/lib/profiles/data-management/Current/LoggingManagement.cpp
jaylogue: No. The number of buffers is determined by the second argument to LoggingManagement() constructor.

Instead of using importance and calling GetImportanceBuffer(), you need to start with mEventBuffer and walk the chain of CircularEventBuffers via the mNext pointer until you reach the end (mNest == NULL).

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +437,210 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+        size_t eventBufferQueueLength;
+
+        err = reader.Next(kTLVType_Structure, ContextTag(importanceType));
+        SuccessOrExit(err);
+        err = reader.EnterContainer(container);
+        SuccessOrExit(err);
+
+        err = reader.Next(kTLVType_ByteString, ContextTag(kTag_PersistEvent_EventBuffer));
+        SuccessOrExit(err);
+        VerifyOrExit(reader.GetLength() == circular_event_buffer->mBuffer.GetQueueSize(), err = WEAVE_ERROR_INVALID_ARGUMENT);
```
jaylogue, line 555: This presents a pretty major problem with this serialization format.  If the event buffer sizes aren't an exact match between serializing and deserializing, it is effectively impossible to restore the data.  That means, for example, a firmware update that increased the buffer size would invalidate any persisted data.  That seems very undesirable.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
didishe90: I see. Thanks.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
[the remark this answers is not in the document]
jaylogue: That's fine.  However, I would expect a method called LoggingManagement::SerializeEvents() to serialize all events in the LoggingManagement object.  So SerializeAllEvents() is a little redundant, IMO.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
didishe90: I see. Then I guess I should serialize only the event data, join the two part together if the event data hits the physical end address of the ring buffer. And when loading the events back, just check if the buffer size if large enough and set mQueueHead = mQueue. Is this a better way?

jaylogue: That sounds more flexible.  One thing to consider is when the buffer is smaller than the data when deserializing.  The ideal implementation would throw away the oldest events in each buffer until the remaining events fit.  For simplicity, you could just throw an error in this case, presuming that buffer sizes only ever grow.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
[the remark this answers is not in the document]
didishe90: Oh you mean they share same name? I see. Sorry that I guess I just always forget about overloading...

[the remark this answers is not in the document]
jaylogue: Sorry, I didn't mean overloading.  I mean they are methods on different classes.  LoggingManagement::SerializeEvents() serializes events for every importance level buffer.  CircularEventBuffer::SerializeEvents() serializes events contained in that buffer.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
jaylogue: Correct.  A TLVWriter initialized with a fixed buffer will fail if you try to write more than will fit.

I'm not sure where there is a good example of using a TLVWriter to write into a dynamic buffer.  We don't generally do this in openweave-core.    I can write you one if you want.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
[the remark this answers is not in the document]
didishe90: I see. Thanks.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
didishe90: Oh I see, thanks, I can write it myself~ 
Sorry I just thought there's some trick to use TLVWriter with fixed buffer to make it grow. I forgot there's dynamic buffer... really dumb...

jaylogue: You might find it a little un-obvious how to do this, as it involves setting a custom GetNewBuffer callback on the TLVWriter object, and there aren't a lot of examples of that. 
 
There is documentation on the GetNewBuffer callback here: https://github.com/openweave/openweave-core/blob/efbed8f77e8c6a371508892abdfe3822f7aee14e/src/lib/core/WeaveTLVWriter.cpp#L69-L71

Let me know if you get stuck.

didishe90: Thank you very much.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -2146,6 +2246,161 @@ void CircularEventReader::Init(CircularEventBuffer * inBuf)
...
+    SuccessOrExit(err);
+    err = reader.Get(counterValue);
+    SuccessOrExit(err);
+
+    err = reader.Next(kTLVType_UnsignedInteger, ContextTag(kTag_PersistEvent_StartingCounterValue));
+    SuccessOrExit(err);
+    err = reader.Get(startingCounterValue);
+    SuccessOrExit(err);
+
+    static_cast<PersistedCounter *>(mEventIdCounter)->ResumeEpoch(startingCounterValue, counterValue);
```
jaylogue, line 2374: Why is this necessary?   The event id counter is already persistent.   Overwriting the counter value when LoadEvents() is called could possibly make the counter value roll backwards, which seems incorrect.

didishe90: This is to fix the mistake I made earlier. When the device wakes up, event counter starts from a new epoch, and event buffer would have events counters from two epochs, which might spread a wide range. The problem is - When going through events in the event buffer for uploading, it always starts from the lowest counter in the buffer and +1 in each iteration... Then the weave thread would be block...

src/lib/core/WeaveTLVWriter.cpp
```diff
@@ -177,6 +177,52 @@ void TLVWriter::Init(uint8_t *buf, uint32_t maxLen)
...
+ * Initializes a TLVWriter object to write into a dynamic buffer.
+ *
+ * @param[in]   buf     A pointer to the dynamic buffer into which TLV should be written.
+ * @param[in]   maxLen  The maximum number of bytes that should be written to the output buffer.
+ * @param[in]   allowResize
+ *                      If true, allocating new space as needed to store the data written.  If
+ *                      false, writing will fail with WEAVE_ERROR_BUFFER_TOO_SMALL if the written
+ *                      data exceeds the initial size of the dynamic buffer
+ */
+void TLVWriter::Init(std::vector<uint8_t> *buf, uint32_t maxLen, bool allowResize)
```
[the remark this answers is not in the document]
didishe90: Just updated, please take a look~

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +438,104 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+    Platform::CriticalSectionExit();
+    return err;
+}
+
+/**
+ * @brief Set mShutdownInProgress flag to true.
+ */
+void LoggingManagement::MarkShutdownInProgress(void)
+{
+    mState = kLoggingManagementState_Shutdown;
```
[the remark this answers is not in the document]
didishe90: kLoggingManagementState_Shutdown is existing state that has been used else where, do you want me to create a new state or change others too? I think both can work.

src/lib/core/WeaveTLVWriter.cpp
```diff
@@ -177,6 +178,43 @@ void TLVWriter::Init(uint8_t *buf, uint32_t maxLen)
     FinalizeBuffer = NULL;
 }
 
+/**
+ * Initializes a TLVWriter object to write into a dynamic buffer.
+ *
+ * @param[in]   buf     A pointer to the dynamic buffer into which TLV should be written.
```
pidarped, line 184: initialBufSize missing from comments.

src/lib/core/WeaveTLVWriter.cpp
```diff
@@ -177,6 +178,43 @@ void TLVWriter::Init(uint8_t *buf, uint32_t maxLen)
...
+ * @param[in]   maxLen  The maximum number of bytes that should be written to the output buffer.
+ * @param[in]   allowResize
+ *                      If true, allocating new space as needed to store the data written.  If
+ *                      false, writing will fail with WEAVE_ERROR_BUFFER_TOO_SMALL if the written
+ *                      data exceeds the initial size of the dynamic buffer
+ */
+void TLVWriter::InitMalloced(uint8_t **&buf, uint32_t initialBufSize, uint32_t maxLen)
+{
+    buf = (uint8_t **)&mBufHandle;
```
pidarped, line 193: Instead of doing this here, would it be simpler to just assign mBufHandle as mBufHandle = (uintptr_t) (*buf); after the allocation is successful in line 199?

src/lib/core/WeaveTLVWriter.cpp
```diff
@@ -177,6 +178,43 @@ void TLVWriter::Init(uint8_t *buf, uint32_t maxLen)
...
+
+    if (initialBufSize <= 0)
```
pidarped, line 195: initialBufSize cannot be < 0. I would recommend, have a VerifyOrExit(initialBufSize > 0, err = WEAVE_ERROR_INVALID_ARGUMENT); 
Also, can initialBufSize be greater than maxLen? Then maybe cap it at maxLen.

src/lib/core/WeaveTLVWriter.cpp
```diff
@@ -177,6 +178,43 @@ void TLVWriter::Init(uint8_t *buf, uint32_t maxLen)
...
+    {
+        initialBufSize = 1;
+    }
+    *buf = (uint8_t *) malloc(initialBufSize);
+
```
pidarped, line 200: Add a VerifyOrExit here with a WEAVE_ERROR_NO_MEMORY.

src/lib/core/WeaveTLVWriter.cpp
```diff
@@ -177,6 +178,43 @@ void TLVWriter::Init(uint8_t *buf, uint32_t maxLen)
...
+    mBufStart = mWritePoint = *buf;
+    mMaxLen = maxLen;
+    mRemainingLen = initialBufSize;
+    if (mRemainingLen > mMaxLen)
+    {
+        mRemainingLen = mMaxLen;
+    }
```
pidarped, line 207: This, potentially, may not be required with some of the checks at the beginning.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +438,104 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+
+        eventBuffer = eventBuffer->mNext;
+    }
+
+    err = writer.EndContainer(container);
+    SuccessOrExit(err);
+
+exit:
+    Platform::CriticalSectionExit();
+    return err;
```
pidarped, line 476: nit: newline before return statement.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -436,6 +438,104 @@ void LoggingManagement::DestroyLoggingManagement(void)
...
+    }
+
+    err = reader.VerifyEndOfContainer();
+    SuccessOrExit(err);
+    err = reader.ExitContainer(container);
+    SuccessOrExit(err);
+
+exit:
+    Platform::CriticalSectionExit();
+    return err;
```
pidarped, line 512: nit: newline before return statement.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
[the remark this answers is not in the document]
pidarped: If it is existing state then it's fine. No need to add a new state for this.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -2146,6 +2246,171 @@ void CircularEventReader::Init(CircularEventBuffer * inBuf)
...
+{
+    WEAVE_ERROR err = WEAVE_NO_ERROR;
+
+    TLVType container;
+
+    uint8_t *inBufStart = mBuffer.QueueHead();
+    size_t inBufLen = mBuffer.DataLength();
+    uint8_t *bufStart = mBuffer.GetQueue();
+    size_t bufLen = mBuffer.GetQueueSize();
+    uint8_t *eventData = new uint8_t[inBufLen];
```
pidarped, line 2259: We should not need to allocate here, and it could be a local array variable, right?

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -2146,6 +2246,171 @@ void CircularEventReader::Init(CircularEventBuffer * inBuf)
...
+    memcpy(eventData, inBufStart, initLen);
+    memcpy(eventData + initLen, bufStart, remainingLen);
+
+    err = writer.StartContainer(AnonymousTag, kTLVType_Structure, container);
+    SuccessOrExit(err);
+
+    err = writer.Put(ContextTag(kTag_PersistEvent_ImportanceLevel), (uint8_t)mImportance);
+    SuccessOrExit(err);
+
+    err = writer.PutBytes(ContextTag(kTag_PersistEvent_EventData), eventData, mBuffer.DataLength());
```
pidarped, line 2281: Shouldn't bufStart be sufficient in putting bytes into the writer. Is eventData required with the associated memcpy?

src/lib/profiles/data-management/Current/LoggingManagement.cpp
pidarped: @jaylogue , if we are persisting events and loading was successful, then does the persistedEventIdCounter need to be used? It should help only in the case of a crash or when the persisting/loading failed, isn't it? Otherwise, it could, potentially, start after the lastEventID?

src/lib/core/WeaveTLVWriter.cpp
didishe90: This function returns void, so i guess we can only print a log if the argument is out of range.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
didishe90: I think so, I blv the event buffer itself is a local array.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
didishe90: do you mean I should put two blocks of data under two different labels? I guess if we want to put two blocks under the same label, we need to do memcpy first, is that right

jaylogue: We could eliminate the need to malloc by adding methods to WeaveTLVWriter for writing a byte string in multiple chunks.  E.g.:

TLVWriter::StartPutBytes(uint64_t tag, uint32_t totalLen);
TLVWriter::ContinuePutBytes(const uint8_t *buf, uint32_t len);

With this, you can put the event data into the writer directly from the queue buffer.

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -27,11 +27,13 @@
 
 #include <Weave/Profiles/data-management/Current/WdmManagedNamespace.h>
 #include <Weave/Profiles/data-management/DataManagement.h>
+#include <Weave/Profiles/WeaveProfiles.h>
 
 #include <Weave/Profiles/bulk-data-transfer/Development/BulkDataTransfer.h>
 #include <Weave/Profiles/bulk-data-transfer/Development/BDXMessages.h>
 
 #include <SystemLayer/SystemTimer.h>
+#include <algorithm>
```
jaylogue, line 36: This is unnecessary now, right?

src/lib/core/WeaveTLV.h
```diff
@@ -30,6 +30,7 @@
 
 #include <stdlib.h>
 #include <stdarg.h>
+#include <vector>
```
jaylogue, line 33: Unnecessary?

PR #596, opened 2020-05-21
Add underflow checking for TLVWriter::VPutStringF

holbrookt:


src/lib/core/WeaveTLVWriter.cpp
```diff
@@ -878,7 +878,10 @@ WEAVE_ERROR TLVWriter::VPutStringF(uint64_t tag, const char *fmt, va_list ap)
     // no facilities for splitting the stream across multiple buffers,
     // just write however much fits in the current buffer.
     // assume conservative tag length at this time (8 bytes)
-    maxLen =  mRemainingLen - (1 + 8 + (1 << static_cast<uint8_t>(lenFieldSize)) + 1); // 1 : control byte, 8 : tag length, stringLen + 1 for null termination
+    size_t headerLen = (1 + 8 + (1 << static_cast<uint8_t>(lenFieldSize)) + 1); // 1 : control byte, 8 : tag length, stringLen + 1 for null termination
```
holbrookt, line 881: Is headerLen an accurate variable name here? Is that what this line is calculating?

src/lib/core/WeaveTLVWriter.cpp
```diff
@@ -878,7 +878,10 @@ WEAVE_ERROR TLVWriter::VPutStringF(uint64_t tag, const char *fmt, va_list ap)
...
+    VerifyOrExit(mRemainingLen >= headerLen, err = WEAVE_ERROR_BUFFER_TOO_SMALL);
```
holbrookt, line 882: Is WEAVE_ERROR_BUFFER_TOO_SMALL an accurate error to return? If this check is false, it means that there is not enough space left

src/lib/core/WeaveTLVWriter.cpp
robszewczyk: `elementHeadLen` would be more aligned with the terminology.

src/lib/core/WeaveTLVWriter.cpp
robszewczyk: Yup, thats the right error

PR #600, opened 2020-06-04
Reduce possibilities of null ptr dereference in PacketBuffer and TCPEndPoint

holbrookt:
1. remove an initial NULL this pointer check in PacketBuffer::Consume()
which might mislead a reader to think that it is checking for this == NULL
case. Compiler assumes this can never be NULL and does not do the initial
check.
2. Check for null sendqueue in TCPEndPoint::HandleDataSent()

### DIFF
--- a/src/inet/TCPEndPoint.cpp
+++ b/src/inet/TCPEndPoint.cpp
@@ -1844,7 +1844,10 @@ void TCPEndPoint::HandleDataSent(uint16_t lenSent)
     if (IsConnected())
     {
         // Consume data off the head of the send queue equal to the amount of data being acknowledged.
-        mSendQueue = mSendQueue->Consume(lenSent);
+        if (mSendQueue)
+        {
+            mSendQueue = mSendQueue->Consume(lenSent);
+        }
 
 #if INET_CONFIG_OVERRIDE_SYSTEM_TCP_USER_TIMEOUT
         // Only change the UserTimeout timer if lenSent > 0,
--- a/src/system/SystemPacketBuffer.cpp
+++ b/src/system/SystemPacketBuffer.cpp
@@ -333,7 +333,7 @@ PacketBuffer* PacketBuffer::Consume(uint16_t aConsumeLength)
 {
     PacketBuffer* lPacket = this;
 
-    while (lPacket != NULL && aConsumeLength > 0)
+    while (aConsumeLength > 0)
     {
         const uint16_t kLength = lPacket->DataLength();
 
@@ -341,6 +341,10 @@ PacketBuffer* PacketBuffer::Consume(uint16_t aConsumeLength)
         {
             lPacket = PacketBuffer::FreeHead(lPacket);
             aConsumeLength -= kLength;
+            if (lPacket == NULL)
+            {
+                break;
+            }
         }
         else
         {


src/system/SystemPacketBuffer.cpp
```diff
@@ -333,14 +333,18 @@ PacketBuffer* PacketBuffer::Consume(uint16_t aConsumeLength)
 {
     PacketBuffer* lPacket = this;
 
-    while (lPacket != NULL && aConsumeLength > 0)
+    while (aConsumeLength > 0)
```
gerickson, line 336: Removing the NULL check causes a potential fault on line 338 if lPacket is NULL. Is the argument that the `this` pointer should never be NULL?

src/system/SystemPacketBuffer.cpp
```diff
@@ -333,14 +333,18 @@ PacketBuffer* PacketBuffer::Consume(uint16_t aConsumeLength)
...
         const uint16_t kLength = lPacket->DataLength();
 
         if (aConsumeLength >= kLength)
         {
             lPacket = PacketBuffer::FreeHead(lPacket);
             aConsumeLength -= kLength;
+            if (lPacket == NULL)
+            {
+                break;
+            }
```
pidarped, line 347: I think leaving the lPacket != NULL check in line 336 to prevent it from entering the while loop at all if it is NULL is simpler than adding the break, isn't it?

robszewczyk: +1. 

Additionally, `this` can be `NULL`:  if and when it is, in case of `virtual` functions, you'd expect to get a fault at the call site; in cases such as this, you'd be faulting in the entry into the loop.

src/inet/TCPEndPoint.cpp
```diff
@@ -1844,7 +1844,10 @@ void TCPEndPoint::HandleDataSent(uint16_t lenSent)
     if (IsConnected())
     {
         // Consume data off the head of the send queue equal to the amount of data being acknowledged.
-        mSendQueue = mSendQueue->Consume(lenSent);
+        if (mSendQueue)
```
robszewczyk, line 1847: I'm trying to work through the logic of the larger state machine to check whether it is possible for `mSendQueue` to be `NULL` at this point in time.  I note that `HandleDataSent` is private and thus should not be invoked at an arbitrary point in the execution.  It is processed as an event , and only as a result of the LwIP `tcp_sent` callback.  There is generally a benefit to having the system fail fast when the core invariants are violated; this change perhaps makes the system fail more gracefully at a cost of sweeping some issues under the rug.  If you have an identified codepath that leads to the check being required here, it would be helpful.

jaylogue: I'm pretty sure this change is hiding a bigger problem.  As I see it, the scenario is: LwIP has called back into the InetLayer (via a call to LwIPHandleDataSent()) stating that a particular amount of previously sent data has been ACKed by the peer.  This caused an event to be posted to the Weave thread.  InetLayer::HandleInetLayerEvent() dispatched this event to TCPEndPoint::HandleDataSent().  At this point, the endpoint is still connected (IsConnected() == true).  However, the pointer to the buffer containing the sent data (mSendQueue) is now NULL.  

This state is inconsistent with the logic of TCPEndPoint, and of the LwIP API.  When using the LwIP raw API, the sender is required to preserve the buffer containing the data to be sent until it receives a callback indicating that all the data has been acknowledged by the peer. 

Now, PacketBuffer::Consume() will return a NULL if the given length consumes the entirety of the data in the buffer.  And, indeed, this is normal way that mSendQueue becomes NULL: the peer ACKs the data, Comsume() consumes the number of bytes ACKed, and if this results in the buffer being empty, it frees the buffer and returns NULL.

In theory, if LwIP generated an _extraneous_ callback to the data acknowledged function after all sent data had been previously acknowledged, then that would result in HandleDataSent() calling Consume() with a NULL this pointer.  (Note that if this did happen, then I would expect lenSent to be 0... because what else could it be?)  Perhaps this situation can arise when ACKs are retransmitted at the TCP level.  Although I still would consider this a bug in LwIP.

That theory aside, if mSendQueue is NULL, something more subtle is going on that needs to be understood.

robszewczyk: @jaylogue -- agreed.

src/system/SystemPacketBuffer.cpp
jaylogue, line 336: I'm not strongly against this change, but its not adding any significant value.  The change merely moves the exit test from one point in the loop to another, under the assumption that lPacket should never be NULL when the loop is entered.   As I see it, it doesn't make the code appreciably faster, smaller, easier to read or more correct.

jaylogue: I understand better now how the compiler was moving the check to the bottom of the loop anyway, and that this change merely made that "optimization" explicit.  Again, I fine with this change.

PR #601, opened 2020-06-10
Support new OpenThread with OPENTHREAD_API_VERSION

bukepo:
To use new OpenThread, set OPENTHREAD_API_VERSION_MINIMAL to a
non-zero value which indicates the minimal OpenThread API version
required for the project.

This commit keeps OPENTHREAD_API_VERSION_MINIMAL to 0 by default because
other openweave projects is still using an old OpenThread.


bukepo:
ping~

PR #602, opened 2020-06-12
Load trait version from persisted storage

didishe90:
Calculate hash of serialized trait and compare it to persisted value. If consistent, load trait version from persisted storage.

### DIFF
--- a/src/lib/profiles/data-management/Current/TraitData.h
+++ b/src/lib/profiles/data-management/Current/TraitData.h
@@ -914,6 +914,9 @@ public:
     bool mRootIsDirty;
 #endif
 
+    // Set current version of the data in this source.
+    void SetVersion(uint64_t version) { mVersion = version; }
+
 protected: // IGetDataDelegate
     /*
      * Defaults to calling GetLeafData if aHandle is a leaf. DataSources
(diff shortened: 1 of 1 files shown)


src/lib/profiles/data-management/Current/TraitData.cpp
```diff
@@ -1332,6 +1336,31 @@ void TraitDataSource::SetDirty(PropertyPathHandle aPropertyHandle)
     }
 }
 
+#if WEAVE_CONFIG_PERSIST_TRAIT
+WEAVE_ERROR TraitDataSource::GenerateTraitHash(uint8_t * hash)
```
pidarped, line 1340: Does this function need to be here? ReadData is a public function and it could as well be part of platform code that is handling the persistence.

src/lib/profiles/data-management/Current/SubscriptionHandler.cpp
```diff
@@ -408,6 +408,12 @@ WEAVE_ERROR SubscriptionHandler::ParsePathVersionEventLists(SubscribeRequest::Pa
             else
             {
                 uint64_t existingVersion;
+
+#if WEAVE_CONFIG_PERSIST_TRAIT
+                err = platform::UpdateVersionFromPersistedStorage(dataSource);
```
pidarped, line 413: For this, we are updating the version for only those in the SubscribeRequest. I guess we can go over all the sources in the Catalog before a subscription is attempted and update the versions then? That way we do not need to insert a version update logic here, potentially.

pidarped (approved):

robszewczyk (approved):

PR #606, opened 2020-06-26
Fix for Message Too Long error

jaylogue:
Modified WeaveConnection::HandleDataReceived() to properly handle the case where an incoming Weave message arrives in multiple TCP segments and the underlying network driver places the first such segment into a packet buffer that is too small to hold the entire message.

Since the message decoding logic in Weave requires message data to be in a contiguous region of memory, code in HandleDataReceived() automatically copies message data received after the initial segment into the initial packet buffer, expecting that buffer to always be big enough to hold the entire message.  However if the underlying network driver places incoming packets into buffers sized only big enough to fit the packet data, reception of large Weave messages fails with WEAVE_ERROR_MESSAGE_TOO_LONG.

The fix is to detect the case where the initial buffer is not big enough to hold the entire Weave message and allocate a new appropriately sized buffer in which to receive the data.

### DIFF
--- a/src/lib/core/WeaveConnection.cpp
+++ b/src/lib/core/WeaveConnection.cpp
@@ -1403,10 +1403,50 @@ void WeaveConnection::HandleDataReceived(TCPEndPoint *endPoint, PacketBuffer *da
         msgInfo.InPacketInfo = &packetInfo;
         msgInfo.InCon = con;
 
-        // Attempt to parse an message from the head of the received data.
+        // Attempt to parse an message from the head of the received queue.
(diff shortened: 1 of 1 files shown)


jaylogue:
The latest build actually succeed per the logs, despite be marked as failed by Travis.  This might be due to the following error from pip:

```
Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/lib/python2.7/dist-packages/protobuf-3.0.0b2-py2.7-nspkg.pth'
Consider using the `--user` option or check the permissions.
```

Since this is unrelated to this change, I'm going to merge anyway.

PR #608, opened 2020-06-26
Enable variable-sized packet buffers in standalone LwIP builds

jaylogue:
- Modified the standalone build to enable the use of variable-sized packet buffers when running under LwIP. This makes it possible to test certain buffer handling scenarios that don’t arise when all buffers are the same size.  Use of variable-sized buffers is enabled by default, but can be disabled if need be by setting LWIP_PBUF_FROM_CUSTOM_POOLS to 0 in lwipopts.h.

- Modified the LwIP input handling function to allocate buffers for inbound packets without reserve space.  Previously the code reserved header space which was never used and limited the maximum size packet that could be received.

- Modified TestRetainedPacketBuffer app to accommodate the use of variable-sized packet buffers.

- Fixed a long-standing bug in the definition of WEAVE_SYSTEM_CONFIG_PACKETBUFFER_CAPACITY_MAX. Previously, this value was defined as PBUF_POOL_BUFSIZE minus the size of a pbuf header.  However, LwIP expects PBUF_POOL_BUFSIZE to be the *payload* size of a pbuf.  Thus one needn’t subtract the pbuf header size when computing a “max capacity”.

- Fixed another long-standing bug in PacketBuffer::NewWithAvailableSize(). When calling pbuf_alloc(), the length parameter is the desired *payload* size of the buffer.  However, NewWithAvailableSize() was adding in the size of a pbuf header before calling the function.

- Eliminated the use of PBUF_LINK_ENCAPSULATION_HLEN in the standalone LwIP output handling function.  This served no purpose, as evidenced by a spurious comment about being “needed by WICED”.

- Disabled MEMP_OVERFLOW_CHECK.  A bug exists in openweave’s version of LwIP (since fixed upstream in commit 2fd2b68) which results in failures in the TestPacketBuffer unit test whenever this feature is enabled.  In a nutshell, the problem is that the runtime size of a pbuf ends up being bigger than the declared size (e.g. PBUF_POOL_BUFSIZE), which results in a mismatch between the size returned by AllocSize() and the value of WEAVE_SYSTEM_CONFIG_PACKETBUFFER_CAPACITY_MAX. This triggers various asserts in TestPacketBuffer.  The temporary fix is to disable the feature until the LwIP code is updated.

- Cleaned up comments and code structure in lwipopts.h.


src/test-apps/TestRetainedPacketBuffer.cpp
```diff
@@ -258,7 +275,7 @@ static void CheckRetainAllocatedBufferWithImplicitRelease(nlTestSuite *inSuite,
 
     // Assert that the buffer has been released to the pool, as expected.
 
-    NL_TEST_ASSERT(inSuite, System::Stats::GetResourcesInUse()[System::Stats::kSystemLayer_NumPacketBufs] == 0);
+    NL_TEST_ASSERT_NO_BUFFERS_IN_USE(inSuite);
```
gerickson, line 278: Unable to channel my inner Giuseppe, what are we losing by eliminating the fault injection / stats?

jaylogue: Actually all I did was define a custom assert macro (see above) that asserts there are no buffers in use, using the same System::Stats interface.  This made is easier to vary the assertion between testing one stat value (in the case of uniform sized buffers) verses three values (for variable-sized buffers).

jaylogue:
Note to reviewers: I have made some significant additions to this PR which likely warrant further review.  In particular, I fixed some long-standing bugs in System::PacketBuffer code, which we might want to review for down-stream consequences (although I think this unlikely).

src/test-apps/TestRetainedPacketBuffer.cpp
gerickson: Missed that, thanks.

src/lwip/standalone/TapInterface.c
```diff
@@ -197,7 +197,7 @@ TapInterface_low_level_output(struct netif *netif, struct pbuf *buf)
     if (buf->tot_len > buf->len)
     {
         // Allocate a buffer from the buffer pool. Fail if none available.
-        outBuf = pbuf_alloc(PBUF_RAW, buf->tot_len + PBUF_LINK_ENCAPSULATION_HLEN, PBUF_POOL);
+        outBuf = pbuf_alloc(PBUF_RAW, buf->tot_len, PBUF_POOL);
```
robszewczyk, line 200: Just to confirm: the bug here was that for the tap device there is no additional link encapsulation like there was for WICED.

jaylogue: Yep.  No bug, per se, just some confusing and unnecessary code/comments.

PR #615, opened 2020-07-23
Add ability to disable rcv and send debug messages

SurekhaMuralidharan777:
Creating pull request to filter rcv and send weave message debug prints by profile.


jaylogue:
I don't understand the utility of this change.  Why is it valuable to be able to disable logging of incoming/outgoing messages based on the message's profile?  As it stands, the feature only disables the "sent"/"rcvd" log lines, but not any other log lines that might be associated with processing the message.  Why is this useful?  (Actually, this seems harmful in that, without the "sent"/"rcvd" log lines, the other log lines will lack the context necessary to make them meaningful).

Additionally, the implementation introduces code and complexity to the core of weave for all platforms, without any conditional compilation.  However, to date, we've never seen a need to dynamically enable/disable message log lines in any product that uses openweave.  Thus, this sounds likes a special case requirement that might be better handled in another, less general way.

PR #624, opened 2020-10-09
Tunnel CASE Session persistence support.

pidarped:
Changes to support persistence of CASE session for the Tunnel.

During connection establishment by the Tunnel Agent, if a persisted
session key exists, the TunnelAgent first establishes an unsecured 
WeaveConnection and then calls back into the application to load the
session key for the created Weave connection object.
When the connection goes down, and before the FabricState removes the
SessionKey, the application is called back to suspend the session into
a serialized form to be persisted.
At time of suspension, resumption message Ids(for both the next message
id to send, as well as the last received message id from peer) are
calculated using a fixed offset and included within the session key.
When the session resumes, it would start off communication using the
resumption message ids.

Added Test Client and Server for testing persistence of Tunnel CASE
session.

### DIFF
--- a/src/lib/core/WeaveFabricState.cpp
+++ b/src/lib/core/WeaveFabricState.cpp
@@ -113,6 +113,10 @@ void WeaveSessionKey::Init(void)
     NodeId = kNodeIdNotSpecified;
     NextMsgId.Init(0);
     MaxRcvdMsgId = 0;
+    InitialSendMsgId = 0;
+    InitialRcvdMsgId = 0;
+    ResumptionSendMsgId = 0;
+    ResumptionRecvMsgId = 0;
     BoundCon = NULL;
     RcvFlags = 0;
     AuthMode = kWeaveAuthMode_NotSpecified;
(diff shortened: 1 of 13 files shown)


robszewczyk (approved):

src/lib/core/WeaveFabricState.cpp
```diff
@@ -113,6 +113,10 @@ void WeaveSessionKey::Init(void)
     NodeId = kNodeIdNotSpecified;
     NextMsgId.Init(0);
     MaxRcvdMsgId = 0;
+    InitialSendMsgId = 0;
```
emargolis, line 116: I think these Initial* fields are redundant and Resumption message Ids should be calculated using NextMsgId value.

src/lib/core/WeaveFabricState.cpp
```diff
@@ -130,6 +134,22 @@ void WeaveSessionKey::Clear(void)
     ClearSecretData((uint8_t *)&MsgEncKey.EncKey, sizeof(MsgEncKey.EncKey));
 }
 
+void WeaveSessionKey::ComputeNextResumptionMsgIds(void)
+{
+     // When calculating the resumption message ids, it needs to be ensured that the next resumption message id is always ahead of the current message ids.
+     while (ResumptionSendMsgId < NextMsgId.GetValue())
+     {
+         ResumptionSendMsgId = (ResumptionSendMsgId == 0) ? InitialSendMsgId + WEAVE_CONFIG_TUNNEL_SESSION_RESUMPTION_MSG_ID_OFFSET :
```
emargolis, line 142: As we discussed offline I propose to change this calculation mechanism.

src/lib/core/WeaveFabricState.cpp
```diff
@@ -113,6 +113,10 @@ void WeaveSessionKey::Init(void)
     NodeId = kNodeIdNotSpecified;
     NextMsgId.Init(0);
     MaxRcvdMsgId = 0;
+    InitialSendMsgId = 0;
+    InitialRcvdMsgId = 0;
+    ResumptionSendMsgId = 0;
```
emargolis, line 118: Should all this new code ifdefed with WEAVE_CONFIG_PERSIST_CONNECTED_SESSION ?

emargolis (approved):

src/lib/core/WeaveFabricState.cpp
pidarped: I initially did that. But Cloud uses the initial values to calculate the resumption msgIds. The reasons range from having to do less lookups/writes to keep track of last messageId to having a resumption messageId available when the device connects a second time while the Cloud side is still holding on to its old connection.

src/lib/core/WeaveFabricState.cpp
pidarped: In FabricState, I am reusing the Suspend and Restore functions that also do persistence for end-to-end CASE sessions over WRM. So, I only had that WEAVE_CONFIG within the Tunnel code.

src/lib/core/WeaveFabricState.cpp
pidarped: Cloud also has the same logic currently. But they can make an adjustment to allow a range from the resumption msgId within a certain limit. That should cover the case of the device moving ahead of the resumptionmsgIds by a few counts. This would not require any additional changes on device side. However, in practice, I am skeptical this would happen as the device ought to tear down the connection entirely and retry, if the TunnelOpen message fails to go or times out.

PR #634, opened 2020-12-07
Fix Weave TCPEndPoint crash in some lwIP cases

holbrookt:
* Weave, when using an lwIP networking interface, employs
  the raw lwIP callback-based scheme to handle TCP sends.
  In order to avoid multiple copies of packet buffers on
  small devices, the TCP sends are done in-place over the
  incoming packet buffers. Packet buffers must be retained
  in full until acknowledgement has been received by lwIP.
* The existing logic had a corner case where if there was
  more data queued than the local TCP send window, the
  accounting of buffers and lengths to retain the un-acked
  buffers could get out of balance, leading to an eventual
  crash. This was root caused to an invalid assumption that
  the packet buffer chain did not need traversal for offset
  readjustment (which is true in most cases, only by
  coincidence).
* This change removes separate accounting of the unsent
  data position and replaces it with simple counting of
  sent-but-not-acked data and a skip-loop at the start
  of DriveSending().

### DIFF
--- a/src/inet/TCPEndPoint.cpp
+++ b/src/inet/TCPEndPoint.cpp
@@ -708,12 +708,6 @@ INET_ERROR TCPEndPoint::Send(PacketBuffer *data, bool push)
 
 #if WEAVE_SYSTEM_CONFIG_USE_LWIP
 
-    if (mUnsentQueue == NULL)
-    {
-        mUnsentQueue = data;
-        mUnsentOffset = 0;
-    }
-
 #if INET_CONFIG_OVERRIDE_SYSTEM_TCP_USER_TIMEOUT
     if (!mUserTimeoutTimerRunning)
     {
@@ -1199,6 +1193,10 @@ void TCPEndPoint::Init(InetLayer *inetLayer)
 #endif // WEAVE_SYSTEM_CONFIG_USE_SOCKETS
 
 #endif // INET_CONFIG_OVERRIDE_SYSTEM_TCP_USER_TIMEOUT
+
+#if WEAVE_SYSTEM_CONFIG_USE_LWIP
+    mUnackedLength = 0;
+#endif // WEAVE_SYSTEM_CONFIG_USE_LWIP
 }
 
 INET_ERROR TCPEndPoint::DriveSending()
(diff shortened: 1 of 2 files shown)


gerickson (approved):

gerickson:
Do we have a Happy / Cirque unit or functional test for this?

holbrookt:
> Do we have a Happy / Cirque or unit or functional test for this?

Not that I know of. I see `TestInetLayer.cpp` in src/test-apps, but it only does one basic test. Where would I look for Cirque tests?

gerickson:
> > Do we have a Happy / Cirque or unit or functional test for this?
> 
> Not that I know of. I see `TestInetLayer.cpp` in src/test-apps, but it only does one basic test. Where would I look for Cirque tests?

Any Happy tests would also be under src/test-apps/happy/....

robszewczyk (approved):

holbrookt:
> > > Do we have a Happy / Cirque or unit or functional test for this?
> > 
> > 
> > Not that I know of. I see `TestInetLayer.cpp` in src/test-apps, but it only does one basic test. Where would I look for Cirque tests?
> 
> Any Happy tests would also be under src/test-apps/happy/....

There is `happy/tests/standalone/inet/test_inet_01.py` which does a basic send/receive test. It currently tests sending various lengths of data, including less than and greater than MTU. I think if we added a case for sending data larger than the send window size, it would cover this fix. Should I add that case?

pidarped:
> > > > Do we have a Happy / Cirque or unit or functional test for this?
> > > 
> > > 
> > > Not that I know of. I see `TestInetLayer.cpp` in src/test-apps, but it only does one basic test. Where would I look for Cirque tests?
> > 
> > 
> > Any Happy tests would also be under src/test-apps/happy/....
> 
> There is `happy/tests/standalone/inet/test_inet_01.py` which does a basic send/receive test. It currently tests sending various lengths of data, including less than and greater than MTU. I think if we added a case for sending data larger than the send window size, it would cover this fix. Should I add that case?

That seems like a relatively easy enough test case to add and should cover this error scenario. Would be good if we can add that in. Adding that in a separate PR is probably ok too since we need this change urgently.

PR #635, opened 2020-12-09
Add a check to drop a large event.

pidarped:
Drop the event from being logged in the buffer if it exceeds the maximum WDM Notification message limit.


src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -1314,6 +1314,7 @@ inline event_id_t LoggingManagement::LogEventPrivate(const EventSchema & inSchem
         CircularEventBuffer * buffer = mEventBuffer;
         do
         {
+            VerifyOrExit(WDM_MAX_NOTIFICATION_SIZE >= writer.GetLengthWritten(), err = WEAVE_ERROR_WDM_EVENT_TOO_BIG);
```
robszewczyk, line 1317: Per conversation in chat -- need to accommodate the some number of addtional headers that are associated with embedding the event in a notification. We could get that number precisely, or have a reasonably conservative estimate.

robszewczyk: I'm approving the PR, with the assumption that the delta is added and reflected in the change

robszewczyk (approved):

src/lib/profiles/data-management/Current/LoggingManagement.cpp
pidarped: Added an estimated max of 64 bytes for the additional meta information needed to formulate the NotifyRequest message.

PR #637, opened 2020-12-29
Add more logs to help debug event missing issue

didishe90:
In this change we 
1) specify the event id range of dropped events, 
2) Log profileId and print out the message when logEvent failed and returned 0.
3) Log range of event_id fetched and being flushed

### DIFF
--- a/src/lib/profiles/data-management/Current/LoggingManagement.cpp
+++ b/src/lib/profiles/data-management/Current/LoggingManagement.cpp
@@ -1094,6 +1094,10 @@ event_id_t LoggingManagement::LogEvent(const EventSchema & inSchema, EventWriter
 
 exit:
     Platform::CriticalSectionExit();
+    if (event_id == 0)
+    {
+        WeaveLogProgress(EventLogging, "Failed to log event with ProfileId: %d, return event id 0", inSchema.mProfileId);
+    }
     return event_id;
 }
 
@@ -2117,7 +2121,9 @@ void CircularEventBuffer::AddEventUTC(utc_timestamp_t inEventTimestamp)
 
 void CircularEventBuffer::RemoveEvent(size_t aNumEvents)
 {
+    event_id_t currentFirstEventID = mFirstEventID;
     mFirstEventID += aNumEvents;
+    WeaveLogDetail(EventLogging, "Dropping events | Move first event id from %u to %u", currentFirstEventID, mFirstEventID);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -1094,6 +1094,10 @@ event_id_t LoggingManagement::LogEvent(const EventSchema & inSchema, EventWriter
 
 exit:
     Platform::CriticalSectionExit();
+    if (event_id == 0)
+    {
+        WeaveLogProgress(EventLogging, "Failed to log event, return event id 0");
```
pidarped, line 1099: Recommend logging the ProfileId from the inSchema as well.

robszewczyk: looks resolved.

robszewczyk (approved):

src/lib/profiles/data-management/Current/LoggingManagement.cpp
```diff
@@ -2117,7 +2121,9 @@ void CircularEventBuffer::AddEventUTC(utc_timestamp_t inEventTimestamp)
 
 void CircularEventBuffer::RemoveEvent(size_t aNumEvents)
 {
+    event_id_t currentFirstEventID = mFirstEventID;
     mFirstEventID += aNumEvents;
+    WeaveLogProgress(EventLogging, "Dropping events | Move first event id from %u to %u", currentFirstEventID, mFirstEventID);
```
robszewczyk, line 2126: I think with this log line in place you're looking at a whole lot of log lines.  In a steady state, you're probably looking at (an average) of one of the above lines per LogEvent.  This is no longer a `Progress` level of logging, more like `debug`. Because `WeaveLog*` family of functions are synchronous, these log lines can occupy a significant time to output, particularly when the underlying serial console is slow.  `WeaveLogDebug` would be far more appropriate here.

robszewczyk (changes requested):

src/lib/profiles/data-management/Current/LoggingManagement.cpp
didishe90: I thought RemoveEvent is only called here https://source.corp.google.com/nest/platform/weave/master/src/lib/profiles/data-management/Current/LoggingManagement.cpp;rcl=6854575c8beb472be83b67590f9830730b733487;l=1781
when events are being dropped due to overflow. This is not where we move the counter when fetching events. And since the buffer size are increased, this shouldn't happen that often. Is my understanding wrong?

robszewczyk: In steady state -- all event buffers are full and the average size of event is stationary -- we expect that for every event logged, we need to remove one to free the space in the circular buffers.

didishe90: you mean when we build the event notify, we don't change mFirstEventID? So the events removed here, some of them actually have been sent to service?

robszewczyk (approved):

PR #640, opened 2021-01-14
Add fileName and lineNumber in WeaveLog*

didishe90:
Help debug and identify spammy log


gerickson (changes requested):
This will need to be conditionalized by a compile-time configuration flag and should be disabled by default. File names, in particular, will lead to an explosion in debug image code size.

PR #644, opened 2021-01-27
Persistent delivered event id

didishe90:
Allow applications to persist the delivered event ids that have been successfully offloaded to the subscriber across power cycles. 
For a service that subscribes to all events but cannot send up-to-date "delivered event ids" during subscription, enabling this feature can avoid sending duplicated events to service.


robszewczyk:
Unit & functional tests?  Seems like we have a decent bit of new functionality here.

robszewczyk:
Closing this as the functionality has been incorporated into #647

PR #650, opened 2021-02-23
Enable TCP Connection Repair.

pidarped:
A TCP Connection Repair is attempted with stored TCP repair meta
information.
Used for restoring the Tunnel TCP conneciton.
If restoration fails, device falls back to establishing a new TCP
Connection.


src/inet/TCPEndPoint.cpp
```diff
@@ -296,6 +296,263 @@ INET_ERROR TCPEndPoint::Listen(uint16_t backlog)
...
+    INET_ERROR res = INET_NO_ERROR;
+
+#if WEAVE_SYSTEM_CONFIG_USE_LWIP
+    res = INET_ERROR_NOT_SUPPORTED;
+    ExitNow();
+#endif // WEAVE_SYSTEM_CONFIG_USE_LWIP
+
+#if WEAVE_SYSTEM_CONFIG_USE_SOCKETS
+
+#ifndef TCP_REPAIR
```
robszewczyk, line 311: Would it make sense to collapse all conditional compilation into the condition here?  I would certainly argue for that all of the ifdefs below should be combined here.

robszewczyk, line 311:
```suggestion
#if !defined(TCP_REPAIR) || !defined(TCPI_OPT_SACK) ||!defined(TCPI_OPT_WSCALE) ||!defined(TCPI_OPT_TIMESTAMP) ||!defined(TCP_REPAIR_OPTIONS) || !defined(TCPOPT_MAXSEG ) || !defined(TCP_REPAIR_WINDOW)```

File: src/inet/TCPEndPoint.h [hunk 2]
```diff
@@ -38,6 +38,30 @@ namespace Inet {
 
 class InetLayer;
 
+#if INET_CONFIG_TCP_CONN_REPAIR_SUPPORTED
+typedef struct TCPConnRepairInfo
```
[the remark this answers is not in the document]
pidarped: Added IsValid() and Dump() methods. Kept the typedef as the platform build required struct keyword to be otherwise included with TCPConnRepairInfo.

PR #658, opened 2021-04-07
Openweave Python packages should not depend on Python ABI version

phaezah7:
We've noticed that Openweave Python packages on PyPI are only uploaded
for specific Python versions. This was a problem when we switched from
Python 3.7 to Python 3.8, and it was finally solved by uploading a new
Python package to PyPI. But now we've switched to Python 3.9 and the
problem surfaced again... so it's high time we find a better solution.

Since Openweave Python package includes, and wraps around, a binary
shared object, it is not a "pure" package.  However, in that case,
Python's "wheel" packer assumes the package uses Python ABI (e.g. it
builds a Python module as a shared object), and tags it with the version
of ABI and the interpreter. Openweave does not use Python ABI, so this
is an unnecessary constraint. This PR removes it.

### DIFF
--- a/src/device-manager/python/build-openweave-wheel.py
+++ b/src/device-manager/python/build-openweave-wheel.py
@@ -119,6 +119,12 @@ try:
         def finalize_options(self):
             bdist_wheel.finalize_options(self)
             self.root_is_pure = False
+
+        def get_tag(self):
+            python, abi, platform = bdist_wheel.get_tag(self)
+            # Openweave does not depend on Python version or ABI
+            python, abi = 'py2.py3', 'none'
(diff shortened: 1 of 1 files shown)


phaezah7:
@googlebot I fixed it.

phaezah7:
@googlebot I fixed it.

robszewczyk (approved):

phaezah7:
3rd attempt to get the spuriously failing test to pass.

phaezah7:
4th attempt to get `happy/tests/standalone/tunnel/test_weave_tunnel_02.py` to pass.

PR #662, opened 2021-04-27
Make sure protocol callback is set on the binding object in SubscriptionClient

didishe90:
SubscriptionClient is written in a way that it's not released upon abort. When peer send
a "cancel subscription" request, SubscriptionClient is reset to initiated state and
is going to be reused by the next subscription attempt. However, SubscriptionClient
only sets the protocol callback on its binding upon init(), which doesn't happen
when the same subscriptionClient is reused for the second (and new) subscription
attempt.
In recent tests, we found that some application resets binding during AbortSubscription.
In some corner cases, when subscriptionClient is reused for second subscription, 
protocol callback is not set for its binding.  We should double check binding has 
valid protocol callback when InitiateSubscription() happens. This PR is also a temporary 
fix to unblock cancel/re-subscribe for some application.

We might also consider release subscriptionClient upon "cancel" to align with this: 
https://github.com/openweave/openweave-core/blob/9f9d2450ba3308f16cc729324801992d37130813/src/lib/profiles/data-management/Current/SubscriptionClient.h#L446

### DIFF
--- a/src/lib/profiles/data-management/Current/SubscriptionClient.cpp
+++ b/src/lib/profiles/data-management/Current/SubscriptionClient.cpp
@@ -863,6 +863,9 @@ void SubscriptionClient::InitiateSubscription(void)
 {
     mConfig = kConfig_Initiator;
 
+    // Make double sure the protocol callback is set on the binding object
+    mBinding->SetProtocolLayerCallback(BindingEventCallback, this);
+
     if (IsRetryEnabled())
     {
         if (false == mBinding->IsPreparing())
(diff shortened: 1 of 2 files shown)


src/lib/profiles/data-management/Current/SubscriptionClient.cpp
```diff
@@ -863,6 +863,9 @@ void SubscriptionClient::InitiateSubscription(void)
 {
     mConfig = kConfig_Initiator;
 
+    // Make double sure the protocol callback is set on the binding object
```
pidarped, line 866: nit: Change comment to "Ensure that the protocol callback is set..."

pidarped:
The change looks reasonable in that it is good to set the BIndingCallback at InitiateSubscription. SubscribeCancel is not that common from the publisher and, especially since it takes a slightly different path than when the Client terminates the subscription, we need to ensure the Binding is initialized correctly. Approving this to have the changes in. We would need a test case around this path.

pidarped (approved):

PR #674, opened 2021-07-29
Add NULL-check for `MessageLayer->FabricState` on connection close.

lcniles:
https://crash.corp.google.com/browse?q=stable_signature%3D%27nl%3A%3AWeave%3A%3AWeaveConnection%3A%3ADoClose-24bb06ea%27


robszewczyk (approved):

robszewczyk:
The only concern I have with this patch is that it could end up masking a deeper problem.  IIRC the code was vetted at some point to convince ourselves that the NULL checks were in place where needed and somehow we believed that the null check was not necessary here.  But I can think about this offline, while we reap the benefits of this patch

PR #686, opened 2021-12-03
Weave Message Capture logic.

pidarped:
This feature allows an application to formulate an Exchange for
capturing Weave messages by setting a WeaveBindiing flag before vending
the ExchangeContext. For tunneled messages, they are directed towards a
'capture subnet' as a means of tagging the messages for capturing, as
they traverse the underlying network stack and get encapsulated within
an [IPv6|UDP] packet. The capture subnet is then reinstated back to the
original subnet before the packet is captured and delivered up to
the application.

* Expose WeaveBinding setting API.
* Percolate setting via message flags from WeaveExchange layer to
  WeaveMessageLayer.
* Encode reserved capture subnet for Tunneled packets to Phoenix
* Reinstate Phoenix address and tag tunneled packet for capture.
* Add functional test case for capturing a tunneled packet.

### DIFF
--- a/src/lib/profiles/weave-tunneling/WeaveTunnelAgent.cpp
+++ b/src/lib/profiles/weave-tunneling/WeaveTunnelAgent.cpp
@@ -40,15 +40,20 @@
 #include <Weave/Support/WeaveFaultInjection.h>
 #include <Weave/Profiles/weave-tunneling/WeaveTunnelCommon.h>
 #include <Weave/Profiles/weave-tunneling/WeaveTunnelControl.h>
+#include <Weave/Support/logging/DecodedIPPacket.h>
+#include <Weave/Core/WeaveEncoding.h>
 
 #if WEAVE_CONFIG_ENABLE_TUNNELING
 
 #if WEAVE_SYSTEM_CONFIG_USE_LWIP
 #include "lwip/ip6.h"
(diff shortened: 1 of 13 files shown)


src/lib/profiles/weave-tunneling/WeaveTunnelAgent.cpp
```diff
@@ -934,6 +939,150 @@ void WeaveTunnelAgent::ParseDestinationIPAddress(const PacketBuffer &inMsg, IPAd
...
+
+    // Calculate checksum for the pseudo header
+    sum = Checksum((uint16_t*)&payloadWithPseudoHdr, sizeof(PseudoHeader));
+    // Calculate the UDP Datagram checksum and add to the sum.
+    sum += Checksum(q, udpPktLen);
+
+    // Adjust for carryover
+    sum = (sum >> 16) + (sum & 0xffff);
+    sum += (sum >> 16);
+    sum = ~sum;
```
robszewczyk, line 1019: Lines 1011-1019 are a classic UDP checksum / RFC 768.  Is there a method we could call instead?

robszewczyk: Alternatively this could be broken out as a separate small function.

robszewczyk (approved):

PR #687, opened 2022-01-07
Weave tool improvements

robszewczyk:
* add a command to print out the contents of the access token
* add a command to print out the contents of the service config object
* add the ability to specify the implicit profile ID to the print-tlv command
* change the printing of binary string dump to dump the contents rather than the pointer value

### DIFF
--- a/src/tools/weave/Cmd_PrintAccessToken.cpp
+++ b/src/tools/weave/Cmd_PrintAccessToken.cpp
@@ -74,7 +74,8 @@ static HelpOptions gHelpOptions(
     "\n"
     "  <access-token>\n"
     "\n"
-    "       A file containing a Weave Access Token, in base-64 format\n"
+    "       A file containing a Weave Access Token either in raw TLV format (default)\n"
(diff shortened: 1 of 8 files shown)


src/tools/weave/Cmd_PrintAccessToken.cpp
```diff
@@ -0,0 +1,193 @@
...
+    certSet.Init(kNumCerts, kCertBufSize);
+    err = LoadAccessTokenCerts(reader, certSet, 0, certData);
+
+    if (err != WEAVE_NO_ERROR)
+    {
+        fprintf(stderr, "Error reading cert info: %s.\n", nl::ErrorStr(err));
+        ExitNow(res = false);
+    }
+
+    err = DetermineCertType(*certData);
```
pidarped, line 138: Is this necessary? For access tokens, this function sets the type as kCertType_General.

pidarped (approved):

emargolis (approved):

PR #691, opened 2022-08-08
Fix error in TextDescriptorReader when data contains leading empty spaces

mburshteyn1:
QR Code parsing is failing when QR code contains leading spaces. 

`Next` function advances the pointer by 1 past `mFieldEnd` to skip the field separator. When empty spaces are present the initial `Next` call advances past the first empty space before performing the empty space check, ending up on the version value instead of the initial key. This change initializes `mFieldEnd` to match `mReadPoint` which is `val` with removed empty spaces.


lcniles:
lgtm

robszewczyk (approved):

